PR #8, opened 2025-06-16
Implemented support for s3 as a source

Ninad-Bhangui:
# 📦 Pull Request

---

## 📋 What does this PR do?

This PR adds support to process files in AWS S3.

It does so by converting s3 file into a pre-signed URL. I've noticed it internally uses `vidgear` library for processing which already supports http streaming so my idea was to convert the s3 path into a pre-signed url and then use the existing logic.

- `TODO:` I am yet to add tests and documentation, and I can proceed to do so if my approach for the same aligns with your vision as well.

- `TODO:` I also did not use black for formatting and ruff to lint yet, because when I did so, it changed some code which was out of the scope for this feature and I wanted to keep the PR minimal. Let me know if I should proceed to run these as well.  

---

## 🔍 Why is this needed?

It's adding a feature to support video files present in AWS S3 buckets. This is also mentioned in the CONTRIBUTING.md as one of the suggested connectors.

---

## 🧪 How was it tested?

I've currently tested this using the hello-ocr repository manually with an s3 file in my bucket instead of local file using below command. As mentioned earlier, I'll proceed to add unit tests if my approach aligns.

```bash
AWS_PROFILE=<profile> openfilter run - VideoIn --sources 's3://<bucket_name>/test/hello.mov!loop' - filter_optical_character_recognition.filter.FilterOpticalCharacterRecognition --ocr_engine easyocr --forward_ocr_texts true - Webvis
```

---

## 🔗 Related Issues

NA

---

---

## ✅ Checklist

- [ ✅ ] I have read and agreed to the terms of the [LICENSE](../LICENSE)
- [ ✅  ] I have read the [CONTRIBUTING](../CONTRIBUTING.md) guide
- [ ✅  ] I have followed the [coding style](../CONTRIBUTING.md#coding-style)
- [ ✅  ] I have signed all commits in compliance with the DCO (`git commit -s`)
- [ ] I have added or updated **tests** as needed
- [ ] I have added or updated **documentation** as needed

> 📄 **DCO Reminder:**  
>
> ```bash
> ```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,6 +102,7 @@ Explore real-world examples covering:
 * Load balancing using multiple filter processes
 * Sending metrics to MQTT
 * Ephemeral side-channel processing
+* S3 integration for cloud video processing
 * Fully declarative + class-based configuration
 
 ➡️ See [`docs/overview.md`](https://github.com/PlainsightAI/openfilter/blob/main/docs/overview.md) for all examples.
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -461,6 +461,31 @@ Webvis video at `http://localhost:8000/yet_another`.
         )),
     ])
 
+## S3 Integration Example
+
+Process video files stored in Amazon S3 with OCR analysis. This example demonstrates downloading video from S3, processing with optical character recognition, and visualizing results. You can get a sample video file to test from `examples/hello-ocr/hello.mov`.
+
+```bash
+    AWS_PROFILE=<profile> openfilter run \
+      - VideoIn --sources 's3://<bucket_name>/test/hello.mov!loop' \
+      - filter_optical_character_recognition.filter.FilterOpticalCharacterRecognition --ocr_engine easyocr --forward_ocr_texts true \
+      - Webvis
+```
+
+Replace `<profile>` with your AWS profile name and `<bucket_name>` with your S3 bucket name. This example:
+
+* Downloads the video file from the specified S3 bucket using your AWS credentials
+* Processes each frame with OCR using EasyOCR to extract text content
+* Forwards the extracted text data along with the video frames
+* Displays both the video and OCR results in the browser at `http://localhost:8000`
+* The `!loop` option continuously replays the video for ongoing processing
+
+Prerequisites:
+* AWS credentials configured (via AWS CLI, environment variables, or IAM roles)
+* S3 bucket with read permissions for the specified video file
+* OpenFilter with S3 support: `pip install openfilter[all]`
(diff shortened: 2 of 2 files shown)


plainsight-M477:
@Ninad-Bhangui Thank you so much for your contribution!!

Generating presigned URLs is a very cool approach and it definitely aligns with our vision :smile: 

I've reviewed and tested the code so far and everything looks awesome! Good call on skipping black for this PR, I'll run it on a separate PR once your PR is merged to avoid expanding this one ^^

If you'd like to add the unit tests and update the documentation as well I would sincerely appreciate it, otherwise I can add the tests/documentation in my next PR :relaxed:

pyproject.toml
```diff
@@ -71,6 +71,7 @@ video = []
 video_in = [
   "setuptools==72.2.0",
   "vidgear[core]==0.3.3",
+  "boto3>=1.26.0",
```
plainsight-M477, line 74: Should we add this to the `all = [` dependency group as well? :smile:

Ninad-Bhangui: You're right. I added it to the all group as well.

Ninad-Bhangui:
@plainsight-M477 I did add some unit tests and updated the documentation.

## Documentation
I added a section in `overview.md`. I could add another folder like `hello-world-ocr-s3` but it would be 99% same with just s3 as a difference. So I wasn't sure if that's the approach I should take.

## Tests

I'm not sure how useful my current tests are. In some projects I've used https://www.localstack.cloud/ via docker for tests against AWS services but it's overkill for this project.

My other idea was to replicate `test_filter_video_in.py`. I like the test there as it works with an actual video `RED_THEN_GREEN_THEN_BLUE_FRAME_MP4` and calls an outermost function using an actual file without any mocks instead of calling internal functions which are implementation details. I thought I would mock the presigned url and replace it with a local http server serving that video but struggled to get it working.

Let me know if the current tests are enough.

plainsight-M477 (approved):
@Ninad-Bhangui Thank you so much!!

I agree that localstack tests are an overkill at this time- the current tests are definitely sufficient :blush: 

A `hello-world-ocr-s3` example would definitely add value even if it was 99% the same :smile: 
Although it will have to use the next release (`0.1.3`) as a dependency (since we're packaging the examples as self-contained examples which pull dependencies from Pypi)
You could either add it in this PR and i'll cut the release immediately once it's merged, or we can ship this PR -> cut the release -> then add the example in a follow up PR
lmk which route you prefer! ^^

The docs are fantastic! Thank you :bowing_man: :pray:

PR #14, opened 2025-07-11
Open Telemetry on OpenFilter

trodrigues138:
# 📦 Pull Request

---

## 📋 What does this PR do?

  - Collect runtime metrics (e.g. CPU, memory, FPS, latency) from each filter
  - Expose those metrics using OpenTelemetry instruments (`ObservableGauge`)
  - Export them periodically to external systems (Google Cloud Monitoring, OTLP, Console, etc.)
  - Aggregate metrics for each filter for  observability and health monitoring
  - Add pipeline_id to OpenFilter pipelines
  - Add device_name to OpenFilter so that the user may know which device is excuting the filter's pipeline
---

## 🔍 Why is this needed?

This is important for **monitoring pipeline performance**, **troubleshooting**, and **scaling decision-making** in  using OpenFilter.
---

## 🧪 How was it tested?

Running the demos from OpenFilter and connecting to the exportes
---

## 🔗 Related Issues

---

---

## ✅ Checklist

- [ ] I have added or updated **tests** as needed

> 📄 **DCO Reminder:**  
>
> ```bash
> ```

---


openfilter/filter_runtime/filter.py
```diff
@@ -357,9 +358,16 @@ def __init__(self,
     ):
         if not (config := simpledeepcopy(config)).get('id'):
             config['id'] = f'{self.__class__.__name__}-{rndstr(6)}'  # everything must hava an ID for sanity
-
+        
+        pipeline_id = config.get("pipeline_id")
+        self.device_id_name = config.get("device_name")
+        self.pipeline_id = pipeline_id  # se quiser guardar como atributo
```
navarmn:
```suggestion
        self.pipeline_id = pipeline_id  # to store as an attribute
```

openfilter/filter_runtime/filter.py
```diff
@@ -357,9 +358,16 @@ def __init__(self,
...
+       
         self.start_logging(config)  # the very firstest thing we do to catch as much as possible
-
+        enabled_otel_env = os.getenv("TELEMETRY_EXPORTER_ENABLED")
+        self.telemetry_enabled:bool =  bool(strtobool(enabled_otel_env)) if enabled_otel_env is not None else False
```
navarmn: This will fail if `enabled_otel_env` is set to an invalid string (e.g., "maybe"). You should catch ValueError here as well.

```suggestion
        try:
             self.telemetry_enabled: bool = bool(strtobool(enabled_otel_env)) if enabled_otel_env is not None else False
        except ValueError:
             logger.warning(f"Invalid TELEMETRY_EXPORTER_ENABLED value: {enabled_otel_env}. Defaulting to False.")
    self.telemetry_enabled = False
```

openfilter/filter_runtime/filter.py
```diff
@@ -403,6 +411,18 @@ def exit(self, reason: str | None = None, exc: BaseException | None = None):
...
         raise exc or Filter.Exit
+    def start_metrics_updater_thread(self):
+        interval = self.otel.export_interval_millis / 1000  
+        
+        def loop():
+            while not self.stop_evt.is_set():
+                try:
+                    self.otel.update_metrics(self.metrics, filter_name=self.filter_name)
+                except Exception as e:
+                    logger.error(f"[metrics_updater] erro ao atualizar métricas: {e}")
```
navarmn:
```suggestion
                    logger.error(f"[metrics_updater] error when trying to update metrics: {e}")
```

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -0,0 +1,47 @@
...
+
+        elif exporter_type == "otlp_grpc":
+            try:
+
+                return OTLPGrpcExporter(
+                    endpoint=kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT"),
+                    insecure=kwargs.get("insecure", os.getenv("OTLP_GRPC_ENDPOINT_SECURITY",True))
+                )
+            except Exception as e:
+                logging.error("Failed to set OTLP_GRPC exporter {e}")
```
navarmn: Missed `f-string`

```suggestion
                logging.error(f"Failed to set OTLP_GRPC exporter {e}")
```

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -0,0 +1,47 @@
...
+
+        elif exporter_type == "otlp_http":
+            try:
+                return OTLPHttpExporter(
+                    endpoint=kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_HTTP_ENDPOINT"),
+                    headers=kwargs.get("headers") or {}
+                )
+            except Exception as e:
+                logging.error("Failed to set OTLP_HTTP exporter {e}")
```
navarmn: Missed `f-string`
```suggestion
                logging.error(f"Failed to set OTLP_HTTP exporter {e}")
```

navarmn (changes requested):
The overall functionalities look good, but my major concerns are explained bellow. Point 1 is a NECESSARY FIX while points 2 and 3 should be tickets created in the backlog.

1) MUST FIX NOW
   
 Your code relies on using `from distutils.util import strtobool` which is deprecated in Python 3.12. See [PEP-632](https://peps.python.org/pep-0632/) and this [discussion](https://stackoverflow.com/questions/77941994/what-is-setuptools-alternative-to-the-deprecated-distutils-strtobool). Basically PEP 632’s migration table lists common helpers that did get replacements, but `distutils.util.strtobool` isn’t one of them. The PEP’s guidance for such cases is to “re-implement the functionality yourself”.

 That will be a major blocker for us to upgrade to Python 3.13 with this [open PR](https://github.com/PlainsightAI/openfilter/pull/13).

 So, I suggest that we use as a drop-in replacement for `strtobool` the following function:

```python
# inside of openfilter/filter_runtime/utils.py
def strtobool(val: str) -> bool:
    """Convert a string representation of truth to True/False.

    Accepts 'y', 'yes', 't', 'true', 'on', '1' as true values and
    'n', 'no', 'f', 'false', 'off', '0' as false values.
    Raises ValueError on anything else (matches distutils.behaviour).
    """
    val = val.lower()
    if val in ('y', 'yes', 't', 'true', 'on', '1'):
        return True
    if val in ('n', 'no', 'f', 'false', 'off', '0'):
        return False
    raise ValueError(f"invalid truth value {val!r}")


from distutils.util import strtobool                      # old
from openfilter.filter_runtime.utils import strtobool     # new
```


2) Not a problem, but a NICE TO HAVE for the future

Metrics Updater Thread:
- The updater thread in `start_metrics_updater_thread` is spawned as a daemon and never joined or explicitly stopped. If the object is deleted or the process exits, it could leave threads dangling.
- The method relies on `self.stop_evt `being set. If `self.stop_evt` is not properly managed, threads could leak or fail to terminate.


```python

# Add these instance variables in __init__:
self._metrics_updater_thread = None

# ....

# Modify start_metrics_updater_thread:
def start_metrics_updater_thread(self):
    interval = self.otel.export_interval_millis / 1000  

    def loop():
        while not self.stop_evt.is_set():
            try:
                self.otel.update_metrics(self.metrics, filter_name=self.filter_name)
            except Exception as e:
                logger.error(f"[metrics_updater] erro ao atualizar métricas: {e}")
            self.stop_evt.wait(interval)  

    # Store the thread handle
    self._metrics_updater_thread = threading.Thread(target=loop, daemon=True)
    self._metrics_updater_thread.start()

def stop_metrics_updater_thread(self):
    iif not getattr(self, "telemetry_enabled", False):
        # Telemetry not enabled, nothing to stop
        return
    if self._metrics_updater_thread is not None:
        self.stop_evt.set()
        self._metrics_updater_thread.join(timeout=5)
        self._metrics_updater_thread = None

# ....
# in the filter.py
# Example: Call stop_metrics_updater_thread in your object's cleanup logic, destructor, or exit method.
def exit(self, reason: str | None = None, exc: BaseException | None = None):
    # ... existing logic ...
    self.stop_metrics_updater_thread()
    # ... raise or return as before ...
```


- The thread handle is stored in `self._metrics_updater_thread`.
- A new method `stop_metrics_updater_thread` is added to join the thread cleanly.
- The thread checks for `self.stop_evt` as before, but now can be joined.
- You should ensure that `self.stop_evt.set()` is called during object shutdown, and call `self.stop_metrics_updater_thread()` in your cleanup code or destructor.

3) NICE TO HAVE for the future

The factory attempts to construct exporters for GCM (Google Cloud Monitoring), OTLP (OpenTelemetry Protocol), and Console. It does not verify if required environment variables, credentials, or configurations are present before creating the exporter. If credentials or configs are missing, it may raise exceptions, but the handling is limited to logging—there’s no fallback or explicit user feedback.

Suggestion: Add Pre-Checks for Critical Environment Variables

```python
if exporter_type == "gcm":
    project_id = kwargs.get("project_id") or os.getenv("PROJECT_ID")
    if not project_id:
        logging.error("GCM exporter requires PROJECT_ID to be set in environment or kwargs.")
        raise ValueError("Missing PROJECT_ID for GCM exporter.")
    return CloudMonitoringMetricsExporter(project_id=project_id)

# ...

elif exporter_type == "otlp_grpc":
    endpoint = kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
    if not endpoint:
        logging.error("OTLP_GRPC exporter requires 'endpoint' to be set in environment or kwargs.")
        raise ValueError("Missing endpoint for OTLP_GRPC exporter.")
    # ... rest of exporter construction ...
```

tagging @plainsight-M477 for visibility as well.

PR #51, opened 2026-02-24
feat: patch filter Dockerfiles to build from source with uv

stwilt:
## Summary

- At cascade build time, patches each filter's Dockerfile to install from the cloned source with `uv` instead of fetching a pinned version from private PyPI
- Eliminates the "VERSION bumped but not published" failure class (e.g. filter-chatgpt-annotator 0.1.2, filter-optical-character-recognition 0.1.7)
- Handles three Dockerfile patterns: `RUN --mount=type=bind` (dominant), simple `RUN pip install`, and already-uses-uv (skipped)

## Open questions

This is speculative — we probably don't want to make a habit of maintaining filter build behavior in the cascade repo instead of in the constituent filter repos themselves.

A lighter-touch alternative: just clone the noncompliant repos that still pull from our private PyPI, and individually patch *them* to use uv + source install in their own Dockerfiles. That way the build contract stays with the filter, and the cascade doesn't need to know about Dockerfile internals at all.

Raising this PR to show the approach works and get feedback on which direction we prefer.

## Depends on

- #49 (cascade build infrastructure)

## Test plan

- [x] Tested locally against `filter-stub-application` (RUN --mount + inline comment pattern)
- [x] Tested locally against `filter-crop` (multi-stage build, RUN --mount pattern)
- [x] Tested locally against `filter-aggregator` (simple pip install pattern)
- [x] `filter-json-transform` (already uses uv) correctly skipped


Makefile
```diff
@@ -39,3 +43,51 @@ clean:  ## Delete all generated files and directories
...
+		--project=$(GCP_PROJECT) \
+		--service-account=projects/$(GCP_PROJECT)/serviceAccounts/$(CLOUDBUILD_SA) \
+		--substitutions=TAG_NAME=$(VERSION),_DRY_RUN=false \
+		.
+
+cloud.cascade.local: ## Run cascade build-filters script locally (dry-run)
+	@WORKSPACE=$$(mktemp -d) && \
+	echo "Workspace: $$WORKSPACE" && \
+	mkdir -p "$$WORKSPACE/filters" "$$WORKSPACE/results" && \
+	echo "$(subst v,,$(VERSION))" > "$$WORKSPACE/openfilter_version" && \
```
lucasmundim, line 73: **Bug:** `$(subst v,,$(VERSION))` strips **all** occurrences of `v` in the string, not just a leading prefix. If the version ever contained `v` elsewhere (e.g., a hypothetical `1.2.0-preview`), this would produce `1.2.0-preiew`.

Every other place in this PR correctly uses `sed 's/^v//'` to strip only the prefix. Use `patsubst` for consistency:

```makefile
echo "$(patsubst v%,%,$(VERSION))" > "$$WORKSPACE/openfilter_version"
```

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+  docker buildx use multiarch
+else
+  docker buildx create --name multiarch --driver docker-container --use
+fi
+docker buildx inspect --bootstrap
+
+# Extract PAT from secret (may be JSON {"password":"ghp_..."} or bare token)
+GH_PAT=$(python3 -c "
+import json, sys
+raw = '''${GITHUB_TOKEN}'''
```
lucasmundim, line 76: **Security hygiene:** `GITHUB_TOKEN` is interpolated directly into a Python triple-quoted string via shell expansion. If the token ever contained `'''` (unlikely for a GitHub PAT, but possible for other secret formats like wrapped JSON), it would break out of the string.

Safer to pass the token via stdin:

```bash
GH_PAT=$(echo "${GITHUB_TOKEN}" | python3 -c "
import json, sys
raw = sys.stdin.read().strip()
try:
    data = json.loads(raw)
    print(data.get('password', raw))
except (json.JSONDecodeError, TypeError):
    print(raw)
")
```

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+
+  # Pre-generate ADC credentials for filter_base builds (shared by all workers).
+  # Done once here to avoid concurrent gcloud calls from parallel builds.
+  CRED_FILE="${WORKSPACE}/google_credentials.json"
+  gcloud auth application-default print-access-token &>/dev/null && \
+    cp "$(gcloud info --format='value(config.paths.global_config_dir)')/application_default_credentials.json" "${CRED_FILE}" 2>/dev/null || \
+    python3 -c "
+import json, subprocess, sys
+token = subprocess.check_output(['gcloud','auth','print-access-token'], text=True).strip()
+json.dump({'type':'authorized_user','token':token}, sys.stdout)
```
lucasmundim, line 61: **Nit:** A standard `authorized_user` ADC credential requires `client_id`, `client_secret`, and `refresh_token` fields — not a bare `token`. If any downstream code uses the Google Auth library to load this file (rather than extracting the token directly), it will fail with a confusing error.

If this is intentionally only consumed as a raw build secret by the filter_base ONBUILD, a comment explaining the format would help future readers.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+    NEXT_URL=$(grep -oi '<[^>]*>; rel="next"' /tmp/gh_headers | sed 's/<\(.*\)>; rel="next"/\1/' || true)
+  done
+  if [[ "${FETCH_OK}" == "true" ]] && [[ -s "${VISIBILITY_FILE}" ]]; then
+    echo "Fetched visibility for $(wc -l < "${VISIBILITY_FILE}") repos"
+    break
+  fi
+  echo "WARNING: GitHub API fetch failed (attempt ${ATTEMPT})"
+  sleep 5
+done
+if [[ ! -s "${VISIBILITY_FILE}" ]]; then
```
lucasmundim, line 133: **Resilience concern:** If the GitHub API is completely unavailable (all 3 retries fail), the cascade silently falls back to building only 2 hardcoded filters. In a production release scenario, this could mean silently skipping 36+ filters.

Consider making this a hard failure for non-dry-run builds:

```bash
if [[ ! -s "${VISIBILITY_FILE}" ]]; then
  if [[ "${DRY_RUN}" != "true" ]]; then
    echo "ERROR: GitHub API unavailable — cannot discover filters for live build"
    exit 1
  fi
  echo "WARNING: GitHub API unavailable, falling back to hardcoded filter list"
  ...
fi
```

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+      git clone --depth 1 "https://github.com/PlainsightAI/${REPO}.git" "${REPO}" 2>&1
+    fi
+    echo "Cloned ${REPO}"
+  ) &
+  CLONE_PIDS+=($!)
+done
+# Wait for all clones, note failures
+CLONE_FAILED=""
+for i in "${!CLONE_PIDS[@]}"; do
+  REPO=$(echo "${FILTER_REPOS}" | tr ' ' '\n' | sed -n "$((i+1))p")
```
lucasmundim, line 201: **Fragile PID-to-repo mapping:** This relies on `FILTER_REPOS` being a space-separated string where positional indexing via `sed -n` matches the PID array. If the string has leading/trailing whitespace or double spaces, `sed -n` could return the wrong repo name.

A more robust approach would be a parallel array built at clone time:

```bash
CLONE_PIDS=()
CLONE_REPOS=()
for REPO in ${FILTER_REPOS}; do
  ( ... ) &
  CLONE_PIDS+=($!)
  CLONE_REPOS+=("${REPO}")
done

for i in "${!CLONE_PIDS[@]}"; do
  REPO="${CLONE_REPOS[$i]}"
  ...
done
```

lucasmundim (commented):
Comprehensive review of the cascade build infrastructure + Dockerfile source-install patching.

The architecture is well-thought-out — auto-discovery, constraint checking, parallel builds, dry-run safety, and the Dockerfile patching approach are all clever solutions to real problems. The PR description is honest about the speculative nature and invites feedback on direction.

7 inline comments below, ranging from a bug to design suggestions.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+            blocks.append((start, i, ''.join(lines[start:i+1])))
+        i += 1
+    return blocks
+
+blocks = collect_run_blocks(lines)
+
+# Find the RUN block that installs the filter package
+target_block = None
+for start, end, text in blocks:
+    if re.search(r'pip install', text, re.IGNORECASE) and re.search(r'filter', text, re.IGNORECASE):
```
lucasmundim, line 357: **Broad heuristic:** This matches any RUN block containing both `pip install` and `filter` (case-insensitive), which could match unrelated instructions. For example, a comment like `# Filter out unused packages` near a `pip install` of something else would trigger the match.

The fallback on line 364 then matches *any* `pip install` block at all.

Given that these are all `filter-*` repos with known structures, this is probably fine in practice, but it's a latent risk if Dockerfiles evolve. Consider matching the actual package name (e.g., `re.search(r'pip install.*filter[-_]', text)`) for a tighter heuristic.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+        out.append('RUN uv pip install --system --no-cache /src && rm -rf /src\n')
+        i = pip_end + 1
+        continue
+
+    out.append(lines[i])
+    i += 1
+
+with open(path, 'w') as f:
+    f.writelines(out)
+print("Patched Dockerfile: source install with uv")
```
lucasmundim, line 406: **No validation of patched Dockerfile:** The `PATCH_SCRIPT` rewrites the Dockerfile in place but never verifies the result. If the heuristics misfire (wrong FROM stage, missing RUN block), the build fails with a confusing Docker error rather than a clear "patch failed" message.

A lightweight sanity check after patching could help:

```python
# Verify patch produced valid-looking output
with open(path) as f:
    patched = f.read()
if 'FROM' not in patched:
    print('ERROR: patched Dockerfile has no FROM', file=sys.stderr)
    sys.exit(1)
if 'COPY --from=ghcr.io/astral-sh/uv' not in patched and 'uv pip install' in patched:
    print('ERROR: uv install without uv COPY', file=sys.stderr)
    sys.exit(1)
```

shingonoide (commented):
Complementary to the existing review, a few build-correctness items on the uv patch path (inline). Also flagging that DCO is red (missing Signed-off-by), which will block merge.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+        out.append(lines[i])
+        out.append('COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/\n')
+        uv_injected = True
+        i += 1
+        continue
+
+    # Replace the target pip install block
+    if i == pip_start:
+        out.append('COPY . /src\n')
+        out.append('RUN uv pip install --system --no-cache /src && rm -rf /src\n')
```
shingonoide, line 397: This install resolves openfilter from each filter's own pyproject specifier against the live index, with no pin to `${OF_VERSION}`. The image is still tagged `-of${OF_VERSION}` (line 416), and since the fleet uses wide ranges (`>=`, `~=`), a build can silently resolve an older published openfilter than the one that triggered the cascade and tag it as the new runtime. The release workflow also creates the trigger tag before publish-to-pypi finishes, so the new wheel may not be on the index yet when resolution runs. Please pin the cascaded version explicitly: install `openfilter==${OF_VERSION}` alongside the source, or thread it in as a build arg and constrain resolution.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+    if i == from_line_idx and not uv_injected:
+        out.append(lines[i])
+        out.append('COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/\n')
+        uv_injected = True
+        i += 1
+        continue
+
+    # Replace the target pip install block
+    if i == pip_start:
+        out.append('COPY . /src\n')
```
shingonoide, line 396: `COPY . /src` followed by `rm -rf /src` in the next RUN does not reclaim space: COPY is its own layer, so the full build context (anything not excluded by `.dockerignore`) persists in the image. Prefer `RUN --mount=type=bind,source=.,target=/src uv pip install --system --no-cache /src` so the source never lands in a layer.

scripts/build-filters.sh
```diff
@@ -0,0 +1,566 @@
...
+
+# Rebuild the Dockerfile
+out = []
+uv_injected = False
+i = 0
+while i < len(lines):
+    # Inject uv COPY after the FROM line for the pip install stage
+    if i == from_line_idx and not uv_injected:
+        out.append(lines[i])
+        out.append('COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/\n')
```
shingonoide, line 389: uv is unpinned here (`:latest`) and again in the host bootstrap at line 96 (`curl ... astral.sh/uv/install.sh`). Docker and buildx are pinned (lines 26-27); pinning uv too keeps builds reproducible.

lmyers-ps:
Closing this PR as obsolete.                                                                                                                  
Since this draft was opened, the cascading release architecture has migrated to a distributed, event-driven model under **DT-145** (merged in  **PR #85**).                                                                                                                                    
                                                                                                                                                  
Instead of maintaining a centralized Google Cloud Build pipeline (`cloudbuild-cascade.yaml`) that clones and dynamically patches downstream filter `Dockerfiles`, releases on `main` now trigger `cascade-on-tag.yaml`. This dispatches individual mechanical bump PRs to each filter repository, letting them build and publish their own Docker images through their own native CI/CD pipelines.                                    
                                                                                                                                                  
Because the centralized build and patching scripts have been fully removed from `main`, the changes here are no longer required.

PR #61, opened 2026-03-25
feat: frame accumulation support for batched processing

stwilt:
## What does this PR do?

Adds optional frame accumulation to the Filter base class. When `batch_size > 1`, the runtime buffers N incoming frames and delivers them to `process_batch()` instead of calling `process()` individually. Flushes on timeout if frames arrive slower than the batch window.

- New config: `batch_size` (default 1), `accumulate_timeout_ms` (default 100)
- Default `process_batch()` calls `process()` per frame (backward compatible)
- Subclasses override `process_batch()` for actual batched inference

## Why is this needed?

GPU-bound filters have expensive operations (e.g. vision backbone passes) that can be amortized across multiple frames via batched tensor operations. This provides the framework hook for that optimization.

## How was it tested?

24 unit tests covering: config validation, accumulation logic, timeout flush, thread safety, backward compatibility, string coercion.

## Related Issues

- Companion PR in a downstream filter repo validates the `process_batch()` override path end-to-end.

## Checklist


lucasmundim (commented):
### Code review

Found 1 issue:

1. `_execute_batch` silently discards all batch results except the last non-`None` one. For a batch of N frames, frames 0 through N-2 are never sent downstream — this is data loss. The `process_batch()` docstring states "List of results, one per input frame", and the `docs/overview.md` example shows `return [self.decode(f, feat) for f, feat in zip(batch, features)]`, both implying every result should be forwarded. The implementation contradicts this contract: only `last_result` is emitted via `mq.send`, and only `last_result` receives timing metadata via `_inject_timings`.

https://github.com/PlainsightAI/openfilter/blob/f0a62f2470b6b1f7bd22f50f1033b8386ceb7c25/openfilter/filter_runtime/filter.py#L1239-L1261

openfilter/filter_runtime/filter.py
```diff
@@ -270,11 +329,14 @@ def _read_models_toml():
         except Exception as e:
             logger.error(f"Error reading models.toml: {e}")
             return {}
-        
+
+
 class Filter:
+    _batch_size: int = 1
```
lucasmundim: `_batch_size: int = 1` is placed before the class docstring, which means Python won't recognize it as a docstring — `Filter.__doc__` will be `None`. Move this below the docstring alongside the other class-level attributes (`FILTER_TYPE`, `_EMA_ALPHA`, etc.).

openfilter/filter_runtime/filter.py
```diff
@@ -766,345 +899,491 @@ def download_cached_files(config: FilterConfig):
...
+                last_result = {"main": result}
+            else:
+                last_result = result
+
+        if last_result is None:
+            if self._is_last_filter and batch:
+                self._inject_timings(batch[-1], t_in, t_out, process_time_ms)
+            return None
+
+        self._inject_timings(last_result, t_in, t_out, process_time_ms)
```
lucasmundim: `_inject_timings` and `_update_process_time_ema` are called once for the entire batch and only applied to the last result. For a batch of N frames, N-1 frames receive no `filter_timings` metadata. This breaks the invariant from PR #52 where every frame carries timing metadata for the dashboard panels.

docs/overview.md
```diff
@@ -55,6 +55,25 @@ Finally, the `process()` function will be called on each frame or multiple frame
...
 Other functions you may find useful to overload is `shutdown()`, this is called on exit (clean or exception) and lets you clean up after yourself.
 
+### Frame Accumulation (Batched Processing)
+
+When `batch_size` is set to a value greater than 1 in the filter config, the runtime accumulates incoming frames and delivers them as a list to `process_batch()` instead of calling `process()` individually. This allows GPU-bound filters to amortize expensive operations (e.g. backbone inference) across multiple frames.
+
+    class MyGPUFilter(Filter):
+        def process_batch(self, batch):
+            # batch is a list of frame dicts, each as would be passed to process()
+            features = self.model.encode_batch([extract_image(f) for f in batch])
```
lucasmundim: `extract_image(f)` and `self.decode(f, feat)` are not part of the openfilter API. Frame images are accessed via `Frame.img` or by indexing the frame dict. Consider using real API calls or marking this as pseudocode.

openfilter/filter_runtime/filter.py
```diff
@@ -766,345 +899,491 @@ def download_cached_files(config: FilterConfig):
...
+                if isinstance(processed_frames, Frame)
+                else processed_frames
+            )
 
-            # Inject timing metadata into output frames
             self._inject_timings(final_frames, t_in, t_out, process_time_ms)
 
             try:
                 self._metadata_queue.put_nowait((final_frames, self.emitter))
             except queue.Full:
```
lucasmundim: Several load-bearing comments were removed during the refactor from `process_frames` into `_process_frames_single`. In particular: "Update per-filter timing unless process() returned a callable, in which case `_timed_callable` will record the real duration later" — this explains a non-obvious conditional. Also `# metadata is best-effort; drop if queue is full` on the bare `pass` in the `except queue.Full` block (line 1182).

openfilter/filter_runtime/filter.py
```diff
@@ -766,345 +899,491 @@ def download_cached_files(config: FilterConfig):
...
+            if batch_mode
+            else POLL_TIMEOUT_MS
+        )
+
+        while (frames := self.mq.recv(min(poll_ms, sources_timeout))) is None:
             if self.stop_evt.is_set():
                 self.exit()
 
-            if (sources_timeout := sources_timeout - POLL_TIMEOUT_MS) <= 0:
+            if batch_mode and self._batch_flush_needed:
```
lucasmundim: `self._batch_flush_needed` is read outside `_batch_lock` here, but set under the lock in `_on_batch_timeout`. This is a data race. CPython's GIL makes it unlikely to cause issues in practice, but consider using the lock or a `threading.Event` for correctness.

lucasmundim (commented):
Additional findings from code review (lower confidence, flagged for awareness):

openfilter/filter_runtime/filter.py
stwilt: Fixed — moved `_batch_size` below the docstring alongside `FILTER_TYPE` and `_EMA_ALPHA` at [`openfilter/filter_runtime/filter.py` L615](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L615).

openfilter/filter_runtime/filter.py
stwilt: Fixed — `_execute_batch` now injects timing into every result ([`openfilter/filter_runtime/filter.py` L1283-L1289](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L1283-L1289)) and returns all normalized results as a list. Callers send each one downstream individually. Also uses the full batch `process_time_ms` per frame since batch processing is indivisible.

docs/overview.md
stwilt: Fixed — replaced with real API: `f["main"].img` and `self.process(f)` at [`docs/overview.md` L65-L67](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/docs/overview.md#L65-L67).

openfilter/filter_runtime/filter.py
stwilt: Restored both comments: "Update per-filter timing unless process() returned a callable..." at [`openfilter/filter_runtime/filter.py` L1159-L1160](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L1159-L1160), and "metadata is best-effort; drop if queue is full" at [`openfilter/filter_runtime/filter.py` L1189](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L1189).

openfilter/filter_runtime/filter.py
stwilt: Fixed — replaced the `bool` flag with `threading.Event` ([`openfilter/filter_runtime/filter.py` L683](https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L683)). `is_set()`/`set()`/`clear()` are thread-safe. All accesses to `_frame_buffer` remain under `_batch_lock`.

lucasmundim (dismissed):
### Code review

Found 1 issue:

1. `process()` docstring states it is "the only method which MUST be implemented by a user Filter," but with the new `process_batch()` override point, this is no longer accurate. A user who overrides only `process_batch()` (without implementing `process()`) will have a working filter in batch mode — but if `batch_size` is ever set to `1` (or omitted, since the default is `1`), the runtime falls through to `_process_frames_single` → `self.process()` → `NotImplementedError`. The docstring should clarify that `process()` is still required unless the filter always runs with `batch_size > 1`, or the default `process_batch()` fallback should handle the case where `process()` is not implemented.

https://github.com/PlainsightAI/openfilter/blob/026a6571e10633420308d1246001a5afc5560a2f/openfilter/filter_runtime/filter.py#L1634-L1695

lucasmundim (commented):
Well-structured feature with a solid backward compatibility story. The core design (dispatch, locking, timer-to-event pattern) is sound. Previous review feedback has been thoroughly addressed. The remaining issues are minor — the `fini()` lock gap is the most notable but is low-risk in practice. Looks good to merge once CI completes.

### Previous review feedback resolution

| Issue | Status |
|---|---|
| `_batch_size` before docstring breaks `Filter.__doc__` | **Fixed** — moved after docstring alongside `FILTER_TYPE`, `_EMA_ALPHA` |
| Data loss: only last batch result sent downstream | **Fixed** — `_execute_batch()` returns all normalized results, `_batch_pending_results` stashes N-1 for `loop_once` to drain |
| Pseudocode in docs | **Fixed** — real openfilter API calls |
| `_batch_flush_needed` bool data race | **Fixed** — replaced with `threading.Event` |
| `process()` docstring inaccuracy | **Fixed** — docstring now clarifies when `process()` is required vs optional |

### Test coverage gaps (non-blocking)

- No test for the `fini()` partial-batch-on-shutdown flush path
- No test for `_batch_pending_results` draining in `loop_once` (the mechanism that prevents the N-1 data loss from the first review round)

openfilter/filter_runtime/filter.py
```diff
@@ -1198,82 +1545,110 @@ def on_exit_msg(reason: str):
...
+        if self.telemetry_enabled and hasattr(self, "otel"):
             self.start_metrics_updater_thread()
 
         # Start metadata worker for processing frame metadata asynchronously
         self._start_metadata_worker()
 
     def fini(self):
         """Shut down inter-filter communication and any other system level stuff."""
+        self._cancel_batch_timer()
+        if self._frame_buffer:
```
lucasmundim: `_frame_buffer` is accessed here without acquiring `_batch_lock`. While practically safe (main loop has exited by this point), this is inconsistent with the locking discipline in every other access path (`_process_frames_batched`, `_take_flush_batch`, `loop_once` timeout flush). If a late timer callback fires between `_cancel_batch_timer()` and the buffer check, it could set the flush event on a cleared buffer — harmless, but wrapping this in `with self._batch_lock:` would be more defensive and consistent.

```python
with self._batch_lock:
    if self._frame_buffer:
        batch = list(self._frame_buffer)
        self._frame_buffer.clear()
    else:
        batch = None
if batch is not None:
    try:
        results = self._execute_batch(batch)
        for r in results:
            self.mq.send(r, POLL_TIMEOUT_MS)
    except Exception:
        logger.warning("Failed to flush partial batch on shutdown")
```

openfilter/filter_runtime/filter.py
```diff
@@ -1000,54 +1185,186 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+                    return
+
+            if (sources_timeout := sources_timeout - poll_ms) <= 0:
                 frames = {}
 
                 break
 
+        # Flush any partial batch before processing the timeout empty-frames
+        if batch_mode:
+            with self._batch_lock:
```
lucasmundim, line 1398: nit: When `frames` arrives as `{}` (sources timeout) and `batch_mode` is true, this block flushes the buffer, then execution continues to `process_frames({})` which hits `_process_frames_batched({})` → returns `None` immediately (empty frames check). Works correctly but the buffer was already flushed here, so the subsequent `process_frames` call is redundant. Not a bug, just a minor inefficiency on the timeout path.

lucasmundim (commented):
One more observation I missed in the previous review:

**Formatting noise in diff**: A large portion of the +941/-397 is formatter changes (single→double quotes, alignment, parenthesization). The actual feature logic is ~150-200 lines. Running the formatter as a separate commit would make the feature diff much easier to review and bisect in the future.

openfilter/filter_runtime/filter.py
```diff
@@ -17,133 +17,189 @@
...
+    JSONType,
+    json_getval,
+    simpledeepcopy,
+    dict_without,
+    split_commas_maybe,
+    rndstr,
+    timestr,
+    parse_time_interval,
+    parse_date_and_or_time,
+    hide_uri_users_and_pwds,
```
lucasmundim, line 30: The vast majority of the diff in this file (~560 of ~960 added lines) is formatting-only changes: single→double quotes, alignment removal, Black-style line wrapping. This buries the actual feature and makes the PR very difficult to review.

Please consider splitting the formatter pass into a separate PR so that this one only contains the frame accumulation logic.

docs/overview.md
```diff
@@ -55,6 +55,29 @@ Finally, the `process()` function will be called on each frame or multiple frame
 
 Other functions you may find useful to overload is `shutdown()`, this is called on exit (clean or exception) and lets you clean up after yourself.
 
+### Frame Accumulation (Batched Processing)
+
+When `batch_size` is set to a value greater than 1 in the filter config, the runtime accumulates incoming frames and delivers them as a list to `process_batch()` instead of calling `process()` individually. This allows GPU-bound filters to amortize expensive operations (e.g. backbone inference) across multiple frames.
+
+    class MyGPUFilter(Filter):
```
lucasmundim, line 62: Two issues in this example:

1. **Indentation bug** — `images` is indented 13 spaces but `detections` on the next line is 12 spaces. This would be an `IndentationError` if anyone copies it.
2. **Wrong attribute** — `f["main"].image` should be `f["main"].img` per the Frame API. (This was flagged in the previous review round and marked as "Fixed" but still appears in the diff.)

tests/test_cli.py
```diff
@@ -9,6 +9,14 @@
...
 
 from openfilter.filter_runtime.filter import Filter
+
+
+def _has_vidgear():
+    try:
+        import vidgear  # noqa: F401
+        return True
+    except ImportError:
+        return False
```
lucasmundim: This `_has_vidgear()` helper + skip decorator and the `shutil.copy2` change in `test_filter.py` are unrelated to frame accumulation. Consider moving them to a separate PR to keep this one focused.

lucasmundim (changes requested):
## Re-review

**Merge conflict**: this PR is currently in a `CONFLICTING` state and cannot be merged. Please rebase on `main` and resolve the conflicts.

The frame accumulation feature itself is well-designed (clean separation, proper locking, timer-based flush, solid backward compat and tests). The main concerns are around PR hygiene — see inline comments.

shingonoide (dismissed):
LGTM — clean design, backward compatible, strong test coverage (24 tests).

**Suggestions (non-blocking):**

1. **Formatting noise** — ~90% of the diff is Black reformatting (quotes, wrapping). Separating formatting into its own commit would make review and future `git blame` much easier.

2. **`fini()` flush swallows exceptions** — The `except Exception` logs a warning but drops the exception type/message. Adding `exc_info=True` to the log call would help debugging.

3. **`_execute_batch` length mismatch at DEBUG level** — When `process_batch()` returns a different count than the input batch, this is likely a subclass bug. Consider WARNING instead of DEBUG.

4. **Timer thread churn** — `threading.Timer` spawns a new thread per timeout. For high-throughput filters, a single background watcher thread would be more efficient. Fine for now.

shingonoide (commented):
Clean design, backward compatible, strong test coverage (24 tests).

**Suggestions:**

1. **Formatting noise** — ~90% of the diff is Black reformatting (quotes, wrapping). Separating formatting into its own commit would make review and future `git blame` much easier.

2. **`fini()` flush swallows exceptions** — The `except Exception` logs a warning but drops the exception type/message. Adding `exc_info=True` to the log call would help debugging.

3. **`_execute_batch` length mismatch at DEBUG level** — When `process_batch()` returns a different count than the input batch, this is likely a subclass bug. Consider WARNING instead of DEBUG.

4. **Timer thread churn** — `threading.Timer` spawns a new thread per timeout. For high-throughput filters, a single background watcher thread would be more efficient. Fine for now.

openfilter/filter_runtime/filter.py
stwilt: Already addressed in the previous round — `fini()` wraps the buffer access in `with self._batch_lock:`. See [`filter.py` L1688-L1696](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1688-L1696). Also added `exc_info=True` to the warning per @shingonoide's suggestion at [L1704](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1704).

openfilter/filter_runtime/filter.py
stwilt: Fixed — the timeout flush path now returns early after flushing, skipping the redundant `process_frames({})` call. See [`filter.py` L1376-L1390](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1376-L1390).

openfilter/filter_runtime/filter.py
stwilt: Split into a dedicated commit: [d6d5794 style: apply Black formatting to filter.py](https://github.com/PlainsightAI/openfilter/commit/d6d5794). The feature commit ([326fbd3](https://github.com/PlainsightAI/openfilter/commit/326fbd3)) now only contains the frame accumulation logic.

docs/overview.md
stwilt: Both fixed. See [`docs/overview.md` L62-L70](https://github.com/PlainsightAI/openfilter/blob/e7730a6/docs/overview.md#L62-L70) — consistent 12-space indentation and `.img` throughout.

tests/test_cli.py
stwilt: Reverted — both `test_cli.py` and `test_filter.py` are now unchanged from main. Note that `vidgear` was removed as a dependency but these test references were never cleaned up — could you track the vidgear test cleanup as a follow-up?

stwilt:
Rebased on main, conflicts resolved. Dropped the version bump commit (`v0.1.25`) since versioning is handled at release time.

Branch now has 3 clean commits:
- d6d5794 — formatter-only pass (Black)
- 326fbd3 — frame accumulation feature
- e7730a6 — replace Timer thread churn with single watcher thread

@shingonoide — all four suggestions addressed:

1. **Formatting noise** — split into dedicated commit [d6d5794](https://github.com/PlainsightAI/openfilter/commit/d6d5794)
2. **`fini()` flush `exc_info`** — added at [`filter.py` L1704](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1704)
3. **Batch length mismatch log level** — promoted to `WARNING` at [`filter.py` L1305](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1305)
4. **Timer thread churn** — replaced `threading.Timer` with a single long-lived daemon watcher thread ([e7730a6](https://github.com/PlainsightAI/openfilter/commit/e7730a6), [`filter.py` L1852-L1886](https://github.com/PlainsightAI/openfilter/blob/e7730a6/openfilter/filter_runtime/filter.py#L1852-L1886))

lucasmundim (commented):
### Code review

Found 2 issues:

1. `_execute_batch` logs batch size mismatch at DEBUG instead of WARNING. When `process_batch()` returns a different number of results than input frames, this is silently logged at DEBUG level. Since this indicates a subclass programming error and potential data loss, it should be at least `logger.warning`. This was also flagged in a previous review by shingonoide.

https://github.com/PlainsightAI/openfilter/blob/59f135f2fb091f0d253c1d68346f8b6e714a27c6/openfilter/filter_runtime/filter.py#L1264-L1270

2. VERSION and RELEASE.md are stale. The PR bumps VERSION to `v0.1.25`, but `main` is already at `v0.1.27`. Merging as-is would regress the version. The branch needs a rebase and the version/release notes need to target the next available version.

https://github.com/PlainsightAI/openfilter/blob/59f135f2fb091f0d253c1d68346f8b6e714a27c6/VERSION#L1

openfilter/filter_runtime/filter.py
```diff
@@ -982,72 +1229,192 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+        Returns a list of frame dicts (possibly empty). The caller is
+        responsible for sending each result downstream.
+        """
+        t_in = time.time()
+        results = self.process_batch(batch)
+        if results is None:
+            results = [None] * len(batch)
+        t_out = time.time()
+        if len(results) != len(batch):
+            logger.warning(
```
stwilt, line 1318: Already fixed — this is `logger.warning` as of e7730a6. The commit you linked (59f135f2) is no longer the branch HEAD; the branch was rebased.

stwilt:
Re: VERSION/RELEASE.md being stale — bumped VERSION to `v0.1.28` and added a release entry. The old version bump commit (`v0.1.25`) was dropped during the rebase since main had moved to `v0.1.27`.

- [`VERSION` L1](https://github.com/PlainsightAI/openfilter/blob/cdc64d9/VERSION#L1)
- [`RELEASE.md` L5-L10](https://github.com/PlainsightAI/openfilter/blob/cdc64d9/RELEASE.md#L5-L10)

lucasmundim (commented):
### Code review

Found 1 issue:

1. `_execute_batch` injects timing metrics only on `batch[-1]` when all results are `None` (batch sink filter). The single-frame path (`_process_frames_single`) injects timing per frame, so EMA aggregates (`_frame_total_time_ema`, etc.) update once per frame. The batch path updates them once per entire batch, causing the EMA to converge ~N times slower at the same effective frame rate. To match the single-frame contract established in PR #52, timing should be injected for each frame in the batch.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1324-L1328

lucasmundim (commented):
### Code review

Found 7 issues:

1. `_execute_batch` injects timing metrics only on `batch[-1]` when all results are `None` (batch sink filter). The single-frame path injects timing per frame, so EMA aggregates update once per frame. The batch path updates them once per entire batch, causing the EMA to converge ~N times slower. Breaks the per-frame timing contract from PR #52.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1324-L1328

2. Batch timer reset is ineffective mid-sleep. Once the watcher thread passes `_batch_wakeup.wait()` and enters `_batch_stop.wait(timeout)`, calling `_reset_batch_timer()` sets `_batch_wakeup` but the watcher is not listening on it. The timeout fires based on when the first frame arrived, not the most recent. With frames arriving every 90ms, `batch_size=4`, and `accumulate_timeout_ms=100`, the batch flushes as 1-frame partial batches instead of accumulating.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1455-L1470

3. `sources_timeout` is decremented by `poll_ms` (which can be as small as `accumulate_timeout_ms`) instead of tracking actual elapsed time. If `accumulate_timeout_ms=10ms` and `POLL_TIMEOUT_MS=50ms`, and `mq.recv` returns faster than `poll_ms`, `sources_timeout` expires prematurely relative to real wall time.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1349-L1371

4. `_batch_pending_results` is assigned (not appended) in `_process_frames_batched`. The timeout-flush branch in `loop_once` returns early without draining pending results. If a timeout flush and a full-batch completion happen in sequence, the next assignment silently overwrites undrained results, dropping frames.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1280-L1283

5. `fini()` flushes `_frame_buffer` (unprocessed frames) but does not drain `_batch_pending_results` (already-processed results). If shutdown occurs between batch completion and the next `loop_once` drain, those processed frames are silently dropped.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1690-L1706

6. The default `process_batch()` eagerly resolves Callable returns from `process()` with `r = r()`. The `process()` docstring states Callables are called "AT THE POINT WHEN IT IS REQUESTED by ALL downstream clients" (lazy evaluation for freshest frames). This contract is silently broken in batch mode with no runtime warning. The `process_batch()` docstring notes "Callables are NOT supported" but `process()` itself has no indication.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/openfilter/filter_runtime/filter.py#L1421-L1432

7. RELEASE.md refers to `_batch_size` (private attribute) instead of `batch_size` (public config key), inconsistent with all docs, examples, and tests.

https://github.com/PlainsightAI/openfilter/blob/cdc64d977597abc694f11813a4e039d8781fa66d/RELEASE.md#L1-L5

stwilt:
All 7 points addressed in 1c7530b.

1. **Sink filter per-frame timing** — `_execute_batch` now injects timing into every input frame when all results are `None`, matching `_process_frames_single` EMA behavior. See [`openfilter/filter_runtime/filter.py` L1325-L1329](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1325-L1329).

2. **Batch timer reset mid-sleep** — Restructured `_batch_watcher_loop` to re-check `_batch_wakeup` after each timeout sleep. If a new frame arrived mid-sleep (via `_reset_batch_timer`), the timeout restarts instead of firing a premature flush. See [`openfilter/filter_runtime/filter.py` L1899-L1908](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1899-L1908).

3. **`sources_timeout` wall-clock tracking** — Replaced the `sources_timeout -= poll_ms` decrement with a `time.time()`-based deadline. `remaining` is now computed from actual elapsed wall time, preventing premature expiry when `mq.recv` returns faster than `poll_ms`. See [`openfilter/filter_runtime/filter.py` L1359](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1359) and [`openfilter/filter_runtime/filter.py` L1375-L1379](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1375-L1379).

4. **`_batch_pending_results` overwrite** — Changed from assignment (`=`) to `extend()` so undrained results from a prior batch aren't silently dropped. See [`openfilter/filter_runtime/filter.py` L1282](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1282).

5. **`fini()` drains `_batch_pending_results`** — `fini()` now sends any already-processed pending results before flushing the unprocessed `_frame_buffer`. See [`openfilter/filter_runtime/filter.py` L1696-L1700](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1696-L1700).

6. **Callable lazy-eval caveat** — Added a note to the `process()` docstring warning that the default `process_batch()` calls Callables eagerly, breaking the lazy-evaluation contract. See [`openfilter/filter_runtime/filter.py` L1818-L1820](https://github.com/PlainsightAI/openfilter/blob/1c7530b625a04ac6cf732e428f5eb4761465ce00/openfilter/filter_runtime/filter.py#L1818-L1820).

7. **RELEASE.md `_batch_size`** — Already fixed in e799e0d (prior commit).

openfilter/filter_runtime/filter.py
```diff
@@ -1015,72 +1229,197 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+        # Normalize results: wrap bare Frames, skip Nones
+        normalized = []
+        for result in results:
+            if result is None:
+                continue
+            normalized.append(self._normalize_frame_result(result))
+
+        if not normalized:
+            # Sink filter (all None results): inject timing into every input
+            # frame so EMA updates per-frame, matching _process_frames_single.
```
lucasmundim, line 1340: **Claimed fix not implemented: sink filter timing still only on `batch[-1]`**

Commit message says: "Inject timing into every frame for sink filters (not just batch[-1]), matching per-frame EMA update behavior of _process_frames_single."

But this code still only injects into `batch[-1]`:
```python
if not normalized:
    if self._is_last_filter and batch:
        self._inject_timings(batch[-1], ...)  # only last frame
```

Should be:
```python
if not normalized:
    if self._is_last_filter:
        for b_frames in batch:
            self._inject_timings(b_frames, t_in, t_out, process_time_ms)
```

openfilter/filter_runtime/filter.py
```diff
@@ -1385,32 +1837,120 @@ def process(self, frames: dict[str, Frame]) -> dict[str, Frame] | Frame | Callab
...
+    def _stop_batch_watcher(self) -> None:
+        self._batch_stop.set()
+        self._batch_wakeup.set()  # unblock the wait
+
+    def _reset_batch_timer(self) -> None:
+        """Reset the accumulation timeout by waking the watcher."""
+        self._ensure_batch_watcher()
+        self._batch_wakeup.set()
+
+    def _cancel_batch_timer(self) -> None:
```
lucasmundim, line 1923: **Claimed fix not implemented: batch watcher mid-sleep timer reset**

Commit message says: "Fix batch watcher mid-sleep timer reset: restart timeout loop when _batch_wakeup fires during accumulation sleep."

But the watcher doesn't check `_batch_wakeup` after the timeout wait:
```python
if not self._batch_stop.wait(timeout):  # sleeping here
    # _reset_batch_timer() sets _batch_wakeup during this sleep,
    # but nothing checks for it — flush fires anyway
    with self._batch_lock:
        self._batch_flush_event.set()  # premature flush!
```

The fix would be to check `_batch_wakeup.is_set()` after the wait returns and `continue` (restart the timeout) instead of flushing:
```python
if not self._batch_stop.wait(timeout):
    if self._batch_wakeup.is_set():
        self._batch_wakeup.clear()
        continue  # timer was reset, restart
    with self._batch_lock:
        self._batch_flush_event.set()
```

openfilter/filter_runtime/filter.py
```diff
@@ -1015,72 +1229,197 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
 
-            if (sources_timeout := sources_timeout - POLL_TIMEOUT_MS) <= 0:
+            if batch_mode and self._batch_flush_event.is_set():
+                flushed = self._process_batch_flush()
+                if flushed is not None:
+                    for result in flushed:
+                        self._send_frames(result, outputs_timeout)
+                    if (
+                        exit_after_t := self.exit_after_t
+                    ) is not None and time() >= exit_after_t:
```
lucasmundim: **Claimed fix not implemented: wall-clock deadline for sources_timeout**

Commit message says: "Track sources_timeout with wall-clock deadline instead of decrementing by poll_ms, preventing premature timeout expiry."

But the code still uses the countdown model:
```python
if (sources_timeout := sources_timeout - poll_ms) <= 0:
```

This is the original approach that was flagged as buggy. With `accumulate_timeout_ms=10ms` and `POLL_TIMEOUT_MS=50ms`, `poll_ms` becomes 10ms, and `sources_timeout` decrements by 10ms per loop iteration regardless of actual elapsed time — leading to premature expiry if `mq.recv` returns faster than `poll_ms`.

openfilter/filter_runtime/filter.py
```diff
@@ -1015,72 +1229,197 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+            batch = list(self._frame_buffer)
+            self._frame_buffer.clear()
+            self._batch_flush_event.clear()
+
+        results = self._execute_batch(batch)
+        # Stash intermediate results for loop_once to drain.
+        # _batch_pending_results is only accessed from the loop_once thread,
+        # so no lock is needed here.
+        if len(results) > 1:
+            self._batch_pending_results.extend(results[:-1])
```
lucasmundim, line 1289: **Claimed fix not implemented: extend() for _batch_pending_results**

Commit message says: "Use extend() instead of assignment for _batch_pending_results to avoid silently overwriting undrained results."

But the code still uses assignment:
```python
self._batch_pending_results = results[:-1]
```

Should be:
```python
self._batch_pending_results.extend(results[:-1])
```

With assignment, if a previous batch's results haven't been drained by `loop_once` yet, they're silently overwritten — data loss.

openfilter/filter_runtime/filter.py
```diff
@@ -1231,82 +1600,121 @@ def on_exit_msg(reason: str):
...
+                logger.info(
+                    f"[csv_exporter] Writing metrics every {csv_interval}s to {csv_path!r}"
+                )
 
         # Start metrics updater thread after MQ is initialized
-        if self.telemetry_enabled and hasattr(self, 'otel'):
+        if self.telemetry_enabled and hasattr(self, "otel"):
             self.start_metrics_updater_thread()
 
         # Start metadata worker for processing frame metadata asynchronously
```
lucasmundim, line 1718: **Claimed fix not implemented: drain _batch_pending_results in fini()**

Commit message says: "Drain _batch_pending_results in fini() before flushing _frame_buffer."

But `fini()` only drains `_frame_buffer` (lines 1691-1705). If there are already-processed results sitting in `_batch_pending_results` at shutdown (between batch completion and the next `loop_once` drain), they're silently dropped.

Add before the `_frame_buffer` flush:
```python
if self._batch_pending_results:
    for pending in self._batch_pending_results:
        self.mq.send(pending, POLL_TIMEOUT_MS)
    self._batch_pending_results.clear()
```

docs/overview.md
```diff
@@ -55,6 +55,29 @@ Finally, the `process()` function will be called on each frame or multiple frame
...
+        def process_batch(self, batch):
+            # batch is a list of frame dicts, each as would be passed to process()
+            images = [f["main"].img for f in batch]
```
lucasmundim: **Wrong Frame attribute**: `f["main"].img` — Frame has no `.img` attribute. Should be `.image` (raw image access) or `.rgb` / `.bgr` (format-converting access).

Also on line 68: `Frame(data={...})` — Frame constructor doesn't take a `data=` keyword argument. Should be `Frame(None, {...})` or `Frame({...})`.

lucasmundim (commented):
Good progress from prior rounds — formatting is split out, `threading.Event` replaces the bool flag, `_execute_batch` returns all results, batch length mismatch is WARNING, `fini()` has `exc_info=True`, and unrelated test changes are reverted.

However, the latest commit (`1c7530b`) claims 6 fixes but **4 of them are not actually implemented in the code**. The commit message describes the intended changes but the code doesn't match.

shingonoide:
Re-reviewing the frame accumulation feature. The core batching, locking, and timer patterns are sound. Lucas's merge conflict and implementation gap items should be the priority.

Additional notes:

**Doc bug still present** (`docs/overview.md`): The example uses `f["main"].img` but Frame has no `.img` property — only `.image`. This will cause AttributeError if copied.

**EMA metric semantics**: In single mode, `_update_process_time_ema` records per-frame time. In batch mode, it records total batch wall-clock time (N frames). This means `process_time_ms` metrics become incomparable between modes. Worth documenting explicitly — not a bug, but dashboards/alerts comparing across modes will see misleading values.

**Misleading comment on _batch_pending_results**: The comment says "only accessed from the loop_once thread — no lock needed" but `fini()` also accesses and clears it. The execution is sequential (fini runs after loop exits), so there's no actual race, but the comment should be updated to reflect that fini() is also an accessor.

Overall well-designed feature. After merge conflict resolution, the .img fix, and Lucas's items, this looks ready.

shingonoide:
Core batch accumulation logic is sound - good thread safety via `_batch_lock`, proper timeout flushing, shutdown draining, and backward compatibility. Previous review items (silent result discard, threading.Event, batch length mismatch logging) are all addressed.

**One note:**

1. **`time()` bare call** (lines 1371, 1400, 1418): The file has `import time` (module) but no `from time import time`. Bare `time()` calls will raise `TypeError: 'module' object is not callable` when `exit_after` is configured. This is a **pre-existing bug on main** (same pattern at lines 1077, 1253), but this PR copies it to 2 additional locations. Since you're touching this code, adding `from time import time` to the imports (or changing to `time.time()`) would be a quick fix.

**Formatting note:** ~70% of the diff is still formatting changes mixed with feature code. Not blocking, but separating them into distinct commits would make future git blame/bisect easier.

Overall the implementation is solid. The accumulation, locking, and drain logic are well-tested.

shingonoide (dismissed):
Strong PR — the batch accumulation design is well-architected: single watcher thread, proper lock separation, clean `_process_frames_single` / `_process_frames_batched` split, and `batch_size=1` takes the old code path with zero overhead. Shutdown ordering in `fini()` is correct. 24 tests with good coverage. DCO compliance verified on all commits.

A few suggestions:

**1. Bug: `time()` should be `time.time()`** (`filter.py`, batch flush paths ~lines 1371, 1400)
Two new lines call bare `time()` but the file imports `time` as the module, so this raises `TypeError: 'module' object is not callable`. Fires when `exit_after` is configured and the batch flush path is taken. Note: this copies a pre-existing pattern, but worth fixing in these new lines.

**2. `_batch_watcher_loop` has no error handling** (`filter.py:1888-1908`)
If the watcher thread crashes from an unhandled exception, it dies silently as a daemon thread. No more timeout flushes happen — partial batches get stuck in the buffer forever with no error logged. Suggestion: wrap loop body in `try/except Exception` that logs and continues.

**3. `fini()` pending results drain is unprotected** (`filter.py:1697-1700`)
The `mq.send` calls for `_batch_pending_results` are outside any try/except. If a downstream filter has disconnected, `mq.send` raises and prevents the rest of `fini()` from executing. Suggestion: wrap in try/except like the subsequent buffer flush.

**4. Nit: log frame count on shutdown flush failure** (`filter.py:1713-1715`)
The shutdown flush catch says "Failed to flush partial batch" but doesn't say how many frames were lost. Adding `len(batch)` would help operators assess impact.

docs/overview.md
lucasmundim: **Still wrong Frame API** (flagged in previous review, not yet addressed)

- Line 65: `f["main"].img` - Frame has no `.img` attribute. The correct accessor is `.image` (raw), or `.rgb` / `.bgr` / `.gray` for format-converting access.
- Line 68: `Frame(data={**f["main"].data, "detections": d})` - Frame's constructor is `Frame(image, data, format)` or `Frame(dict)` for data-only frames. The keyword-arg form `Frame(data=...)` doesn't match the constructor signature.

Suggested fix:
```python
images = [f["main"].image for f in batch]
detections = self.model.infer_batch(images)
return [
    {"main": Frame(f["main"].image, {**f["main"].data, "detections": d}, f["main"].format)}
    for f, d in zip(batch, detections)
]
```

lucasmundim (changes requested):
All 5 previously-flagged issues (sink filter timing, watcher timer reset, wall-clock deadline, extend(), fini drain) are now properly fixed. CI is green.

Two items remain:

1. **docs/overview.md Frame API** (inline comment) - still uses wrong attribute names

2. **Needs rebase on main** - the branch forked from `c291174` (v0.1.27) before PR #77 (benchmarking harness) was merged. The branch carries its own older versions of `Makefile`, `pyproject.toml`, `tests/helpers.py`, and `tests/test_benchmarks.py`. When merged, this will silently **revert** PR #77's additions: the `make bench` target, `pytest-benchmark` dependency, 845-line `tests/test_benchmarks.py`, the `slow` pytest marker, and `tests/helpers.py`. Please rebase on `main` to pick up those changes.

stwilt:
Both points addressed.

1. **Frame API in docs/overview.md** — Fixed per your suggestion. Uses `.image` and the positional `Frame(image, data, format)` constructor. See [`docs/overview.md` L65](https://github.com/PlainsightAI/openfilter/blob/3683a6b38a30c986c25639a4dde72214259f70fb/docs/overview.md#L65) and [`docs/overview.md` L68](https://github.com/PlainsightAI/openfilter/blob/3683a6b38a30c986c25639a4dde72214259f70fb/docs/overview.md#L68).

2. **Rebased on main** — Branch is now based on current main (`f3384ad`), picking up PR #77 (benchmarking harness: `make bench`, `pytest-benchmark`, `tests/test_benchmarks.py`, `tests/helpers.py`, `slow` marker) and PR #75 (framework-agnostic GPU detection). VERSION bumped to `v0.1.29` and RELEASE.md notes combined with main's v0.1.28 GPU detection entry.

The Black formatting commit was squashed into the feature commit during the rebase — main's reformatted `_validate_device` conflicted with the formatting-only pass, and resolving it commit-by-commit would have produced a broken intermediate state. The feature commit now carries both changes; the other commits (refactor, version bump, review fixes) remain separate.

docs/overview.md
stwilt: Fixed at [`docs/overview.md` L65](https://github.com/PlainsightAI/openfilter/blob/3683a6b38a30c986c25639a4dde72214259f70fb/docs/overview.md#L65) and [`docs/overview.md` L68](https://github.com/PlainsightAI/openfilter/blob/3683a6b38a30c986c25639a4dde72214259f70fb/docs/overview.md#L68). Applied your suggested fix: `.image` attribute and positional `Frame(image, data, format)` constructor.

lucasmundim (dismissed):
All previous review comments addressed. Docs example uses correct Frame API, rebase onto main preserves the benchmarking harness from PR #77. Batch accumulation logic is solid — clean design, proper locking, thorough test coverage.

shingonoide (dismissed):
Clean, well-designed feature. Frame accumulation with proper thread safety, thorough tests (24 cases), and solid backward compatibility (batch_size=1 takes the existing code path unchanged). The single watcher thread replacing per-timeout Timers is a nice improvement.

All issues from the 5 review rounds have been addressed. Two minor non-blocking suggestions below.

openfilter/filter_runtime/filter.py
```diff
@@ -1019,72 +1233,197 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+        return results if results else None
+
+    def _execute_batch(self, batch: list[dict[str, Frame]]) -> list[dict[str, Frame]]:
+        """Run process_batch() and return normalized, timing-injected results.
+
+        Returns a list of frame dicts (possibly empty). The caller is
+        responsible for sending each result downstream.
+        """
+        t_in = time.time()
+        results = self.process_batch(batch)
```
shingonoide: nit (non-blocking): Consider logging the batch size here before re-raising if `process_batch()` throws - helps operators assess frame loss impact on transient errors (e.g., GPU OOM with a large batch).

openfilter/filter_runtime/filter.py
```diff
@@ -1019,72 +1233,197 @@ def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callabl
...
+            return self._normalize_frame_result(processed_frames)
+
+    def _process_frames_batched(
+        self, frames: dict[str, Frame]
+    ) -> dict[str, Frame] | None:
+        if not frames:
+            return None
+        with self._batch_lock:
+            self._cancel_batch_timer()
+            self._frame_buffer.append(frames)
```
shingonoide, line 1274: nit (non-blocking): You could also clear `_batch_flush_event` here (alongside `_cancel_batch_timer`) to prevent a minor race where the watcher's expiring timeout sets a stale flush event just as a new frame arrives. Not a correctness issue - `_take_flush_batch` handles it - but avoids a slightly suboptimal premature flush at the timeout boundary.

openfilter/filter_runtime/filter.py
stwilt: Done in 10ce6af — wrapped the `process_batch()` call in try/except that logs the batch size before re-raising. See [`openfilter/filter_runtime/filter.py` L1302-L1310](https://github.com/PlainsightAI/openfilter/blob/10ce6afada0f047a2a1afb9442a4868002c234f5/openfilter/filter_runtime/filter.py#L1302-L1310).

openfilter/filter_runtime/filter.py
stwilt: Done in 10ce6af — `_batch_flush_event.clear()` now runs alongside `_cancel_batch_timer()` inside `_process_frames_batched`. See [`openfilter/filter_runtime/filter.py` L1270-L1273](https://github.com/PlainsightAI/openfilter/blob/10ce6afada0f047a2a1afb9442a4868002c234f5/openfilter/filter_runtime/filter.py#L1270-L1273).

lucasmundim (dismissed):
Latest commit addresses shingonoide's nit (batch size logging on process_batch failure) and fixes a stale flush event race. All prior review comments resolved. CI green.

shingonoide (dismissed):
Re-approving after the latest commit. Both changes look good - the stale flush event clear prevents the race condition we flagged, and the batch-size logging on process_batch() failure adds useful observability. Clean fix.

leandrobmarinho (dismissed):
**Approve with comments** — The feature is well motivated, exercised by a solid unit-test suite, and several follow-up fixes (watcher thread, `time.time()`, `sources_timeout` deadline, shutdown flush, race on `_batch_flush_event`) show real production awareness. I would merge after addressing or explicitly accepting the **result-length / `None` semantics** called out below (documentation or stricter validation).

## Understanding

This PR adds optional **frame batching** to `Filter`: when `batch_size > 1`, frames are buffered and `process_batch()` runs on a full batch or after `accumulate_timeout_ms`. A **single daemon watcher thread** replaces per-timeout timers. Default `process_batch()` delegates to `process()` per frame for backward compatibility. Docs (`docs/overview.md`), release notes (`RELEASE.md`), version **v0.1.29**, and **`tests/test_frame_accumulation.py`** are included.

## Summary of changes

- **`openfilter/filter_runtime/filter.py`**: Batch state (buffer, lock, flush event, watcher), `process_frames` → `_process_frames_single` / `_process_frames_batched`, `_execute_batch`, `loop_once` integration (poll tuning, flush on MQ timeout, pending results), `fini()` flush, `normalize_config` validation, `process_batch` hook, Black-heavy formatting across the file.
- **`tests/test_frame_accumulation.py`**: New tests (config, accumulation, timeout flush, watcher lifecycle, threading, backward compat for `batch_size=1`).
- **`docs/overview.md`**, **`RELEASE.md`**, **`VERSION`**.

## Findings

### Blocking (must fix before merge)

None identified from static review. The `time()` → `time.time()` fixes and watcher `try/except` address real failure modes.

### Important (should fix or track before release)

1. **`process_batch` return value contract vs `_execute_batch` normalization**  
   Docstrings say each list element corresponds to an input frame, and `_execute_batch` warns when `len(results) != len(batch)`. The implementation then **drops `None` entries** when building `normalized`:

   - If `process_batch` returns `[None, frame_dict, None]` for three inputs, only one output is finalized and sent downstream; the other two inputs are not represented as outputs. That may be intentional for “sink” behavior but it is **not** “one result per frame” anymore.  
   - **Suggestion**: Either (a) enforce 1:1 mapping (keep `None` as placeholders and let `_send_frames` / MQ layer skip only where that is explicitly defined), or (b) document that **`None` means “no downstream payload for that slot”** and that non-`None` results are compacted—so implementers are not surprised.  
   - The unit test `test_batch_with_none_results_skipped` validates the current behavior but does not assert downstream parity for all three inputs.

2. **Mismatched `len(results)` vs `len(batch)` after `None` stripping**  
   Even when lengths match before stripping, skipping `None` can make the number of emitted frames differ from `len(batch)`. Consider logging at **warning** when `len(normalized) != len(batch)` after normalization, not only when raw `len(results) != len(batch)`.

### Minor / nitpick (optional)

1. **Diff noise**: A large fraction of `filter.py` is Black/quote reflow unrelated to batching. It is already in the branch; for future PRs, splitting **format-only** from **logic** reduces review load and cherry-pick pain.

2. **Class attribute `Filter._batch_size = 1`** alongside `self._batch_size` in `__init__` is slightly redundant; harmless but could confuse readers (instance always overrides).

3. **`docs/overview.md` example** uses `Frame(...)` construction in a snippet—ensure that matches the real `Frame` API exported to users (quick sanity check on import path / constructor signature).

4. **`fini()` shutdown path** uses `mq.send(..., POLL_TIMEOUT_MS)` with a single short timeout per pending item; under backpressure this may still drop or partially flush—acceptable if documented as best-effort (the code already logs warnings on failure).

## Testing & verification

- **Author**: 24 unit tests covering config, accumulation, timeout, watcher, threads, `batch_size=1` compat—good breadth for unit scope.
- **Recommended before release**:
  - Run full existing test suite (not just new file) after such a large `filter.py` touch.
  - **Integration**: One pipeline with `batch_size > 1` and real MQ, plus a downstream consumer, to validate ordering, partial batches, and shutdown under load (author references a downstream PR—align before tagging).

## Notes / questions for the author

1. For filters that mix **`process()` returning Callables** with eventual `batch_size > 1`, the doc now states default `process_batch` calls Callables eagerly—confirm that is acceptable for all internal use cases, or if a guard/warning in config is warranted.

2. Should **`FILTER_BATCH_SIZE`** / **`FILTER_ACCUMULATE_TIMEOUT_MS`** appear in `get_config()`-driven help anywhere else (CLI docs) beyond `overview.md`?

## What went well (brief)

- Clear **problem statement** (GPU amortization) and **defaults** preserving legacy behavior.
- Thoughtful **concurrency** design (one watcher thread, flush event coordination, lock scope) and iterative fixes from review (deadline-based `sources_timeout`, `_batch_pending_results` drain, OOM logging on `process_batch` failure).
- **Tests** are practical: validation, flush, watcher lifecycle, and **no watcher when `batch_size=1`**

openfilter/filter_runtime/filter.py
stwilt: Fixed — `_execute_batch` now iterates `for b_frames in batch:` and calls `_inject_timings` per frame at [`openfilter/filter_runtime/filter.py` L1342-L1343](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/openfilter/filter_runtime/filter.py#L1342-L1343).

stwilt:
All 18 open review threads addressed. The five "claimed fix not implemented" items are all implemented in the current HEAD — the review comments were against an earlier state that has since been fixed. Added an integration test for the `loop_once` timeout-flush path and clarified the `process_batch` docstring in `cdea414`. See thread replies below for per-item details.

openfilter/filter_runtime/filter.py
stwilt: Fixed — the inner `while True` loop checks `_batch_wakeup.is_set()` and restarts the timeout at [`openfilter/filter_runtime/filter.py` L1942-L1945](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/openfilter/filter_runtime/filter.py#L1942-L1945).

openfilter/filter_runtime/filter.py
stwilt: Fixed — uses `deadline = time.time() + sources_timeout / 1000.0` and `remaining = deadline - time.time()` at [`openfilter/filter_runtime/filter.py` L1373-L1393](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/openfilter/filter_runtime/filter.py#L1373-L1393).

openfilter/filter_runtime/filter.py
stwilt: Fixed — uses `.extend(results[:-1])` at [`openfilter/filter_runtime/filter.py` L1289](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/openfilter/filter_runtime/filter.py#L1289).

openfilter/filter_runtime/filter.py
stwilt: Fixed — `fini()` drains `_batch_pending_results` before buffer flush at [`openfilter/filter_runtime/filter.py` L1725-L1736](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/openfilter/filter_runtime/filter.py#L1725-L1736).

docs/overview.md
stwilt: Fixed — now uses `.image` (the correct Frame attribute) throughout at [`docs/overview.md` L65-L69](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/docs/overview.md#L65-L69).

stwilt: Fixed — `.image` throughout at [`docs/overview.md` L65-L69](https://github.com/PlainsightAI/openfilter/blob/cdea4147da3ea727a378e8105622a012ae4bb227/docs/overview.md#L65-L69).

openfilter/filter_runtime/filter.py
stwilt: Acknowledged — good idea for a follow-up. Not adding in this PR to avoid more churn.

lucasmundim (approved):
Good addition — Callable docstring clarification and loop_once timeout flush integration test. CI green.

shingonoide:
Posted a small follow-up at #83 with three regression tests covering the watcher timer reset, fini() pending-results drain, and loop_once intermediate-result send order. Tests-only; no production changes.

PR #75, opened 2026-04-02
feat: framework-agnostic GPU detection via ctypes (replaces torch.cuda + nvidia-smi)

shingonoide:
## Summary

- **GPU detection via CUDA Driver API** — `detect_gpu()` uses `cuInit`, `cuDeviceGetCount`, `cuDeviceGetName` via ctypes. No torch import required in the core framework.
- **GPU metrics via NVML** — `get_gpu_metrics()` uses `nvmlDeviceGetUtilizationRates`, `nvmlDeviceGetMemoryInfo` via ctypes. Replaces `nvidia-smi` subprocess calls (faster, no PATH dependency).
- **GPU library preloading** — Preloads `libcuda.so.1` and `libnvidia-ml.so.1` with `RTLD_GLOBAL` via absolute paths before any framework import. Fixes CUDA init failures on GKE where `LD_LIBRARY_PATH` is not set at process startup.
- **Zero torch imports in core** — `openfilter/filter_runtime/` no longer imports torch. Individual filters still import torch in their own `setup()`.
- **Secondary torch validation** — When torch is installed, `_validate_device()` also checks `torch.cuda.is_available()` to catch CPU-only wheels or incompatible CUDA runtimes.
- **OPENFILTER_FORK=1 fix** — Now explicitly forces fork mode (previously only skipped spawn, relying on platform default).

## New file: `openfilter/filter_runtime/gpu.py`

| Function | Replaces | API |
|---|---|---|
| `preload_gpu_libs()` | `_preload_gpu_libs()` in filter.py | ctypes.CDLL with RTLD_GLOBAL |
| `detect_gpu()` → `GPUInfo` | `torch.cuda.is_available()`, `torch.cuda.device_count()` | CUDA Driver API |
| `get_gpu_metrics()` → `list[GPUMetrics]` | `subprocess.run(['nvidia-smi', ...])` | NVML |
| `nvml_shutdown()` | N/A | Clean NVML state on exit |

## Why

The openfilter framework should not depend on torch for GPU detection. Filters using TensorFlow, ONNX Runtime, or custom CUDA code shouldn't need torch installed just to detect GPUs.

## Review fixes (v0.1.28)

- `_find_gpu_lib` no longer returns a bare name fallback - returns `None`, callers use `path or "libcuda.so.1"` to try the bare name
- `set_start_method` moved from import-time side effect to `_configure_start_method()` called at top of `cmd_run()`
- `NvmlUtilization`/`NvmlMemory` ctypes structures moved to module level (avoid recreation per call)
- `preload_gpu_libs()` now populates `_libcuda`/`_libnvml` cache to avoid double-loading

## Tests

34 tests covering GPU detection, NVML metrics, preloading, and device validation (including driver-present-but-torch-unavailable scenarios).

## Test plan

- [x] 34 GPU tests pass
- [x] 379 other tests pass (9 pre-existing failures unrelated)
- [x] ruff check + format clean
- [ ] Deploy to GKE and verify torch.cuda.is_available() == True
- [ ] Verify GPU metrics in Grafana dashboard (NVML vs old nvidia-smi values)


tests/test_gpu_preload.py
```diff
@@ -0,0 +1,136 @@
...
+class TestPreloadRunsBeforeApplyEnvVars:
+    """Verify call order in filter.py module-level code."""
+
+    def test_preload_runs_before_apply_env_vars(self):
+        """_preload_gpu_libs() is called before _apply_append_env_vars() at module level."""
+        import inspect
+        from openfilter.filter_runtime import filter as filter_module
+
+        source = inspect.getsource(filter_module)
+        preload_pos = source.index("_preload_gpu_libs()")
```
lucasmundim: **Bug:** `source.index("_preload_gpu_libs()")` finds the first occurrence of that substring in the source — which is the **function definition** `def _preload_gpu_libs():`, not the module-level call. Same for `_apply_append_env_vars()`.

Since both function definitions come before both calls, and the definitions are already in the desired order, this test always passes regardless of the actual call order. It would still pass if the two module-level calls were swapped.

Consider searching for the bare call (no `def ` prefix) or using a more targeted pattern, e.g.:
```python
import re
calls = re.findall(
    r'^(?!\s*def ).*(_preload_gpu_libs|_apply_append_env_vars)\(\)',
    source, re.MULTILINE,
)
assert calls[0] == '_preload_gpu_libs'
assert calls[1] == '_apply_append_env_vars'
```

lucasmundim (commented):
The core fix is solid — using `ctypes.CDLL` with absolute paths to bypass the `LD_LIBRARY_PATH` limitation is the right approach, and the PR is well-documented with references. Found a few issues worth addressing.

openfilter/cli/cmd_run.py
```diff
@@ -4,6 +4,19 @@
 import os
 from pprint import pp
 
+# IMPORTANT: Set multiprocessing start method BEFORE importing Filter, because importing Filter
+# initializes the multiprocessing context with the default start method. On Linux, this defaults
+# to 'fork', which in Python 3.12+ causes issues when fork() is called in a multi-threaded process
+# (threading is introduced by SignalStopper). This must be set with force=True since we can't know
+# if another module has already initialized the context.
+# See: https://github.com/PlainsightAI/openfilter/issues/XX
+if not os.environ.get('OPENFILTER_FORK'):
```
lucasmundim: **Bug:** The old code used `opts.fork` (the `--fork` / `-f` CLI flag) to decide the start method. This replaces it with the `OPENFILTER_FORK` env var, but the `--fork` argparse flag is still defined (line 77) and still parsed — it's just silently ignored now.

Users who relied on `openfilter run -f` will get `spawn` instead of `fork` with no warning.

Either:
- Wire `--fork` to set the env var before this block runs (tricky since this is module-level), or
- Remove the `--fork` flag entirely and document `OPENFILTER_FORK` as the replacement, or
- Move this logic back into `cmd_run()` but before the Filter import (lazy import).

openfilter/cli/cmd_run.py
```diff
@@ -4,6 +4,19 @@
 import os
 from pprint import pp
 
+# IMPORTANT: Set multiprocessing start method BEFORE importing Filter, because importing Filter
+# initializes the multiprocessing context with the default start method. On Linux, this defaults
+# to 'fork', which in Python 3.12+ causes issues when fork() is called in a multi-threaded process
+# (threading is introduced by SignalStopper). This must be set with force=True since we can't know
+# if another module has already initialized the context.
+# See: https://github.com/PlainsightAI/openfilter/issues/XX
```
lucasmundim: Nit: placeholder issue number `XX` — should be a real issue link or removed.

openfilter/filter_runtime/filter.py
```diff
@@ -40,6 +40,47 @@
...
+    for search_dir in ld_path.split(":"):
+        for lib_name in _GPU_LIBS:
+            lib_path = os.path.join(search_dir, lib_name)
+            if lib_path in _preloaded_gpu_libs:
+                continue
+            if os.path.exists(lib_path):
+                try:
+                    ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)
+                    _preloaded_gpu_libs.add(lib_path)
+                except OSError:
```
lucasmundim: **Suggestion:** This silently swallows load failures. Since this code targets GKE GPU debugging scenarios, a `logger.debug()` or `logger.warning()` here would help diagnose cases where the library file exists but still can't be loaded (e.g., architecture mismatch, missing transitive deps).

Note: `logger` isn't defined yet at this point in the file, but you could use `logging.getLogger(__name__).debug(...)` inline or move logger init earlier.

openfilter/filter_runtime/filter.py
```diff
@@ -66,7 +107,8 @@ def _apply_append_env_vars():
                 os.environ[target_var] = new_val
 
 
-_apply_append_env_vars()
+_preload_gpu_libs()   # Must run BEFORE any torch import
+_apply_append_env_vars()  # Still useful for child processes and PATH
```
lucasmundim, line 73: **Suggestion:** The `_apply_append_env_vars()` docstring above still says "Must be called before any native library imports (cv2, torch) so that the dynamic linker sees the updated paths" — but this PR itself demonstrates that setting `LD_LIBRARY_PATH` from Python does NOT affect `dlopen()`. Consider updating the docstring to reflect that this function is only effective for `PATH` and for child processes that inherit the environment.

tests/test_gpu_preload.py
shingonoide: Good catch. Fixed in `d43dda2` — replaced `source.index()` with `re.search()` using a negative lookahead for `def` to match only module-level calls, not function definitions. Also added assertions that the matches are found.

openfilter/cli/cmd_run.py
shingonoide: Fixed in `d43dda2` — added a `DeprecationWarning` when `--fork` is used, informing users to migrate to `OPENFILTER_FORK=1`. Also wired it to set the env var via `os.environ.setdefault` so existing behavior is preserved for backwards compatibility.

openfilter/cli/cmd_run.py
shingonoide: Fixed in `d43dda2` — removed the placeholder comment entirely. No matching issue was found in the repo.

openfilter/filter_runtime/filter.py
shingonoide: Good suggestion. Fixed in `d43dda2` — replaced bare `pass` with `logging.getLogger(__name__).debug(\"Could not preload %s: %s\", lib_path, e)` in the except block.

openfilter/filter_runtime/filter.py
shingonoide: Fixed in `d43dda2` — updated the docstring to clarify that setting `LD_LIBRARY_PATH` from Python does NOT affect `dlopen()` in the current process, and the function is only effective for `PATH` and for child processes.

openfilter/cli/cmd_run.py
```diff
@@ -84,6 +98,15 @@ def cmd_run(args):
...
     opts = parser.parse_args(opts)
 
+    if opts.fork:
+        warnings.warn(
+            "The --fork / -f flag is deprecated. Set OPENFILTER_FORK=1 environment "
+            "variable instead. The flag will be removed in a future release.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        os.environ.setdefault('OPENFILTER_FORK', '1')
```
lucasmundim: **`--fork` still doesn't work — it just warns now.**

The module-level block at lines 12–17 reads `OPENFILTER_FORK` and calls `mp.set_start_method('spawn', force=True)` **at import time**. By the time `cmd_run()` parses `opts.fork` and calls `os.environ.setdefault('OPENFILTER_FORK', '1')` here, the start method is already locked in as `spawn`. Setting the env var afterwards has no effect on the current process — nothing rereads it.

Result: a user who runs `openfilter run -f` gets the deprecation warning, but the process still uses spawn, not fork. The flag is silently broken in a different way than before.

To actually make `--fork` work during the deprecation window, call `set_start_method` directly here:
```python
if opts.fork:
    warnings.warn(...)
    try:
        mp.set_start_method('fork', force=True)
    except RuntimeError:
        pass
```

The `os.environ.setdefault` line can be removed — it's load-bearing for nothing in the current process, and child processes spawned by `mp` inherit the already-set start method, not the env var.

openfilter/cli/cmd_run.py
```diff
@@ -4,6 +4,20 @@
...
+# (threading is introduced by SignalStopper). This must be set with force=True since we can't know
+# if another module has already initialized the context.
+if not os.environ.get('OPENFILTER_FORK'):
+    try:
+        mp.set_start_method('spawn', force=True)
+    except RuntimeError:
+        # Context already set, continue with current method
+        pass
+
+import warnings
```
lucasmundim: Nit: `import warnings` is placed mid-file between the `mp.set_start_method` block and the Filter import. PEP 8 groups all imports at the top. Move it up with the other stdlib imports (after `from pprint import pp`) for consistency. Not a bug, just style.

lucasmundim (commented):
Good progress — 4 of the 5 items from the previous review are correctly addressed: the test regex now properly finds module-level calls, the `XX` placeholder is removed, `OSError` failures now log at DEBUG, and the `_apply_append_env_vars` docstring is updated.

However, the `--fork` deprecation fix has a subtle but real problem: **the `--fork` flag still doesn't actually work**, it just emits a warning.

openfilter/cli/cmd_run.py
shingonoide: Fixed - now calls `mp.set_start_method('fork', force=True)` directly instead of setting env var after the fact.

openfilter/cli/cmd_run.py
shingonoide: Fixed - moved to top with other stdlib imports.

openfilter/cli/cmd_run.py
```diff
@@ -2,8 +2,21 @@
...
 import os
+import warnings
 from pprint import pp
 
+# IMPORTANT: Set multiprocessing start method BEFORE importing Filter, because importing Filter
+# initializes the multiprocessing context with the default start method. On Linux, this defaults
+# to 'fork', which in Python 3.12+ causes issues when fork() is called in a multi-threaded process
+# (threading is introduced by SignalStopper). This must be set with force=True since we can't know
+# if another module has already initialized the context.
+if not os.environ.get('OPENFILTER_FORK'):
```
lucasmundim: **Bug:** `os.environ.get("OPENFILTER_FORK")` is truthy for any non-empty string, including `"0"`, `"false"`, `"no"`. A user setting `OPENFILTER_FORK=0` would accidentally skip the spawn override and get fork mode.

The deprecation warning says "Set `OPENFILTER_FORK=1`", implying a value-based check. The codebase already has `strtobool` in `openfilter/filter_runtime/utils.py` — consider using it here, or at minimum compare `== "1"`.

openfilter/filter_runtime/filter.py
```diff
@@ -40,6 +40,49 @@
...
+
+    On Kubernetes (GKE), the NVIDIA device plugin mounts driver libraries at
+    /usr/local/nvidia/lib64/ but does NOT add this path to LD_LIBRARY_PATH.
+
+    The dynamic linker (ld.so) reads LD_LIBRARY_PATH once at process startup
+    and caches it. Setting os.environ['LD_LIBRARY_PATH'] from Python has NO
+    effect on dlopen() — the cached value is used (see CPython issue #100567).
+
+    However, dlopen() with an ABSOLUTE PATH bypasses LD_LIBRARY_PATH entirely.
+    By preloading libcuda.so.1 with RTLD_GLOBAL before torch is imported,
```
lucasmundim: **Suggestion:** The docstring frames this as a PyTorch-specific fix, but the `ctypes.CDLL(..., RTLD_GLOBAL)` preload is GPU-runtime-agnostic — it makes NVIDIA driver symbols available to *any* subsequent `dlopen()` call. TensorRT, ONNX Runtime, CuPy, and other CUDA consumers benefit equally.

Since OpenFilter is open-source and runtime-agnostic, consider reframing: replace "before torch is imported, PyTorch's CUDA initialization can find" with something like "before any CUDA runtime is imported, CUDA initialization (PyTorch, TensorRT, ONNX Runtime, etc.) can find".

lucasmundim (dismissed):
All 5 items from the previous rounds are addressed. The core ctypes.CDLL preload fix is correct, well-documented, and well-tested. Approving with a couple of comments.

**Not inline (unchanged line):** The `--fork` argparse help text doesn't mention deprecation. Users reading `--help` won't know until they use it. Consider: `help="(deprecated: use OPENFILTER_FORK=1 env var) use fork method instead of spawn"`.

openfilter/cli/cmd_run.py
```diff
@@ -2,8 +2,27 @@
...
+# Filter may initialize the multiprocessing context. On Linux, the default is
+# 'fork', which in Python 3.12+ causes issues when fork() is called in a
+# multi-threaded process (threading is introduced by SignalStopper).
+#
+# OPENFILTER_FORK=1 explicitly forces 'fork' mode (e.g. for non-CUDA filters
+# where fork is faster). Without it, 'spawn' is used for CUDA compatibility.
+if os.environ.get("OPENFILTER_FORK"):
+    try:
+        mp.set_start_method("fork", force=True)
+    except RuntimeError:
```
lucasmundim: **Carried over from previous review:** `os.environ.get("OPENFILTER_FORK")` is truthy for any non-empty string, including `"0"`, `"false"`, `"no"`. A user setting `OPENFILTER_FORK=0` would accidentally get fork mode.

The codebase has `strtobool` in `openfilter/filter_runtime/utils.py` — consider using it, or compare `== "1"`.

openfilter/filter_runtime/gpu.py
```diff
@@ -0,0 +1,273 @@
...
+    if path is None:
+        return None
+    try:
+        _libcuda = ctypes.CDLL(path, mode=ctypes.RTLD_GLOBAL)
+        return _libcuda
+    except OSError as exc:
+        logger.debug("Failed to load libcuda: %s", exc)
+        return None
+
+
```
lucasmundim, line 107: **Nit:** `_find_gpu_lib` never returns `None` — the `return name` fallback on line 98 means the `if path is None: return None` guards in `_load_libcuda` (line 108) and `_load_libnvml` (line 125) are dead code.

The fallback is reasonable (let dlopen try its default search), but the dead-code guards are misleading. Either:
- Remove the `if path is None` checks (the `except OSError` already handles the failure), or
- Have this function return `None` when the lib isn't found on disk, and let callers explicitly try the bare name

lucasmundim (dismissed):
Great evolution — extracting a dedicated `gpu.py` module with ctypes-based detection (CUDA Driver API) and metrics (NVML) is exactly the right approach. The layered `_validate_device` (driver check first, optional torch secondary) is clean, and replacing `nvidia-smi` subprocess with NVML resolves the existing TODO. Tests are thorough.

Two minor items remain, plus one carried over from the previous review.

**Carried over (unchanged line, can't inline):** The `--fork` argparse help text still doesn't mention deprecation. Consider: `help="(deprecated: use OPENFILTER_FORK=1) use fork instead of spawn"`.

openfilter/cli/cmd_run.py
shingonoide: Fixed - changed `os.environ.get("OPENFILTER_FORK")` to `== "1"` so strings like `"0"`, `"false"`, or `"no"` no longer accidentally enable fork mode. Kept it simple and consistent with the `== "1"` pattern rather than adding a `strtobool` dependency in the CLI startup block.

openfilter/filter_runtime/filter.py
shingonoide: Updated the docstring to say "before any CUDA consumer is imported (PyTorch, TensorRT, ONNX Runtime, CuPy, etc.)" - no longer singles out PyTorch, which was misleading given the ctypes approach is fully framework-agnostic.

shingonoide:
Updated `--fork` help text to: `"(deprecated: use OPENFILTER_FORK=1 env var) use fork method instead of spawn"` - users will now see the deprecation notice in `--help` output before hitting the runtime warning.

openfilter/cli/cmd_run.py
shingonoide: Fixed in the same commit — changed to `== "1"`, so only the exact string `"1"` enables fork mode.

openfilter/filter_runtime/gpu.py
shingonoide: Good catch. `_find_gpu_lib` always returns a string (falls back to bare name for dlopen). Removed the dead `if path is None` guards in both `_load_libcuda` and `_load_libnvml`.

lucasmundim:
Hey @shingonoide — just a heads up, there are 3 minor items still open from the last review (all nits, nothing blocking):

1. **`OPENFILTER_FORK` truthiness** (cmd_run.py:18) — `os.environ.get("OPENFILTER_FORK")` treats `"0"`/`"false"` as truthy. Consider using `strtobool` or `== "1"`.
2. **Dead code** (gpu.py) — `_find_gpu_lib` never returns `None` (line 98 fallback), so the `if path is None` guards in `_load_libcuda`/`_load_libnvml` are unreachable.
3. **`--fork` help text** — argparse help doesn't mention the flag is deprecated.

Already approved — these can be addressed in this PR or a follow-up, your call.

lucasmundim (commented):
Overall the design is solid — ctypes for GPU detection is the right call, the NVML metrics path is cleaner than shelling out to nvidia-smi, and the test coverage is thorough. A few issues worth addressing:

openfilter/filter_runtime/gpu.py
```diff
@@ -0,0 +1,273 @@
...
+        if os.path.exists(lib_path):
+            return lib_path
+
+    # Fall back to system library search
+    found = ctypes.util.find_library(name.split(".so")[0].lstrip("lib"))
+    if found:
+        return found
+
+    # Last resort: try the bare name and let dlopen search
+    return name
```
lucasmundim: This "last resort" fallback means `_find_gpu_lib` never returns `None`, despite the `str | None` return type. The `if path is None: return None` checks in `_load_libcuda` and `_load_libnvml` are dead code.

Not a runtime bug (the `OSError` from `ctypes.CDLL("libcuda.so.1")` is caught), but the function contract is misleading. Consider either:
- Removing this fallback and returning `None` here (let callers decide whether to try the bare name), or
- Removing the `None` checks in callers and updating the return type to `str`

openfilter/cli/cmd_run.py
```diff
@@ -2,8 +2,27 @@
...
+        pass
+else:
+    try:
+        mp.set_start_method("spawn", force=True)
+    except RuntimeError:
+        pass
```
lucasmundim: This runs as a side effect on `import cmd_run`. Any test or tool that imports this module will silently force the multiprocessing start method for the entire process. This could cause confusing failures in tests that set their own start method, or in contexts where `cmd_run` is imported for introspection (e.g. CLI help generation).

Consider deferring this into a function called at the top of `cmd_run()` instead.

openfilter/filter_runtime/gpu.py
```diff
@@ -0,0 +1,273 @@
...
+
+        class NvmlUtilization(ctypes.Structure):
+            _fields_ = [("gpu", ctypes.c_uint), ("memory", ctypes.c_uint)]
+
+        class NvmlMemory(ctypes.Structure):
+            _fields_ = [
+                ("total", ctypes.c_ulonglong),
+                ("free", ctypes.c_ulonglong),
+                ("used", ctypes.c_ulonglong),
+            ]
```
lucasmundim: `NvmlUtilization` and `NvmlMemory` are defined inside `get_gpu_metrics()`, so they get recreated on every call. Since this runs in a polling loop (every `GPU_METRICS_INTERVAL` seconds), these should be module-level constants.

openfilter/filter_runtime/gpu.py
```diff
@@ -0,0 +1,273 @@
...
+        for lib_name in _GPU_LIB_NAMES:
+            lib_path = os.path.join(search_dir, lib_name)
+            if lib_path in _preloaded_gpu_libs:
+                continue
+            if os.path.exists(lib_path):
+                try:
+                    ctypes.CDLL(lib_path, mode=ctypes.RTLD_GLOBAL)
+                    _preloaded_gpu_libs.add(lib_path)
+                except OSError as exc:
+                    logger.debug("Failed to preload GPU library %s: %s", lib_path, exc)
```
lucasmundim, line 150: `preload_gpu_libs()` loads the libraries with `RTLD_GLOBAL` but doesn’t populate `_libcuda`/`_libnvml`. When `detect_gpu()` later calls `_load_libcuda()`, it creates a second `ctypes.CDLL` wrapper for the same underlying library. Not a bug (dlopen returns the cached handle), but a minor redundancy — could populate the cached handles here to avoid the double-load.

openfilter/filter_runtime/gpu.py
shingonoide: Fixed - removed the bare-name fallback. _find_gpu_lib now returns None, callers use path or bare name.

openfilter/cli/cmd_run.py
shingonoide: Fixed - moved to _configure_start_method() called at top of cmd_run().

openfilter/filter_runtime/gpu.py
shingonoide: Fixed - moved to module-level _NvmlUtilization and _NvmlMemory.

openfilter/filter_runtime/gpu.py
shingonoide: Fixed - preload_gpu_libs now populates _libcuda/_libnvml cache after successful load.

lucasmundim (approved):
All review comments addressed cleanly. LGTM.

PR #77, opened 2026-04-08
feat(tests): add IPC/serialization overhead benchmarking harness

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Adds a pytest-benchmark harness (`make bench`) that measures the serialization and IPC overhead of the OpenFilter runtime — the cost of moving frames between filters — separately from filter processing time.

9 benchmark groups covering frame serde, JPG codec, JSON metadata, frame format conversions, timing injection, ZMQ IPC round-trips, cv2.resize resampling, resolution scaling, and end-to-end multi-filter pipeline simulations with per-stage overhead breakdowns.

## 🔍 Why is this needed?

The runtime already has per-filter timing instrumentation (filter_timings), but there was no way to isolate the cost of the machinery *between* filters — ZMQ transport, frame serialization, JPG encode/decode, buffer copies — from the filter work itself. Without this you can't tell whether a slow pipeline needs a faster model or a faster transport, and you can't verify that an optimization actually helped.

## 🧪 How was it tested?

- `make bench` runs the full 55-test harness in ~25s
- `make test` skips benchmarks via --benchmark-disable (no regressions)
- ruff check and black --target-version py312 pass clean
- Pipeline simulation models 6 realistic topologies (480p–4K, 2–4 filters, raw and JPG transport) with real ZMQ IPC

## 🔗 Related Issues

FILTER-415 (under FILTER-367: Inference Performance & Scaling)

Baseline pipeline simulation output:

| Pipeline                     | Filter | IPC OH  | OH%  | Max FPS |
|------------------------------|--------|---------|------|---------|
| 480p, 3 filters, raw         | 11.1ms | 1.3ms   | 10%  | 80.9    |
| 4K→1080p, 3 filters, raw     | 20.4ms | 6.4ms   | 24%  | 37.3    |
| 1080p, 4 filters, raw        | 35.2ms | 9.3ms   | 21%  | 22.5    |
| 4K native, 2 filters, raw    | 20.1ms | 34.5ms  | 63%  | 18.3    |

## ✅ Checklist

- [x] I have added or updated documentation as needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,11 @@ build-wheel:  ## Build python wheel
 
 .PHONY: test
 test:  ## Run basic unit tests
-	pytest -v --cov=tests -s tests
+	pytest -v --cov=tests -s tests --benchmark-disable
+
+.PHONY: bench
+bench:  ## Run performance benchmarks
+	pytest tests/test_benchmarks.py -v --benchmark-group-by=group -s --benchmark-max-time=0.5
 
 .PHONY: test-all
 test-all:  ## Run all unit tests
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ build-wheel:  ## Build python wheel
 
 .PHONY: test
 test:  ## Run basic unit tests
-	pytest -v --cov=tests -s tests --benchmark-disable
+	pytest -v --cov=tests -s tests --benchmark-disable -m "not slow"
 
 .PHONY: bench
 bench:  ## Run performance benchmarks
@@ -34,7 +34,7 @@ test-all:  ## Run all unit tests
 .PHONY: test-coverage
 test-coverage:  ## Run unit tests and generate coverage report
 	@mkdir -p Reports
-	@pytest -v --cov=tests --junitxml=Reports/coverage.xml --cov-report=json:Reports/coverage.json
+	@pytest -v --cov=tests --junitxml=Reports/coverage.xml --cov-report=json:Reports/coverage.json --benchmark-disable -m "not slow"
 	@jq -r '["File Name", "Statements", "Missing", "Coverage%"], (.files | to_entries[] | [.key, .value.summary.num_statements, .value.summary.missing_lines, .value.summary.percent_covered_display]) | @csv'  Reports/coverage.json >  Reports/coverage_report.csv
 	@jq -r '["TOTAL", (.totals.num_statements // 0), (.totals.missing_lines // 0), (.totals.percent_covered_display // "0")] | @csv'  Reports/coverage.json >>  Reports/coverage_report.csv
 
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -421,23 +421,26 @@ class TestZmqIpcBenchmarks:
         )
         receiver = MQReceiver(addr, f"{mq_id}-recv")
 
-        # Warm up
-        for _ in range(2):
-            sender.send(frames)
-            result = receiver.recv(timeout=5000)
-            if result is None:
-                pytest.fail("receiver.recv() timed out during warm-up")
+        try:
+            # Warm up
+            for _ in range(2):
+                sender.send(frames)
+                result = receiver.recv(timeout=5000)
+                if result is None:
+                    pytest.fail("receiver.recv() timed out during warm-up")
 
-        t0 = time.perf_counter()
-        for _ in range(n):
-            sender.send(frames)
-            result = receiver.recv(timeout=5000)
-            if result is None:
-                pytest.fail("receiver.recv() timed out during measurement")
-        elapsed = time.perf_counter() - t0
+            t0 = time.perf_counter()
+            for _ in range(n):
+                sender.send(frames)
+                result = receiver.recv(timeout=5000)
+                if result is None:
+                    pytest.fail("receiver.recv() timed out during measurement")
+            elapsed = time.perf_counter() - t0
+        finally:
+            # Cleanup: destroy sender before receiver to avoid ZMQ PUSH blocking
+            sender.destroy()
+            receiver.destroy()
 
-        sender.destroy()
-        receiver.destroy()
         return (elapsed / n) * 1000
 
     @pytest.mark.slow
--- a/tests/test_benchmarks.py
+++ b/tests/test_benchmarks.py
@@ -424,16 +424,20 @@ class TestZmqIpcBenchmarks:
         # Warm up
         for _ in range(2):
             sender.send(frames)
-            receiver.recv()
+            result = receiver.recv(timeout=5000)
+            if result is None:
+                pytest.fail("receiver.recv() timed out during warm-up")
 
         t0 = time.perf_counter()
         for _ in range(n):
             sender.send(frames)
-            receiver.recv()
+            result = receiver.recv(timeout=5000)
+            if result is None:
+                pytest.fail("receiver.recv() timed out during measurement")
         elapsed = time.perf_counter() - t0
 
-        receiver.destroy()
         sender.destroy()
+        receiver.destroy()
         return (elapsed / n) * 1000
 
     @pytest.mark.slow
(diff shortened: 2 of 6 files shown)


lucasmundim (commented):
Well-structured and useful addition — the benchmark groups are logically organized and the pipeline simulation is a clever approach to quantify overhead. Found a few issues worth addressing below.

**Looks good:**
- Benchmark groups cover the right layers (serde, JPG codec, JSON, frame conversion, timing injection, ZMQ IPC, resampling, pipeline simulation)
- `_make_bare_filter` correctly mirrors all Filter instance attributes
- MQ method calls match the real signatures
- `make bench` target with `--benchmark-max-time=0.5` keeps the suite fast
- PR description and overhead table are excellent

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+            print(f"{label:>15} {avg_ms:>12.3f} {pct:>13.1f}%")
+
+        print("=" * 72)
+        logging.disable(logging.NOTSET)
+
+
+# ---------------------------------------------------------------------------
+# Group: overhead_ratio — quantify IPC+serde overhead as % of total per-frame time
+# ---------------------------------------------------------------------------
+
```
lucasmundim, line 481: `test_zmq_roundtrip_report` and `test_pipeline_simulation_report` don't use `@pytest.mark.benchmark` — they use manual `time.perf_counter()` loops. Since `--benchmark-disable` only skips tests that use the `benchmark` fixture, these two slow tests (real ZMQ sockets, `time.sleep` simulations) will execute on every `make test` and in CI.

They should be gated with a custom marker and excluded in the test target, or moved behind a `--run-benchmarks` flag.

Makefile
```diff
@@ -19,7 +19,11 @@ build-wheel:  ## Build python wheel
 
 .PHONY: test
 test:  ## Run basic unit tests
-	pytest -v --cov=tests -s tests
+	pytest -v --cov=tests -s tests --benchmark-disable
```
lucasmundim: `test-coverage` also runs `pytest` without `--benchmark-disable` (line 33, outside this diff). All pytest-benchmark tests will run during coverage, and the two manual-timing tests (`test_zmq_roundtrip_report`, `test_pipeline_simulation_report`) will run regardless since they don't use the benchmark fixture.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+
+class TestPipelineSimulation:
```
lucasmundim, line 483: `logging.disable(logging.CRITICAL)` is process-global — if tests ever run in parallel (pytest-xdist), this will mask logs from other tests. A `caplog` fixture or scoped log-level fixture would be safer.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+            if sender.send(frames, 10):
+                frames = False
+        sender.destroy()
+
+
+# ---------------------------------------------------------------------------
+# Fixtures: realistic frames at varying sizes
+# ---------------------------------------------------------------------------
+
+
```
lucasmundim, line 33: This class is nearly identical to the one in `tests/test_mq.py:20-54` (only differences: method name `_run` vs `thread_func`, and `join(timeout=5)` vs `join()`). Consider extracting it to a shared `conftest.py` fixture or `tests/helpers.py` to avoid drift between the two copies.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+    @pytest.mark.benchmark(group="frame_serde")
+    def test_serialize_1080p_raw(self, benchmark):
+        frames = {"main": FRAME_1080P}
+        benchmark(MQ.frames2topicmsgs, frames, False)
+
+    @pytest.mark.benchmark(group="frame_serde")
+    def test_serialize_1080p_jpg(self, benchmark):
+        frames = {"main": FRAME_1080P}
+        benchmark(MQ.frames2topicmsgs, frames, True)
```
lucasmundim, line 120: These module-level constants allocate ~24 MB of random arrays at import time, even when benchmarks are skipped via `--benchmark-disable`. Moving them into fixtures or using lazy initialization would avoid this overhead during normal test runs.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+
+    @pytest.mark.benchmark(group="frame_sizes")
+    def test_deserialize_720p(self, benchmark):
+        msgs = MQ.frames2topicmsgs({"main": FRAME_720P}, False)
+        benchmark(MQ.topicmsgs2frames, msgs)
+
+    @pytest.mark.benchmark(group="frame_sizes")
+    def test_deserialize_1080p(self, benchmark):
+        msgs = MQ.frames2topicmsgs({"main": FRAME_1080P}, False)
+        benchmark(MQ.topicmsgs2frames, msgs)
```
lucasmundim, line 764: Nit: this assertion is environment-dependent and could flake on constrained CI runners. Consider a more generous threshold or making it a soft check (print a warning instead of asserting).

lucasmundim (dismissed):
### Code review

No blocking issues. Three non-blocking observations:

1. **Teardown order**: receivers are destroyed before senders in `_measure_roundtrip` (L435-436) and `_run_pipeline` (L563-566). If the ZMQ push socket blocks waiting for a peer after the pull socket is gone, the sender thread may not exit within the 5s `join` timeout, silently leaking the thread and socket. Consider destroying senders first.

https://github.com/PlainsightAI/openfilter/blob/0431e93bb5f622ee19b3f0fab1eeefdcb421f447/tests/test_benchmarks.py#L434-L437

2. **`receiver.recv()` has no timeout** in the warm-up and measurement loops. If the sender thread stalls, the main thread blocks indefinitely. A timeout would prevent CI hangs.

https://github.com/PlainsightAI/openfilter/blob/0431e93bb5f622ee19b3f0fab1eeefdcb421f447/tests/test_benchmarks.py#L425-L432

3. **480p overhead regression guard uses `warnings.warn`** instead of `assert` or `pytest.fail`. Since this test only runs via `make bench`, a warning may go unnoticed. Consider a hard assertion if the target matters.

https://github.com/PlainsightAI/openfilter/blob/0431e93bb5f622ee19b3f0fab1eeefdcb421f447/tests/test_benchmarks.py#L706-L712

tests/test_benchmarks.py
```diff
@@ -0,0 +1,829 @@
...
+            sender.send(frames)
+            receiver.recv()
+
+        t0 = time.perf_counter()
+        for _ in range(n):
+            sender.send(frames)
+            receiver.recv()
+        elapsed = time.perf_counter() - t0
+
+        receiver.destroy()
```
lucasmundim: **Still open from prior review — teardown order:** Receivers are destroyed before senders here and in `_run_pipeline` (lines 563–566). If the ZMQ push socket blocks waiting for a peer after the pull socket is already gone, the sender thread may not exit within the 5s `join` timeout, silently leaking the thread and socket fd.

Swap to destroy senders first, then receivers.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,829 @@
...
+
+        # Warm up
+        for _ in range(2):
+            sender.send(frames)
+            receiver.recv()
+
+        t0 = time.perf_counter()
+        for _ in range(n):
+            sender.send(frames)
+            receiver.recv()
```
lucasmundim: **Still open from prior review — no recv timeout:** `receiver.recv()` has no timeout in both the warm-up and measurement loops. If the sender thread stalls (e.g., ZMQ socket error), the main thread blocks indefinitely and CI hangs with no diagnostic output.

A timeout + `pytest.fail` on `None` would prevent silent hangs.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,829 @@
...
+        """Run a simulated pipeline and return timing results.
+
+        Args:
+            stages: list of (name, process_ms, resolution, jpg) tuples.
+                    resolution is (h, w) or None to reuse previous frame.
+                    First stage is the source (process_ms = resample time).
+            n_frames: number of frames to push through.
+            label: human-readable pipeline name.
+        """
+        logging.getLogger().setLevel(logging.CRITICAL)
```
lucasmundim: **Partially addressed:** The commit message says "Replace process-global `logging.disable()` with scoped alternatives" — `test_zmq_roundtrip_report` correctly uses `caplog.set_level()`, but `_run_pipeline` still mutates the root logger globally via `logging.getLogger().setLevel(logging.CRITICAL)` (here) and restores it to `WARNING` on line 567.

This is still process-global: if tests run in parallel (pytest-xdist), it will mask logs from other tests. And the restore sets `WARNING` unconditionally, which may differ from the original level.

Since `_run_pipeline` is a `@staticmethod` called from a test method, consider accepting a `caplog` parameter or using `logging.getLogger('openfilter').setLevel()` on the specific logger that's noisy, rather than the root logger.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,829 @@
...
+            return MQ.topicmsgs2frames(MQ.frames2topicmsgs(frames, True))
+
+        benchmark(roundtrip)
+
+    @pytest.mark.benchmark(group="frame_serde")
+    def test_serialize_4k_raw(self, benchmark):
+        frames = {"main": FRAME_4K()}
+        benchmark(MQ.frames2topicmsgs, frames, False)
+
+    @pytest.mark.benchmark(group="frame_serde")
```
lucasmundim, line 170: Nit: `@functools.cache` on zero-arg functions works but is unconventional — these look like constants (`FRAME_480P()`) but are function calls. A common alternative is `@functools.lru_cache(maxsize=1)` which signals "compute once" more explicitly. Minor — no action needed.

lucasmundim (commented):
Good progress — the feedback from the first review was well addressed: slow tests are gated with `@pytest.mark.slow`, `ThreadMQSender` is extracted to `helpers.py`, frame constants use `@functools.cache`, the flaky assertion is now a warning, and both `test` and `test-coverage` Makefile targets exclude benchmarks.

Three items remain from the prior review that weren't addressed, plus one new observation.

shingonoide (changes requested):
Well-structured benchmarking harness with excellent coverage - 55 tests across 9 groups, proper pytest-benchmark integration, clean ThreadMQSender refactoring. Two critical issues to address:

**CRITICAL - Teardown order causes thread/socket leak:**
In `_measure_roundtrip` (L537-538) and `_run_pipeline` (L564-567), receivers are destroyed before senders. ZMQ PUSH sockets block when the PULL peer disappears, causing the sender thread to hang past the 5s join timeout, silently leaking threads and socket file descriptors. This accumulates across the 7 scenarios in `test_zmq_roundtrip_report` and 6 pipelines in `test_pipeline_simulation_report`. Fix: destroy senders first, then receivers.

**CRITICAL - No recv timeout can hang CI:**
`receiver.recv()` is called with no timeout in both warm-up and measurement loops. If the sender thread stalls, the main test thread blocks indefinitely with no diagnostic output, causing CI to hang until job timeout. Fix: add a timeout to `receiver.recv()` and `pytest.fail()` on None.

**MEDIUM - Global logging mutation not scoped:**
`_run_pipeline` (L602, L668) mutates root logger level to CRITICAL and restores to hardcoded WARNING unconditionally. If the method raises, restore is never reached. Fix: use try/finally with the original level, or `caplog.set_level()` as done for `test_zmq_roundtrip_report`.

**Positive:**
- Excellent benchmark isolation from regular test suite
- Sound methodology with real IPC paths
- Good resource cleanup patterns with daemon threads as safety net
- Pipeline simulation covering real topologies is valuable

tests/test_benchmarks.py
stwilt: Already addressed — both tests use `@pytest.mark.slow` and both `make test` ([`Makefile` L22](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/Makefile#L22)) and `test-coverage` ([`Makefile` L37](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/Makefile#L37)) exclude them via `-m "not slow"` + `--benchmark-disable`.

Makefile
stwilt: Already addressed — `test-coverage` now includes `--benchmark-disable -m "not slow"`: [`Makefile` L37](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/Makefile#L37).

tests/test_benchmarks.py
stwilt: Fixed — `_run_pipeline` now scopes to `logging.getLogger("openfilter")` instead of the root logger, saves/restores the original level, and wraps the body in `try/finally` for guaranteed restore: [`tests/test_benchmarks.py` L505-L507](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L505-L507) and [`tests/test_benchmarks.py` L574-L580](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L574-L580).

tests/test_benchmarks.py
stwilt: Already addressed in a prior commit — `ThreadMQSender` is extracted to `tests/helpers.py`.

tests/test_benchmarks.py
stwilt: Already addressed — these are `@functools.cache` zero-arg functions, so they only allocate on first call. With `--benchmark-disable -m "not slow"`, no test in this file runs during `make test`, so the functions are never invoked and nothing is allocated.

tests/test_benchmarks.py
stwilt: Already addressed — this is a `warnings.warn` (soft check), not an assertion.

tests/test_benchmarks.py
stwilt: Fixed — senders are now destroyed before receivers in both `_measure_roundtrip` ([`tests/test_benchmarks.py` L439-L440](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L439-L440)) and `_run_pipeline` ([`tests/test_benchmarks.py` L575-L579](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L575-L579)).

tests/test_benchmarks.py
stwilt: Fixed — all `receiver.recv()` calls now use `timeout=5000` with `pytest.fail()` on `None`:

- `_measure_roundtrip` warm-up: [`tests/test_benchmarks.py` L427-L429](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L427-L429)
- `_measure_roundtrip` measurement: [`tests/test_benchmarks.py` L434-L436](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L434-L436)
- `_run_pipeline` warm-up: [`tests/test_benchmarks.py` L534-L536](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L534-L536)
- `_run_pipeline` measurement: [`tests/test_benchmarks.py` L562-L564](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L562-L564)

tests/test_benchmarks.py
stwilt: Fixed — now scoped to `logging.getLogger("openfilter")` with original level saved/restored, wrapped in `try/finally`: [`tests/test_benchmarks.py` L505-L507](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L505-L507) and [`tests/test_benchmarks.py` L574-L580](https://github.com/PlainsightAI/openfilter/blob/aea3ef5b2f3ca54119bff31ded632de16fefdfcd/tests/test_benchmarks.py#L574-L580).

tests/test_benchmarks.py
stwilt: Acknowledged — keeping `@functools.cache` as-is. Agreed no action needed.

shingonoide:
All 3 findings addressed cleanly:

1. **Teardown order** — senders destroyed before receivers in both `_measure_roundtrip` and `_run_pipeline`. Wrapped in `finally` for guaranteed cleanup.
2. **Recv timeout** — all `recv()` calls now have `timeout=5000` with `pytest.fail()` on None. No more silent CI hangs.
3. **Scoped logging** — uses `openfilter` logger instead of root, saves original level, restores in `finally` block.

LGTM, approving.

shingonoide (dismissed):
All review findings addressed. Clean fixes.

tests/test_benchmarks.py
```diff
@@ -0,0 +1,842 @@
...
+        t0 = time.perf_counter()
+        for _ in range(n):
+            sender.send(frames)
+            result = receiver.recv(timeout=5000)
+            if result is None:
+                pytest.fail("receiver.recv() timed out during measurement")
+        elapsed = time.perf_counter() - t0
+
+        sender.destroy()
+        receiver.destroy()
```
lucasmundim: Nit (non-blocking): unlike `_run_pipeline`, this cleanup isn't wrapped in `try/finally`. If `pytest.fail()` fires from the timeout guard above (warm-up or measurement), `sender.destroy()` / `receiver.destroy()` are skipped and the sockets rely on the daemon-thread/process-exit safety net. For consistency with `_run_pipeline`, consider wrapping the sender/receiver setup and the loops in `try/finally`.

lucasmundim (dismissed):
Re-review of `aea3ef5`: all three prior findings addressed cleanly.

- Teardown order fixed in both `_measure_roundtrip` and `_run_pipeline` (senders destroyed first).
- `receiver.recv()` now uses `timeout=5000` with `pytest.fail()` on `None` in every warm-up and measurement loop, with hop/phase context in the failure message.
- `_run_pipeline` now scopes logging to the `openfilter` logger, saves the original level, and restores it in `finally`.

LGTM. One minor non-blocking nit below.

lucasmundim (commented):
`d00f5a9` addresses the `_measure_roundtrip` nit — warm-up/measurement loops wrapped in `try` with sender/receiver destroys (correct order) moved into `finally`. Consistent with `_run_pipeline` now. LGTM.

lucasmundim (approved):
LGTM on `d00f5a9`.

PR #79, opened 2026-04-13
feat(tracing): add OpenTelemetry distributed tracing for per-filter execution spans

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Adds OpenTelemetry distributed tracing to the openfilter runtime so that per-filter `process()` calls emit spans, and so that filter pods can pick up a parent trace context from their environment.

Concretely:

- New `openfilter/observability/tracing.py` module exposing:
  - `build_span_exporter(type, **config)` — factory with `console`, `silent`, `otlp_grpc`, `otlp_http` branches; unknown values fall back to console (never crashes a filter at startup).
  - `extract_parent_context()` — reads the W3C `TRACEPARENT` env var (set by e.g. a pipelines controller from a CR annotation) and returns an OTel Context that new spans can start under as children.
  - `setup_tracer_provider(resource, exporter_type, exporter_config)` — creates and registers a `TracerProvider` sharing the same `Resource` the metrics pipeline uses, wrapped in a `BatchSpanProcessor`.
- `openfilter/observability/client.py` — `OpenTelemetryClient` now builds a tracer alongside its meter, using the same `exporter_type` / `exporter_config` contract as the metrics side so nothing new has to be configured by operators.
- `openfilter/filter_runtime/filter.py` — `process_frames()` now wraps the user's `process()` call in a span when a tracer is available, attributing each span with `filter.name`, `filter.id`, and `pipeline.id` so spans land in the right service tile in the trace UI.
- Version bump: `VERSION` → `v0.1.28`, matching `RELEASE.md` entry, so the manual create-release workflow can publish.

---

## 🔍 Why is this needed?

Plainsight's platform team is rolling out end-to-end distributed tracing so a user request that enters plainsight-api can be followed all the way into the filter pods running user workloads, stitched into a single trace in Cloud Trace. openfilter is the leaf of that chain — without it emitting spans, the trace stops at the pipelines-controller hop and we lose visibility into actual frame processing.

The defaults are deliberately conservative so this is a no-op for anyone not actively opting in:

- Default exporter type is `silent`, which accepts spans and drops them.
- Operators opt in by setting `TELEMETRY_EXPORTER_TYPE=otlp_grpc` and `TELEMETRY_EXPORTER_OTLP_ENDPOINT=<collector host:port>` — the same env-var contract the existing metrics exporter uses.
- The OTLP gRPC exporter defaults to `insecure=True` so it works against plaintext in-cluster collectors out of the box. Operators who need TLS can pass `insecure=False` via `exporter_config`.

---

## 🧪 How was it tested?

- **`tests/test_tracing.py`** — 16 unit tests covering:
  - every branch of `build_span_exporter` (console, silent, otlp_grpc, otlp_http, unknown-falls-back-to-console, case-insensitivity)
  - OTLP gRPC default endpoint + `TELEMETRY_EXPORTER_OTLP_ENDPOINT` env override
  - **insecure-default regression** (the Python OTLP exporter defaults to TLS unless told otherwise; this test pins the insecure default so nobody accidentally breaks plaintext in-cluster export)
  - `extract_parent_context` unset / set / set-with-tracestate
  - tracer provider creation including a real-span-produced check
- **`tests/integration/test_tracing_export.py`** — 2 integration tests, gated behind the existing `slow` pytest marker and a new `make test-integration` target. They spin up `jaegertracing/all-in-one` via a module-scoped docker fixture (plain `subprocess`, no new dependency), emit real spans through the OTLP gRPC exporter, poll the jaeger query API, and assert the service name appears in the returned trace. One of the two also verifies `TRACEPARENT` propagation end-to-end.
- **Default `make test` is unchanged.** The integration tests are deselected by `-m "not slow"` and require docker to run.
- Full unit suite runs green locally: 434 passed (5 pre-existing environment-only failures in `test_cli.py` and `test_filter.py` that reproduce on main and are unrelated to this PR — can share details on request).

---

## 🔗 Related Issues

- PLAT-848 (this PR, sub-task)
- Parent: PLAT-827 — OTel tracing: pipeline instance and benchmark path
- Related: PLAT-847 (wave 3 service instrumentation rollout; its three sub-tasks depend on this library release being available to consume)
- A follow-up ticket PLAT-845 tracks promoting a multi-repo chain test (not upstreamed here, lives as a local dev harness) into CI, targeted at Goldenrod.2.

---

Excerpt from a passing local run of the integration tests against a fresh jaeger container:

```
tests/integration/test_tracing_export.py::test_otlp_grpc_export_reaches_jaeger PASSED [ 50%]
tests/integration/test_tracing_export.py::test_traceparent_env_is_honored PASSED      [100%]
============================== 2 passed in 10.88s ==============================
```

And the unit suite with the new file in place:

```
tests/test_tracing.py ................                                     [100%]
============================== 16 passed in 0.20s ==============================
```

---

## ✅ Checklist

> 📄 **DCO Reminder:**
> Confirmed: `git log -1 --format="%(trailers)" HEAD` shows `Signed-off-by: stwilt <swilt@plainsight.ai>` on the single commit.

### DIFF
--- a/openfilter/filter_runtime/filter.py
+++ b/openfilter/filter_runtime/filter.py
@@ -1028,9 +1028,12 @@ class Filter:
     def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callable[[], dict[str, Frame] | None] | None:
         """Call process() and deal with it if returns a Callable."""
 
-        # Wrap process() with timing and optional tracing span
-        tracer = getattr(getattr(self, 'otel', None), 'tracer', None)
-        parent_ctx = getattr(getattr(self, 'otel', None), 'parent_context', None) if tracer else None
+        # Wrap process() with timing and optional tracing span.
+        # Timers sit inside the span context manager so per-filter timing
+        # metrics reflect pure process() cost, not span creation overhead.
+        otel = getattr(self, 'otel', None)
+        tracer = getattr(otel, 'tracer', None) if otel else None
+        parent_ctx = getattr(otel, 'parent_context', None) if tracer else None
         span_ctx_mgr = (
             tracer.start_as_current_span(
                 f"{self.__class__.__name__}.process",
@@ -1056,10 +1059,10 @@ class Filter:
             if tracer
             else contextlib.nullcontext()
         )
-        t_in = time.time()
         with span_ctx_mgr:
+            t_in = time.time()
             processed_frames = self.process(frames)
-        t_out = time.time()
+            t_out = time.time()
 
         # Update per-filter timing unless process() returned a callable,
         # in which case _timed_callable will record the real duration later.
--- a/openfilter/filter_runtime/filter.py
+++ b/openfilter/filter_runtime/filter.py
@@ -1,3 +1,4 @@
+import contextlib
 import json
 import logging
 import multiprocessing as mp
@@ -73,6 +74,28 @@ preload_gpu_libs()  # Must run BEFORE any torch import
 _apply_append_env_vars()  # Still useful for child processes and PATH
 
 logger = logging.getLogger(__name__)
+
+
+def _flush_tracer_provider(otel: Any) -> None:
+    """Force-flush and shut down the OTel TracerProvider attached to an
+    OpenTelemetryClient, if any. Safe to call when ``otel`` is None, when it
+    has no ``tracer_provider`` attribute, or when ``tracer_provider`` is None
+    (all of which happen when tracing is disabled). Logs but never re-raises
+    — this runs in shutdown / exception-handling paths and must not mask the
+    original exception."""
+
+    tracer_provider = getattr(otel, 'tracer_provider', None)
+    if tracer_provider is None:
+        return
+    try:
+        tracer_provider.force_flush(timeout_millis=5000)
+    except Exception as e:
+        logger.warning("tracer provider force_flush failed during shutdown: %s", e)
+    try:
+        tracer_provider.shutdown()
+    except Exception as e:
+        logger.warning("tracer provider shutdown failed: %s", e)
+
 
 scarf_elogger = ScarfEventLogger(
     endpoint_url=f"https://python.openfilter.io/openfilter"
@@ -1005,9 +1028,37 @@ class Filter:
     def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callable[[], dict[str, Frame] | None] | None:
         """Call process() and deal with it if returns a Callable."""
 
-        # Wrap process() with timing
+        # Wrap process() with timing and optional tracing span
+        tracer = getattr(getattr(self, 'otel', None), 'tracer', None)
+        parent_ctx = getattr(getattr(self, 'otel', None), 'parent_context', None) if tracer else None
+        span_ctx_mgr = (
+            tracer.start_as_current_span(
+                f"{self.__class__.__name__}.process",
+                context=parent_ctx,
+                attributes={
+                    "filter.name": self.filter_name or self.__class__.__name__,
+                    "filter.id": self._filter_id or "",
+                    # `pipeline.id` is retained for backward compat with any
+                    # existing dashboards keyed on it, but its value is
+                    # actually the pipeline *instance* name (from the
+                    # PIPELINE_ID env var injected by the pipelines-controller
+                    # via `pipelineInstance.Name`). The canonical grouping
+                    # attribute used consistently across plainsight-api,
+                    # plainsight-deployment-agent, and here is
+                    # `pipeline_instance.id`, so we stamp both with the same
+                    # value. A single Cloud Trace query
+                    # `pipeline_instance.id = "<uuid>"` returns the full
+                    # api → agent → filter waterfall end-to-end.
+                    "pipeline.id": self.pipeline_id or "",
+                    "pipeline_instance.id": self.pipeline_id or "",
+                },
+            )
+            if tracer
+            else contextlib.nullcontext()
+        )
         t_in = time.time()
-        processed_frames = self.process(frames)
+        with span_ctx_mgr:
+            processed_frames = self.process(frames)
         t_out = time.time()
 
         # Update per-filter timing unless process() returned a callable,
@@ -1383,6 +1434,14 @@ class Filter:
 
     def shutdown(self) -> None:
         """Clean up resources used."""
+        # Flush any buffered spans to the OTel collector before the container exits.
+        # The BatchSpanProcessor in setup_tracer_provider() buffers spans on an
+        # interval; without an explicit flush on shutdown, spans emitted just
+        # before exit can be dropped silently. Gated on self.otel being
+        # initialised so filters with tracing disabled remain a no-op. Logs but
+        # never re-raises — shutdown runs in error paths and must not mask the
+        # original exception.
+        _flush_tracer_provider(getattr(self, 'otel', None))
 
     def process(self, frames: dict[str, Frame]) -> dict[str, Frame] | Frame | Callable[[], dict[str, Frame] | Frame | None] | None:
         """Main processing thingy, this is the only method which MUST be implemented by a user Filter.
--- a/openfilter/observability/client.py
+++ b/openfilter/observability/client.py
@@ -25,6 +25,7 @@ from opentelemetry.sdk.metrics.export import PeriodicExportingMetricReader
 from opentelemetry.metrics import Observation, set_meter_provider, get_meter
 
 from .bridge import OTelLineageExporter
+from .tracing import setup_tracer_provider, extract_parent_context
 
 
 def strtobool(val: str) -> bool:
@@ -191,11 +192,29 @@ class OpenTelemetryClient:
                         self.business_meter = None
                 else:
                     self.business_meter = None
+                # --- Distributed tracing ---
+                try:
+                    self.tracer_provider = setup_tracer_provider(
+                        resource=resource,
+                        exporter_type=exporter_type,
+                        exporter_config=exporter_config,
+                    )
+                    self.parent_context = extract_parent_context()
+                    self.tracer = self.tracer_provider.get_tracer(service_name)
+                except Exception as e:
+                    logging.error(f"Failed to initialise tracing: {e}")
+                    self.tracer_provider = None
+                    self.parent_context = None
+                    self.tracer = None
+
             except Exception as e:
                 logging.error(f"Error setting Open Telemetry: {e}")
         else:
             self.provider = None
             self.meter = None
+            self.tracer_provider = None
+            self.parent_context = None
+            self.tracer = None
             logging.info("telemetry is disabled")
 
         self._lock = threading.Lock()
--- a/openfilter/observability/tracing.py
+++ b/openfilter/observability/tracing.py
@@ -1,0 +1,142 @@
+"""
+OpenTelemetry distributed tracing for OpenFilter pipeline execution.
+
+Provides TracerProvider setup, parent trace context extraction from the
+TRACEPARENT environment variable, and a span exporter factory that mirrors
+the metrics ExporterFactory pattern.
+"""
+
+import os
+import logging
+from typing import Optional
+
+from opentelemetry import context, trace
+from opentelemetry.context.context import Context
+from opentelemetry.propagate import extract
+from opentelemetry.sdk.resources import Resource
+from opentelemetry.sdk.trace import TracerProvider
+from opentelemetry.sdk.trace.export import (
+    BatchSpanProcessor,
+    ConsoleSpanExporter,
+    SpanExporter,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def build_span_exporter(exporter_type: str, **config) -> SpanExporter:
+    """Build a span exporter matching the metrics exporter factory pattern.
+
+    Supported types: console, silent, otlp, otlp_grpc, otlp_http.
+    """
+    exporter_type = exporter_type.lower()
+
+    if exporter_type == "console":
+        return ConsoleSpanExporter()
+
+    if exporter_type == "silent":
+        from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
+
+        class SilentSpanExporter(SpanExporter):
+            def export(self, spans):
+                return SpanExportResult.SUCCESS
+
+            def shutdown(self):
+                pass
+
+            def force_flush(self, timeout_millis=30000):
+                return True
+
+        return SilentSpanExporter()
+
+    if exporter_type in ("otlp", "otlp_grpc"):
+        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
+            OTLPSpanExporter as OTLPGrpcSpanExporter,
+        )
+
+        endpoint = (
+            config.get("endpoint")
+            or os.getenv("TELEMETRY_EXPORTER_OTLP_ENDPOINT")
+            or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
+            or "http://localhost:4317"
+        )
+        # The Plainsight in-cluster collector and the upstream OTel collector
+        # default to plaintext gRPC on 4317. The Python exporter defaults to TLS
+        # unless told otherwise, which silently breaks every export with a TLS
+        # handshake error. Mirror the Go services (otlptracegrpc.WithInsecure)
+        # and opt out by default; operators who need TLS can pass insecure=False
+        # via exporter_config.
+        insecure = config.get("insecure", True)
+        return OTLPGrpcSpanExporter(endpoint=endpoint, insecure=insecure)
+
+    if exporter_type == "otlp_http":
+        from opentelemetry.exporter.otlp.proto.http.trace_exporter import (
+            OTLPSpanExporter as OTLPHttpSpanExporter,
+        )
+
+        endpoint = (
+            config.get("endpoint")
+            or os.getenv("OTEL_EXPORTER_OTLP_HTTP_ENDPOINT")
+            or "http://localhost:4318"
+        )
+        return OTLPHttpSpanExporter(endpoint=endpoint)
+
+    # Fall back to console
+    logger.warning(
+        "Unknown tracing exporter type %r, falling back to console", exporter_type
+    )
+    return ConsoleSpanExporter()
+
+
+def extract_parent_context() -> Optional[Context]:
+    """Extract parent trace context from the TRACEPARENT environment variable.
+
+    The pipeline controller sets TRACEPARENT on the pod from the CR annotation
+    ``traces.opentelemetry.io/traceparent``.  We use the W3C propagator to
+    parse it into an OTel context so that all spans in this process are
+    children of the orchestration trace.
+
+    Returns the extracted context, or ``None`` if TRACEPARENT is not set.
+    """
+    traceparent = os.getenv("TRACEPARENT")
+    if not traceparent:
+        return None
+
+    carrier = {"traceparent": traceparent}
+    # Also pick up tracestate if available
+    tracestate = os.getenv("TRACESTATE")
+    if tracestate:
+        carrier["tracestate"] = tracestate
+
+    ctx = extract(carrier)
+    logger.info("Extracted parent trace context from TRACEPARENT=%s", traceparent)
+    return ctx
+
+
+def setup_tracer_provider(
+    resource: Resource,
+    exporter_type: str = "silent",
+    exporter_config: Optional[dict] = None,
+) -> TracerProvider:
+    """Create and register a TracerProvider sharing the same Resource as metrics.
+
+    Args:
+        resource: OTel Resource (reused from MeterProvider).
+        exporter_type: Exporter type string (same values as metrics).
+        exporter_config: Extra kwargs forwarded to the span exporter builder.
+
+    Returns:
+        The configured TracerProvider.
+    """
+    exporter_config = exporter_config or {}
+    span_exporter = build_span_exporter(exporter_type, **exporter_config)
+    processor = BatchSpanProcessor(span_exporter)
+
+    provider = TracerProvider(resource=resource)
+    provider.add_span_processor(processor)
+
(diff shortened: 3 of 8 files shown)


openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+            def export(self, spans):
+                return SpanExportResult.SUCCESS
+
+            def shutdown(self):
+                pass
+
+            def force_flush(self, timeout_millis=30000):
+                return True
+
+        return SilentSpanExporter()
```
lucasmundim: **`SilentSpanExporter` defined inside `build_span_exporter` — consider promoting**

This class is defined inline inside the `if exporter_type == "silent"` branch, so a new instance (and a new class object) is created on every call. In practice this is called once at startup so it doesn't matter, but:

1. It's not importable/testable independently
2. If someone greps for `SilentSpanExporter` to understand the codebase, they won't find it at module level

Consider promoting it to a module-level class. Minor — the current approach works fine.

openfilter/filter_runtime/filter.py
```diff
@@ -1001,9 +1002,25 @@ def _inject_timings(self, frames: dict[str, Frame], t_in: float, t_out: float, d
...
+                    "filter.id": self._filter_id or "",
+                    "pipeline.id": self.pipeline_id or "",
+                },
+            )
+            if tracer
+            else contextlib.nullcontext()
+        )
         t_in = time.time()
-        processed_frames = self.process(frames)
+        with span_ctx_mgr:
```
lucasmundim: **Span is created on every `process()` call — hot path consideration**

For streaming filters processing 30fps video, this creates a span per frame (or per batch depending on the filter). With `BatchSpanProcessor` and the `silent` exporter the cost is near-zero, but with `otlp_grpc` this could produce a very high span volume.

Not an issue for v1 (operators who enable tracing accept the overhead), but worth keeping in mind. A future enhancement could be sampling at this level — e.g., only span every Nth frame, or use the OTel SDK's built-in `TraceIdRatioBasedSampler`.

Also: the double `getattr` chain (`getattr(getattr(self, 'otel', None), 'tracer', None)`) is safe but a bit hard to read. Works fine since `otel` may not always be set on the filter instance.

openfilter/observability/client.py
```diff
@@ -191,11 +192,29 @@ def __init__(
...
+                        exporter_config=exporter_config,
+                    )
+                    self.parent_context = extract_parent_context()
+                    self.tracer = self.tracer_provider.get_tracer(service_name)
+                except Exception as e:
+                    logging.error(f"Failed to initialise tracing: {e}")
+                    self.tracer_provider = None
+                    self.parent_context = None
+                    self.tracer = None
+
```
lucasmundim, line 209: **Good: tracing failure is caught independently and doesn't take down metrics**

The inner `try/except` around `setup_tracer_provider` means a tracing initialization failure (e.g., bad exporter config) logs an error and sets `self.tracer = None`, but the metrics pipeline — which was already initialized above — continues working. This is the right design for a library where metrics and tracing are orthogonal opt-ins.

One edge case: if `setup_tracer_provider` succeeds but `extract_parent_context()` raises (unlikely, but could happen with a malformed `TRACEPARENT`), the `tracer_provider` is set but `self.tracer` is set to `None`, orphaning the provider without a shutdown. Low risk since `extract_parent_context` is purely string parsing and unlikely to raise.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+        The configured TracerProvider.
+    """
+    exporter_config = exporter_config or {}
+    span_exporter = build_span_exporter(exporter_type, **exporter_config)
+    processor = BatchSpanProcessor(span_exporter)
+
+    provider = TracerProvider(resource=resource)
+    provider.add_span_processor(processor)
+
+    trace.set_tracer_provider(provider)
```
lucasmundim: **`trace.set_tracer_provider(provider)` is a global singleton**

The OTel Python SDK's `set_tracer_provider` is a one-shot: calling it again after the first call is a no-op (silently ignored). This means if multiple filters somehow share a process and each calls `setup_tracer_provider`, only the first one wins.

In practice this is fine — openfilter runs one filter per container. But the unit tests work around this correctly (the comment in `test_returned_provider_produces_real_spans` acknowledges the one-shot behavior and tests the returned provider directly rather than the global). Good awareness.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+            or "http://localhost:4317"
+        )
+        # The Plainsight in-cluster collector and the upstream OTel collector
+        # default to plaintext gRPC on 4317. The Python exporter defaults to TLS
+        # unless told otherwise, which silently breaks every export with a TLS
+        # handshake error. Mirror the Go services (otlptracegrpc.WithInsecure)
+        # and opt out by default; operators who need TLS can pass insecure=False
+        # via exporter_config.
+        insecure = config.get("insecure", True)
+        return OTLPGrpcSpanExporter(endpoint=endpoint, insecure=insecure)
```
lucasmundim, line 92: **Env var precedence chain for gRPC endpoint**

The order is: `config.get("endpoint")` -> `TELEMETRY_EXPORTER_OTLP_ENDPOINT` -> `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` -> `http://localhost:4317`.

Note that `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` is not a standard OTel env var — the standard one is `OTEL_EXPORTER_OTLP_ENDPOINT` (without `_GRPC`). The Python OTel SDK actually respects `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` natively, but since this code sets the endpoint explicitly, the SDK's own env var handling is bypassed.

This is fine — `TELEMETRY_EXPORTER_OTLP_ENDPOINT` is the openfilter convention (matching the metrics side), and `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` is a fallback. Just worth documenting that the standard `OTEL_EXPORTER_OTLP_ENDPOINT` is *not* checked here, in case an OSS user expects it to work.

lucasmundim (commented):
Well-structured PR. Evaluated with the OSS lens — openfilter is used beyond Plainsight's environments, so defaults and opt-in behavior matter.

**What looks good (OSS perspective):**
- Default exporter is `silent` — zero behavioral change for anyone not opting in. No new mandatory dependencies, no spans shipped unless explicitly configured
- Unknown exporter types fall back to `console` with a warning instead of crashing — this is exactly right for a library that runs inside user workloads
- Shares the same `Resource` with the metrics pipeline — spans and metrics land under the same service tile in any trace backend
- `TRACEPARENT` env var is the W3C standard mechanism — works with any OTel-compatible orchestrator, not just Plainsight's controller
- `insecure=True` default for gRPC is documented with the rationale, and overridable via `exporter_config`
- Integration tests use plain `subprocess` + `urllib` — no new `docker-py` or `testcontainers` dependency. Clean choice for an OSS project
- `make test` unchanged, integration tests gated behind `slow` marker + separate `make test-integration` target
- Test coverage is thorough: 16 unit tests + 2 integration tests covering the full export chain
- CI green across Python 3.10-3.13

Five observations inline — nothing blocking.

openfilter/filter_runtime/filter.py
```diff
@@ -1001,9 +1024,25 @@ def _inject_timings(self, frames: dict[str, Frame], t_in: float, t_out: float, d
...
+                attributes={
+                    "filter.name": self.filter_name or self.__class__.__name__,
+                    "filter.id": self._filter_id or "",
+                    "pipeline.id": self.pipeline_id or "",
+                },
+            )
+            if tracer
+            else contextlib.nullcontext()
+        )
         t_in = time.time()
```
lucasmundim: **Timing metrics include span overhead**

`t_in` is recorded before the `with span_ctx_mgr:` block, but `t_out` is after both the span and `process()` complete. This means the per-filter timing metrics will include span creation/context-propagation overhead.

On hot paths this is probably sub-microsecond, but if timing precision matters the timer could move inside the context manager:

```python
with span_ctx_mgr:
    t_in = time.time()
    processed_frames = self.process(frames)
    t_out = time.time()
```

Not blocking — just flagging the coupling between tracing and timing.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+
+            def shutdown(self):
+                pass
+
+            def force_flush(self, timeout_millis=30000):
+                return True
+
+        return SilentSpanExporter()
+
+    if exporter_type in ("otlp", "otlp_grpc"):
```
lucasmundim, line 65: **`otlp_grpc` / `otlp_http` type strings diverge from the metrics factory**

The existing metrics `build_exporter` in `client.py` only supports `"otlp"` (mapped to gRPC). This new factory adds `otlp_grpc`, `otlp_http`, and keeps `otlp` as an alias — a better API, but the two factories now accept different type strings.

If someone passes `"otlp_http"` as `TELEMETRY_EXPORTER_TYPE`, traces will go to the HTTP exporter while metrics will silently fall through to the console fallback. Not a bug today since this is unlikely, but a follow-up to align the metrics factory (or at least a comment here noting the divergence) would prevent confusion.

lucasmundim (commented):
Two additional observations not covered in the previous review.

openfilter/filter_runtime/filter.py
```diff
@@ -70,6 +71,28 @@ def _apply_append_env_vars():
...
+    if tracer_provider is None:
+        return
+    try:
+        tracer_provider.force_flush(timeout_millis=5000)
+    except Exception as e:
+        logger.warning("tracer provider force_flush failed during shutdown: %s", e)
+    try:
+        tracer_provider.shutdown()
+    except Exception as e:
+        logger.warning("tracer provider shutdown failed: %s", e)
```
lucasmundim, line 97: **Good addition: graceful tracer shutdown**

This addresses a real problem — `BatchSpanProcessor` buffers spans on a timer, and without explicit flush/shutdown, the last batch of spans before container exit can be silently dropped. The defensive coding is solid:
- Safe when `otel` is None or has no `tracer_provider`
- Separate try/except for flush and shutdown (flush failure shouldn't skip shutdown)
- Logs but never re-raises — critical since this runs in error/shutdown paths

This is something the plainsight-api PR (#657) is still missing (stale shutdown context). Worth cross-referencing.

openfilter/filter_runtime/filter.py
```diff
@@ -1001,9 +1024,37 @@ def _inject_timings(self, frames: dict[str, Frame], t_in: float, t_out: float, d
...
+                    # PIPELINE_ID env var injected by the pipelines-controller
+                    # via `pipelineInstance.Name`). The canonical grouping
+                    # attribute used consistently across plainsight-api,
+                    # plainsight-deployment-agent, and here is
+                    # `pipeline_instance.id`, so we stamp both with the same
+                    # value. A single Cloud Trace query
+                    # `pipeline_instance.id = "<uuid>"` returns the full
+                    # api → agent → filter waterfall end-to-end.
+                    "pipeline.id": self.pipeline_id or "",
+                    "pipeline_instance.id": self.pipeline_id or "",
```
lucasmundim, line 1056: **`pipeline_instance.id` — consistent end-to-end attribute**

Stamping both `pipeline.id` (legacy) and `pipeline_instance.id` (canonical) is the right call. The comment explaining the naming history and cross-repo consistency is helpful for future contributors.

One note: `self.pipeline_id` comes from the `PIPELINE_ID` env var, which — as flagged on the controller PR #43 — is injected by both the deployment agent *and* the controller with potentially different values. Whichever value wins at runtime is what lands here.

lucasmundim (commented):
**Re-review after latest push.** Three substantive additions since last review:

1. **`_flush_tracer_provider()` + shutdown integration** — new module-level function that force-flushes and shuts down the TracerProvider, called from `Filter.shutdown()`. Handles exceptions gracefully (logs, never re-raises). This addresses the "spans dropped on exit" risk inherent in `BatchSpanProcessor`.
2. **`pipeline_instance.id` span attribute** — `process_frames()` now stamps both `pipeline.id` (legacy/backward-compat) and `pipeline_instance.id` (canonical end-to-end key) on every span. Consistent with the same attribute name used in plainsight-api and deployment-agent.
3. **Significantly expanded tests** — `test_tracing.py` grew from 185 to 321 lines, with new tests for `_flush_tracer_provider` (including the broken-provider-swallows-exceptions case) and `TestProcessFramesSpanAttributes` verifying both pipeline attributes are stamped.

Previous review observations still apply. New observations inline.

lucasmundim (dismissed):
Clean addition of `pipeline_instance.id` span attribute for end-to-end Cloud Trace queries, with backward-compat `pipeline.id` retained. Test coverage is thorough. Previous non-blocking observations still apply but nothing blocking.

lucasmundim (commented):
Six non-blocking nits on the current code — none of these need to block merge.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+    """
+    exporter_type = exporter_type.lower()
+
+    if exporter_type == "console":
+        return ConsoleSpanExporter()
+
+    if exporter_type == "silent":
+        from opentelemetry.sdk.trace.export import SpanExporter, SpanExportResult
+
+        class SilentSpanExporter(SpanExporter):
```
lucasmundim: nit: `SilentSpanExporter` is defined inline inside the function body, creating a new class object on every call. Works fine since it's only called once at startup, and the metrics side does the same with `SilentMetricExporter`. Moving it to module level would be slightly cleaner but purely cosmetic.

openfilter/filter_runtime/filter.py
```diff
@@ -1001,9 +1024,37 @@ def _inject_timings(self, frames: dict[str, Frame], t_in: float, t_out: float, d
     def process_frames(self, frames: dict[str, Frame]) -> dict[str, Frame] | Callable[[], dict[str, Frame] | None] | None:
         """Call process() and deal with it if returns a Callable."""
 
-        # Wrap process() with timing
+        # Wrap process() with timing and optional tracing span
+        tracer = getattr(getattr(self, 'otel', None), 'tracer', None)
+        parent_ctx = getattr(getattr(self, 'otel', None), 'parent_context', None) if tracer else None
+        span_ctx_mgr = (
+            tracer.start_as_current_span(
```
lucasmundim, line 1038: nit: This creates a span on every `process()` call. For a 30fps video filter that's 30 spans/second per filter. `BatchSpanProcessor` batches the export, but span objects are still created individually per frame. With `silent` (default) this is near-zero cost; with `otlp_grpc` it could add up on high-throughput pipelines. An opt-in sampling strategy or config toggle could help if this becomes a concern — but it's a reasonable tradeoff for now (more spans = better observability).

openfilter/observability/tracing.py
lucasmundim: nit: `trace.set_tracer_provider()` is a process-wide one-shot — the OTel SDK silently ignores subsequent calls. If two `OpenTelemetryClient` instances were ever created in the same process, the second provider would only work via the returned object, not via the global `trace.get_tracer()`. In practice there's one client per filter process so this is fine. The test at line 789 of `test_tracing.py` already acknowledges this.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,142 @@
...
+
+    if exporter_type in ("otlp", "otlp_grpc"):
+        from opentelemetry.exporter.otlp.proto.grpc.trace_exporter import (
+            OTLPSpanExporter as OTLPGrpcSpanExporter,
+        )
+
+        endpoint = (
+            config.get("endpoint")
+            or os.getenv("TELEMETRY_EXPORTER_OTLP_ENDPOINT")
+            or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
```
lucasmundim, line 82: nit: `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` is not part of the OTel spec — the standard env var is `OTEL_EXPORTER_OTLP_ENDPOINT` (protocol-agnostic). If this is a Plainsight internal convention, a comment would help. Otherwise someone setting the standard `OTEL_EXPORTER_OTLP_ENDPOINT` won't get it picked up.

openfilter/filter_runtime/filter.py
```diff
@@ -1001,9 +1024,37 @@ def _inject_timings(self, frames: dict[str, Frame], t_in: float, t_out: float, d
...
+                    # `pipeline_instance.id = "<uuid>"` returns the full
+                    # api → agent → filter waterfall end-to-end.
+                    "pipeline.id": self.pipeline_id or "",
+                    "pipeline_instance.id": self.pipeline_id or "",
+                },
+            )
+            if tracer
+            else contextlib.nullcontext()
+        )
         t_in = time.time()
```
lucasmundim: nit: `t_in` is before the span context manager opens and `t_out` is after it closes, so the timing EMA includes both span creation and span teardown overhead. In practice this is sub-microsecond, but it means the metric measures "process + tracing" rather than pure "process". Moving both timers inside the `with` block would decouple them.

openfilter/observability/tracing.py
lucasmundim, line 65: nit: The metrics factory (`build_exporter` in `client.py`) only accepts `"otlp"` (mapped to gRPC). This tracing factory adds `"otlp_grpc"` and `"otlp_http"` — a better API, but the two factories now accept different type strings from the same `TELEMETRY_EXPORTER_TYPE` env var. If someone sets `TELEMETRY_EXPORTER_TYPE=otlp_http`, traces go to HTTP but metrics silently fall back to console. Not a bug today, but a latent inconsistency worth aligning in a follow-up.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,174 @@
...
+        #   3. OTEL_EXPORTER_OTLP_ENDPOINT (standard OTel env var)
+        #   4. OTEL_EXPORTER_OTLP_GRPC_ENDPOINT (non-standard, kept for
+        #      backward compat — not part of the OTel spec)
+        #   5. localhost:4317 fallback
+        endpoint = (
+            config.get("endpoint")
+            or os.getenv("TELEMETRY_EXPORTER_OTLP_ENDPOINT")
+            or os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+            or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
+            or "http://localhost:4317"
```
lucasmundim, line 83: **Review feedback addressed: standard `OTEL_EXPORTER_OTLP_ENDPOINT` now in the chain**

The precedence comment is a great addition for OSS users. One subtle thing to note: for the `otlp_http` branch, `OTEL_EXPORTER_OTLP_ENDPOINT` is checked before `OTEL_EXPORTER_OTLP_HTTP_ENDPOINT`. Per the OTel spec, the signal-specific env var (`OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`) should take precedence over the generic one (`OTEL_EXPORTER_OTLP_ENDPOINT`). Not blocking — `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` isn't checked at all, and `TELEMETRY_EXPORTER_OTLP_ENDPOINT` (position 2) covers the Plainsight use case. Just noting for completeness.

openfilter/observability/tracing.py
```diff
@@ -0,0 +1,174 @@
...
+    """Create and optionally register a TracerProvider sharing the same Resource as metrics.
+
+    Args:
+        resource: OTel Resource (reused from MeterProvider).
+        exporter_type: Exporter type string (same values as metrics).
+        exporter_config: Extra kwargs forwarded to the span exporter builder.
+        set_global: When True (default), call ``trace.set_tracer_provider()``
+            to register as the process-global provider. Set to False if the
+            host application manages its own TracerProvider and you want to
+            avoid mutating global OTel state.
```
lucasmundim, line 155: **Review feedback addressed: `set_global` parameter**

Clean solution. Default `True` preserves existing behavior, `False` lets host applications opt out of global mutation. The log line now includes `global=%s` which helps debugging when someone wonders why their TracerProvider didn't take effect.

lucasmundim (commented):
**Re-review after latest push** — three targeted changes, all directly addressing prior review feedback. Nice.

**What changed:**

1. **`SilentSpanExporter` promoted to module level** (`tracing.py:28`) — addresses my earlier nit about discoverability. Now importable and testable independently, matching the metrics side's pattern.

2. **`OTEL_EXPORTER_OTLP_ENDPOINT` added to env var precedence chain** (`tracing.py:81,102`) — fixes the non-standard env var issue I flagged. The standard OTel env var is now checked at position 3 (after explicit config and `TELEMETRY_EXPORTER_OTLP_ENDPOINT`), with the non-standard `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` kept at position 4 for backward compat. Also added to the `otlp_http` branch. The precedence comment documenting all 5 levels is excellent — OSS users can now just set the standard `OTEL_EXPORTER_OTLP_ENDPOINT` and it works.

3. **`set_global` parameter on `setup_tracer_provider`** (`tracing.py:144`) — addresses the global-singleton concern. Host applications that manage their own TracerProvider can now pass `set_global=False`. Default is `True` so existing behavior is unchanged.

4. **`process_frames` cleanup** (`filter.py:1027-1030`) — the double `getattr` chain is now split into two readable lines (`otel = getattr(self, 'otel', None)` then `tracer = getattr(otel, 'tracer', None)`). Also moved `t_in`/`t_out` timers inside the span context manager so timing reflects pure `process()` cost.

5. **New test** for `otlp_http` endpoint from `TELEMETRY_EXPORTER_OTLP_ENDPOINT` env var.

All prior review observations are now addressed. Looks good.

lucasmundim (dismissed):
All review feedback from previous rounds has been addressed. Defaults are safe for OSS users (silent exporter, opt-in via env vars), test coverage is thorough, and the code is clean.

LGTM — ship it.

lucasmundim (approved):
Re-approving after rebase onto main. Tracing code is unchanged — the new base includes the GPU refactoring but the PR's additions are identical. LGTM.

PR #80, opened 2026-04-14
test: replace sleep-based timing with polling in ephemeral topology tests

shingonoide:
## Summary

Replaces fragile `sleep()` + `assertRaises(Empty, ...)` patterns in ephemeral join/tee topology tests with a polling-based `assert_empty()` helper.

## Changes

- New `assert_empty(q, settle=0.2)` helper in `tests/helpers.py` - blocks for `settle` seconds and fails if unexpected data appears
- Refactored `test_topo_ephemeral_join_step` - removed 8 sleep calls, uses polling
- Refactored `test_topo_doubly_ephemeral_tee_step` - removed 4 sleep calls, uses polling
- Data-expected assertions rely on `getdatas()` built-in 5s blocking timeout

## Why

The `sleep(0.05)` timing was too tight for GitHub Actions runners on Python 3.13, causing flaky test failures that blocked the v0.1.28 release. Polling is deterministic regardless of runner speed.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -7,7 +7,7 @@ from openfilter.filter_runtime import Frame
 from openfilter.filter_runtime.mq import MQSender
 
 
-def assert_empty(q, settle=0.5):
+def assert_empty(q, settle=0.2):
     """Assert that *q* receives no data within *settle* seconds.
 
     Blocks for *settle* seconds waiting on the queue. Returns
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -7,6 +7,19 @@ from openfilter.filter_runtime import Frame
 from openfilter.filter_runtime.mq import MQSender
 
 
+def assert_empty(q, settle=0.5):
+    """Assert that *q* receives no data within *settle* seconds.
+
+    Blocks for *settle* seconds waiting on the queue. Returns
+    if the queue stays empty, raises AssertionError otherwise.
+    """
+    try:
+        data = q.get(True, settle)
+    except Empty:
+        return
+    raise AssertionError(f'expected queue to be empty, but got: {data}')
+
+
 class ThreadMQSender(Thread):
     """Threaded MQ sender for tests — ZMQ sender blocks until a subscriber
     issues a request, so it must run in its own thread."""
--- a/tests/test_filter.py
+++ b/tests/test_filter.py
@@ -17,6 +17,7 @@ from openfilter.filter_runtime import Filter, FilterConfig, Frame, FilterContext
 from openfilter.filter_runtime.test import RunnerContext, FiltersToQueue, QueueToFilters
 from openfilter.filter_runtime.utils import setLogLevelGlobal
 from openfilter.filter_runtime.filters.util import Util
+from tests.helpers import assert_empty
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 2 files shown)


tests/helpers.py
```diff
@@ -7,6 +7,19 @@
 from openfilter.filter_runtime.mq import MQSender
 
 
+def assert_empty(q, settle=0.5):
```
lucasmundim: **Nit: wall-clock cost.** The 0.5s default means each `assert_empty()` call blocks for half a second. This test file calls it 4 times total = 2.0s of mandatory wait (vs ~0.45s from the old `sleep(0.05)` + `getdatas(qin, 0.05)` pattern). That's the right trade-off for flaky CI, but worth being aware of if more call sites are added in the future.

tests/test_filter.py
```diff
@@ -250,58 +251,34 @@ def test_topo_ephemeral_join_step(self):
 
         try:
             qout1.put(d := {'main': {'val': 0}})
```
lucasmundim, line 253: The old code had `sleep(0.05)` between `qout.put()` and `runner.step()` for all cases — including the ones where data *is* expected. Those sleeps were removed here too (not just the `assert_empty` cases). Is that intentional? If ZMQ delivery can lag behind the put, `runner.step()` might execute before the data arrives. Likely fine if `step()` processes synchronously from the queue, but wanted to confirm this was considered.

tests/helpers.py
```diff
@@ -7,6 +7,19 @@
...
+    """Assert that *q* receives no data within *settle* seconds.
+
+    Blocks for *settle* seconds waiting on the queue. Returns
+    if the queue stays empty, raises AssertionError otherwise.
+    """
+    try:
+        data = q.get(True, settle)
+    except Empty:
+        return
+    raise AssertionError(f'expected queue to be empty, but got: {data}')
```
lucasmundim, line 20: **Minor:** This raises a bare `AssertionError` rather than going through `unittest`'s `self.fail()` / `self.assert*` machinery, so failure output won't include the extra test context that `unittest` normally adds. Not a blocker — the custom message is informative enough — just noting the trade-off of a free function vs. a method on `TestCase`.

lucasmundim (commented):
Good change overall — replacing fragile `sleep()` calls with a polling-based `assert_empty()` helper is strictly better for CI reliability and readability. A few observations below.

**Note:** The PR title ("increase sleep timing") and body (describing a bump from 0.05 to 0.1) don't match the actual implementation, which replaces sleeps entirely with polling. Worth updating the title/description to reflect the commit message.

lucasmundim (dismissed):
LGTM — solid improvement over sleep-based timing. See comments from previous review for minor observations.

tests/test_filter.py
```diff
@@ -250,58 +251,34 @@ def test_topo_ephemeral_join_step(self):
...
             self.assertIs(runner.step(), False)
             self.assertEqual(getdatas(qin), d)
 
             qout2.put(e := {'other': {'val': 1}})
-
-            sleep(0.05)
-
             self.assertIs(runner.step(), False)
-            self.assertRaises(Empty, lambda: getdatas(qin, 0.05))
+            assert_empty(qin)
```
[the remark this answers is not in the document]
shingonoide: Updated the PR title and description to match the actual implementation. Good catch.

tests/test_filter.py
```diff
@@ -250,58 +251,34 @@ def test_topo_ephemeral_join_step(self):
...
 
             qout1.put(d := {'main': {'val': 1}})
-
-            sleep(0.05)
-
             self.assertIs(runner.step(), False)
             self.assertEqual(getdatas(qin), {**d, **e})
```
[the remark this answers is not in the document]
shingonoide: Agreed - reduced the default from 0.5s to 0.2s. Still 4x the original 0.05s timing, but brings total wall-clock overhead from 2.0s down to 0.8s.

tests/helpers.py
```diff
@@ -7,6 +7,19 @@
 from openfilter.filter_runtime.mq import MQSender
 
 
+def assert_empty(q, settle=0.5):
+    """Assert that *q* receives no data within *settle* seconds.
+
+    Blocks for *settle* seconds waiting on the queue. Returns
+    if the queue stays empty, raises AssertionError otherwise.
```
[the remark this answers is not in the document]
shingonoide: Reduced to 0.2s - see above.

tests/helpers.py
shingonoide: Fair point. Reduced default to 0.2s (total ~0.8s vs ~2.0s). Still conservative enough to avoid flakiness on slow runners.

tests/test_filter.py
shingonoide: Yes, intentional. `qout.put()` writes to a Python Queue (in-process), and `runner.step()` reads from it synchronously. No ZMQ hop between put and step, so no timing gap needed. The old sleep was unnecessary.

tests/helpers.py
shingonoide: Since `assert_empty` is a standalone module-level function (not a test method), it can't use `self.fail()`. Raising `AssertionError` directly is the standard pattern for standalone assertion helpers (same as what `assert` statement does). The error message includes the unexpected data for debugging.

PR #82, opened 2026-04-19
perf(mq): zero-copy IPC transport with SHM + perf waterfall display

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Eliminates redundant image copies on the MQ inter-filter transport hot path through four layered optimizations, then adds an opt-in POSIX shared-memory transport that bypasses the kernel socket buffer entirely. Also adds a terminal-based waterfall display for visualizing the IPC performance gains across git refs.

Four perf optimizations (cumulative on 4K native, 2 filters, raw):

1. Replace bytearray(memoryview(frame.image)) with memoryview.cast('B') in frames2topicmsgs — drops a full pre-send image copy (~17ms → ~10ms/hop)
2. recv_multipart(copy=False) on the ZMQ SUB socket — receiver gets zmq.Frame objects and np.frombuffer builds zero-copy views (~10ms → ~7.5ms/hop)
3. send_multipart(copy=False) on the ZMQ PUB socket — sender hands the buffer to ZMQ without an internal copy (~7.5ms → ~6.5ms/hop)
4. Opt-in SHM transport (OPENFILTER_SHM_ENABLE=1) — sender writes image into a POSIX shm ring slot and sends only a tiny reference over ZMQ; receiver attaches once and builds a zero-copy numpy view (~6.5ms → ~2.9ms/hop)

Plus a terminal waterfall display (make waterfall) that runs the bench harness's pipeline_simulation scenarios at two git refs and renders a rich-styled before/after comparison with per-pipeline gains and per-stage delta annotations. Useful for PR descriptions and demos.

## 🔍 Why is this needed?

The IPC benchmark harness (PR #77 / FILTER-415) showed that 4K-native raw pipelines spent 63% of their per-frame budget on IPC overhead — the cost of moving 24MB image buffers between filters via ZMQ was larger than the filter processing itself. The four optimizations in this PR reduce that overhead by 83% (35ms → 5.7ms per frame at 4K), lifting max throughput from 18 fps to 39 fps on the same hardware without touching any filter code.

The waterfall display was built to show these gains — it runs the same bench scenarios against any two git refs (e.g., HEAD vs origin/main) and renders a terminal panel showing exactly where the time moved. It's anchored to the bench harness's TestPipelineSimulation so it automatically follows any future runtime composition changes.

## 🧪 How was it tested?

- make test passes (--benchmark-disable, existing test suite)
- make bench passes (pipeline_simulation shows reduced IPC overhead)
- OPENFILTER_SHM_ENABLE=1 make test passes (SHM path exercised)
- make waterfall renders correctly against origin/main
- make waterfall.profile produces /tmp/perf-waterfall.svg flamegraph
- py-spy profiling confirmed bottleneck attribution at each stage
- 200-frame bench-mirror reproducer verified gains with realistic access patterns (fresh .ro per iteration + filter sleep)

Measured results (4K native, 2 filters, raw):

    Optimization            IPC OH/frame   Per hop    Max FPS
    Baseline (main)         ~35 ms         ~17.5 ms   18
    + memoryview send       ~20 ms         ~10 ms     25
    + recv copy=False       ~15 ms         ~7.5 ms    29
    + send copy=False       ~13 ms         ~6.5 ms    30
    + SHM transport         ~5.7 ms        ~2.9 ms    39

## 🔗 Related Issues

- FILTER-419: Investigate shared-memory frame transport (this PR)
- FILTER-415: IPC benchmark harness (PR #77, merged — provides the measurement infrastructure this PR optimizes against)
- FILTER-367: Inference Performance & Scaling (parent epic)
- PLAT-866: OTel hop/sub-hop span instrumentation (follow-up — the same runtime paths instrumented here will get OTel spans)

The make waterfall output (run with --frames 100 --baseline-ref origin/main) shows the per-pipeline gains and a frame waterfall for the 4K case. Best viewed in a terminal with truecolor support; Plainsight brand palette applied to structural chrome, conventional green/red for diagnostic signals.

## ✅ Checklist

- [x] I have added or updated documentation as needed

### DIFF
--- a/openfilter/filter_runtime/mq.py
+++ b/openfilter/filter_runtime/mq.py
@@ -27,6 +27,7 @@ import numpy as np
 
 from .frame import Frame
 from .metrics import Metrics
+from .shm_transport import SHMAttachCache, SHMPool, shm_enabled
 from .utils import JSONType, json_getval, rndstr
 from .zeromq import ZMQ_POLL_TIMEOUT as POLL_TIMEOUT_MS, is_zeromq_addr as is_mq_addr, ZMQMessage, ZMQSender, ZMQReceiver
 
@@ -96,6 +97,9 @@ class MQ:
         self.metrics_ = Metrics() if outs_metrics or metrics_cb else DummyMetrics()
         self.metrics  = {'ts': time(), 'fps': 15.0, 'cpu': 0.0, 'mem': 0.0, 'uptime_count': 0}  # initial guaranteed-to-be-present metrics, for outside querying, not used here
 
+        self.shm_pool  = SHMPool(prefix=f'of-{self.mq_id}') if (shm_enabled() and self.sender is not None) else None
+        self.shm_cache = SHMAttachCache() if (shm_enabled() and self.receiver is not None) else None
+
     def _get_filter_data(self, frames: dict[str, Frame]) -> dict:
         """Extract frame IDs from frames or generate one. Returns data for _filter topic."""
         frame_ids = []
@@ -113,6 +117,13 @@ class MQ:
 
     def destroy(self):
         self.metrics_.destroy()
+
+        if self.shm_pool is not None:
+            self.shm_pool.destroy()
+            self.shm_pool = None
+        if self.shm_cache is not None:
+            self.shm_cache.destroy()
+            self.shm_cache = None
 
         if self.receiver:
             self.receiver.destroy()
@@ -176,7 +187,7 @@ class MQ:
             if self.outs_filter is True:
                 frames = {**frames, '_filter': Frame(self._get_filter_data(frames))}
 
-            return MQ.frames2topicmsgs(frames, self.outs_jpg)
+            return MQ.frames2topicmsgs(frames, self.outs_jpg, pool=self.shm_pool)
 
         metrics = None
 
@@ -207,12 +218,12 @@ class MQ:
         topicmsgs, self.send_state = res
         self.recv_state            = None  # we already used up this recv_state so set to None to increment automatically next time in case send() is not called to get new state
 
-        self.metrics_.incoming(frames := MQ.topicmsgs2frames(topicmsgs))
+        self.metrics_.incoming(frames := MQ.topicmsgs2frames(topicmsgs, cache=self.shm_cache))
 
         return frames
 
     @staticmethod
-    def frames2topicmsgs(frames: dict[str, Frame], outs_jpg: bool | None = None) -> dict[str, ZMQMessage]:
+    def frames2topicmsgs(frames: dict[str, Frame], outs_jpg: bool | None = None, *, pool: SHMPool | None = None) -> dict[str, ZMQMessage]:
         topicmsgs = {}
 
         for topic, frame in frames.items():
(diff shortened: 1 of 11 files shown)


openfilter/filter_runtime/shm_transport.py
```diff
@@ -0,0 +1,72 @@
...
+            except FileNotFoundError:
+                pass
+            self.slots.append(shared_memory.SharedMemory(name=name, create=True, size=slot_size))
+
+    def put(self, src: memoryview) -> tuple[str, int]:
+        """Copy src into the next slot and return (slot_name, nbytes)."""
+        nbytes = len(src)
+        slot = self.slots[self._idx % self.nslots]
+        self._idx += 1
+        slot.buf[:nbytes] = src
```
lucasmundim, line 46: **SHM `put()` still copies — but that's the right tradeoff**

`slot.buf[:nbytes] = src` is a memcpy into the SHM segment. This is intentional — the sender needs to release the numpy buffer for the next frame while the receiver reads from SHM. The win is eliminating the *second* copy (ZMQ internal serialization) and the kernel socket buffer transit.

**Overflow concern:** If `nbytes > slot_size`, this will raise silently or corrupt. The default `SHM_SLOT_SIZE` is 32MB which covers 4K raw (2160*3840*3 = ~24.9MB), but a filter outputting higher-res frames or RGBA (4 channels) could exceed it. Consider either a bounds check with a clear error, or documenting the limit in the env var description.

openfilter/filter_runtime/shm_transport.py
```diff
@@ -0,0 +1,72 @@
+"""Shared-memory frame transport for zero-copy inter-filter image handoff."""
+import os
+from multiprocessing import shared_memory
+
+SHM_NSLOTS = max(2, int(os.environ.get('OPENFILTER_SHM_NSLOTS', '4')))
+SHM_SLOT_SIZE = int(os.environ.get('OPENFILTER_SHM_SLOT_SIZE', str(32 * 1024 * 1024)))  # 32MB default
```
lucasmundim, line 6: **Ring buffer slot count vs pipeline depth**

`SHM_NSLOTS` defaults to 4 with `min(2, ...)`. The round-robin `_idx % nslots` means the sender can overwrite a slot the receiver hasn't consumed yet if the receiver is slower than `nslots` frames behind. With 4 slots and a 2-filter pipeline this is fine, but a deeper pipeline (5+ hops) or a slow receiver could see corruption.

The ZMQ send/recv is synchronous in the current pipeline loop (send -> recv -> process -> send), so in practice the sender blocks until the receiver has consumed, making overwrite impossible. But if the runtime ever moves to async pipelining, this would need a generation counter or semaphore. Worth a comment noting the assumption.

openfilter/filter_runtime/mq.py
```diff
@@ -224,32 +235,59 @@ def frames2topicmsgs(frames: dict[str, Frame], outs_jpg: bool | None = None) ->
...
+            if shm_info is not None:
+                if cache is None:
+                    raise RuntimeError('received SHM frame reference but SHM is not enabled on this receiver')
+                shm_name, shm_nbytes = shm_info
+                shm     = cache.get(shm_name)
+                shape   = xtra[:2] if xtra[2] == 'GRAY' else (xtra[0], xtra[1], 3)
+                image   = np.frombuffer(shm.buf, np.uint8, count=shm_nbytes).reshape(shape)
+                dataidx = 1
+            else:
+                dataidx = 2 if xtra else 1
```
lucasmundim, line 269: **Good: `writeable = False` on received frames**

Setting `image.flags.writeable = False` on both the SHM path (line 269) and the zero-copy ZMQ path (line 275) is critical — without this, a downstream filter calling `frame.image[...] = x` would corrupt the shared buffer.

**OSS consideration:** This is a behavioral change for existing filters. Any filter that mutates `frame.image` in-place (e.g., `cv2.rectangle(frame.image, ...)` without `.copy()`) will now get a `ValueError: assignment destination is read-only`. The non-SHM raw path also sets `writeable = False` now, so this affects everyone on this release, not just SHM users.

This is arguably the correct semantic (frames should be treated as immutable between filters), but it's a breaking change that should be documented in RELEASE.md. Filters that need to mutate should call `frame.image = frame.image.copy()` first.

openfilter/filter_runtime/mq.py
```diff
@@ -224,32 +235,59 @@ def frames2topicmsgs(frames: dict[str, Frame], outs_jpg: bool | None = None) ->
...
                 xtra = {'img': [frame.height, frame.width, frame.format, enc]}
-                img  = frame.jpg if do_jpg else bytearray(memoryview(frame.image))
-                msg  = [xtra, img] if data is None else [xtra, img, data]
+
+                if pool is not None and not do_jpg:
+                    shm_name, shm_nbytes = pool.put(memoryview(frame.image).cast('B'))
+                    xtra['shm'] = [shm_name, shm_nbytes]
+                    msg = [xtra] if data is None else [xtra, data]
+                else:
+                    img = frame.jpg if do_jpg else memoryview(frame.image).cast('B')
```
lucasmundim, line 244: **`memoryview(frame.image).cast('B')` — the key optimization**

This replaces the old `bytearray(memoryview(frame.image))` which allocated a new bytearray copy. `cast('B')` just reinterprets the view as unsigned bytes without copying. Clean and correct.

Used in both the SHM path (line 240, fed into `pool.put()`) and the non-SHM path (line 244, passed directly to ZMQ). Combined with `send_multipart(copy=False)`, this eliminates two copies on the send side.

openfilter/filter_runtime/mq.py
```diff
@@ -224,32 +235,59 @@ def frames2topicmsgs(frames: dict[str, Frame], outs_jpg: bool | None = None) ->
...
         for topic, msg in topicmsgs.items():
-            xtra    = xtra['img'] if (xtra := msg[0]) else None
-            dataidx = 2 if xtra else 1
+            xtra_dict = msg[0]
+            xtra      = xtra_dict['img'] if xtra_dict else None
+            shm_info  = xtra_dict.get('shm') if xtra_dict else None
+
+            if shm_info is not None:
+                if cache is None:
+                    raise RuntimeError('received SHM frame reference but SHM is not enabled on this receiver')
```
lucasmundim, line 262: **Good error message for SHM mismatch**

`'received SHM frame reference but SHM is not enabled on this receiver'` — clear diagnostic for the case where sender has `OPENFILTER_SHM_ENABLE=1` but receiver doesn't. This would happen if one filter in the pipeline has the env var and another doesn't. The error is immediate and obvious, which is better than a silent misparse.

lucasmundim (commented):
Strong performance PR with impressive measured gains (83% IPC overhead reduction at 4K). The layered approach is clean — each optimization is independently valuable and they compose well. CI green across Python 3.10-3.13.

**What looks good:**
- Four independent optimizations that compose cleanly (memoryview send, recv copy=False, send copy=False, SHM)
- SHM is fully opt-in (`OPENFILTER_SHM_ENABLE=1`) — zero behavioral change for existing deployments
- `image.flags.writeable = False` on received frames — prevents downstream mutation of SHM/zero-copy buffers
- Stale SHM segment cleanup on startup (sweeps prior crashed runs)
- `SHMAttachCache` lazily attaches — no upfront allocation on the receiver side
- Pipeline scenarios lifted to module scope for tooling reuse — DRY between bench and measurement scripts
- Per-hop IPC timing added to the bench harness (`hop_times_ms`) — good observability
- Waterfall display uses PEP 723 inline deps — no pyproject changes needed
- Test fixture cleanup in `test_filter.py` (`_with_models_toml` context manager) is a nice improvement
- `test_zeromq.py` properly handles `zmq.Frame` objects from `copy=False`

Five observations inline.

openfilter/filter_runtime/shm_transport.py
stwilt: Addressed in 41fec8d: `SHMPool.put()` now raises `ValueError` with a clear message (naming the overflow size and pointing at `OPENFILTER_SHM_SLOT_SIZE`) when `nbytes > slot_size`. Agreed on the 4K-RGBA/higher-res scenarios — the explicit error beats silent corruption.

openfilter/filter_runtime/shm_transport.py
stwilt: Good catch on the synchrony invariant. Addressed in 41fec8d with a comment on `SHMPool` spelling out that round-robin reuse is safe only because the current send→recv pipeline loop is synchronous, and that async dispatch would need a generation counter or semaphore to gate slot reuse.

openfilter/filter_runtime/mq.py
stwilt: Agreed on the breaking-change concern. Documented in 177084c under `## v0.1.28` → `### Breaking Changes` in RELEASE.md: covers the semantic shift, the motivation (SHM slot corruption, `Frame.copy`/`Frame.jpg` optimizations), the migration path (`image = frame.image.copy()` or construct a new `Frame(...)`), and the exact exception text callers will see.

openfilter/filter_runtime/mq.py
stwilt: Thanks — yeah, `cast('B')` + `send_multipart(copy=False)` is where a lot of the send-side win lives.

openfilter/filter_runtime/mq.py
stwilt: Thanks. The cross-mode mismatch (one filter with SHM env var, another without) was the case I most wanted to fail loudly on.

lucasmundim (approved):
All five observations addressed cleanly: bounds check on SHM slot overflow, synchrony invariant documented, breaking change for read-only frames in RELEASE.md. LGTM.

PR #83, opened 2026-04-20
test(batch): add regression tests for frame accumulation edge paths

shingonoide:
## What does this PR do?

Adds three regression tests to `tests/test_frame_accumulation.py` covering frame accumulation behaviors that PR #61 introduced but did not directly exercise. Tests-only; no production code changes.

1. `TestTimerReset.test_timer_reset_extends_accumulation_window`
   Guards the watcher's mid-sleep restart in `_batch_watcher_loop` (the `if self._batch_wakeup.is_set(): self._batch_wakeup.clear(); continue` block in `openfilter/filter_runtime/filter.py`). A frame arriving while the watcher is already sleeping must extend the accumulation window rather than trigger a premature flush at the original deadline.

2. `TestFiniShutdownFlush.test_fini_drains_pending_results`
   Guards the `_batch_pending_results` drain in `fini()` (the `if self._batch_pending_results: ... self.mq.send(pending, POLL_TIMEOUT_MS)` block). Processed results stashed by `_process_frames_batched` but not yet consumed by `loop_once` must be flushed on shutdown instead of being dropped.

3. `TestLoopOncePendingResultsDrain.test_loop_once_sends_all_batch_results`
   Guards the `_batch_pending_results` drain in `loop_once` (the `if self._batch_pending_results: for pending in ...: self._send_frames(...)` block that runs before the final-result send). When `_execute_batch` returns `N > 1` results for a full batch, all `N` must reach `mq.send` in input order; the current path sends `results[:-1]` via the drain and `results[-1]` via the normal return path.

## Why is this needed?

PR #61 (frame accumulation) added three behaviors without direct regression coverage: the watcher's timer-reset path, the shutdown drain in `fini()`, and the multi-result send path in `loop_once`. Each is on a silent-failure trajectory if broken (premature flushes, dropped results on shutdown, or dropped intermediate results from a full batch). Adding focused unit tests makes future regressions surface in CI instead of production.

## How was it tested?

- `pytest tests/test_frame_accumulation.py` passes with all 28 tests green.
- Each new test targets a specific block of production code so a regression in that block surfaces as a clear test failure.

## Related Issues

- Follow-up to #61.

## Checklist

- [ ] I have added or updated **documentation** as needed (tests-only change)

### DIFF
--- a/tests/test_frame_accumulation.py
+++ b/tests/test_frame_accumulation.py
@@ -583,6 +583,7 @@ class TestLoopOncePendingResultsDrain(unittest.TestCase):
         )
 
         f._stop_batch_watcher()
+        f._batch_watcher.join(timeout=1.0)
 
 
 if __name__ == "__main__":
--- a/tests/test_frame_accumulation.py
+++ b/tests/test_frame_accumulation.py
@@ -382,5 +382,198 @@ class TestLoopOnceTimeoutFlush(unittest.TestCase):
         f._stop_batch_watcher()
 
 
+class TestTimerReset(unittest.TestCase):
+    """Verify that _reset_batch_timer correctly extends the accumulation window."""
+
+    def _make_filter(self, batch_size, timeout_ms):
+        config = FilterConfig(
+            id="test-timer-reset",
+            batch_size=batch_size,
+            accumulate_timeout_ms=timeout_ms,
+        )
+        stop_evt = threading.Event()
+        f = CountingBatchFilter(config, stop_evt)
+        f.mq = MagicMock()
+        f.mq.metrics = {}
+        f.mq.sender = None
+        f.logger = MagicMock()
+        f.logger.enabled = False
+        f._is_last_filter = False
+        f._metadata_queue = MagicMock()
+        f._metadata_queue.put_nowait = MagicMock()
+        f.emitter = None
+        f.setup(config)
+        return f
+
+    def test_timer_reset_extends_accumulation_window(self):
+        """A new frame arriving mid-sleep must restart the watcher timeout.
+
+        Regression guard for _batch_watcher_loop lines:
+            if self._batch_wakeup.is_set():
+                self._batch_wakeup.clear()
+                continue
+        Removing those lines causes a premature flush at the original deadline.
+        """
+        # batch_size=5 so the batch never completes via accumulation in this test
+        f = self._make_filter(batch_size=5, timeout_ms=100.0)
+
+        # Frame 1 at t≈0ms: starts the 100ms watcher timer
+        f.process_frames({"main": Frame({"val": 1})})
+
+        # Sleep ~60ms, well inside the original 100ms window
+        time.sleep(0.06)
+
+        # Frame 2 at t≈60ms: _reset_batch_timer() sets _batch_wakeup, watcher restarts
+        # The new flush deadline is ~60ms + 100ms = ~160ms from start
+        f.process_frames({"main": Frame({"val": 2})})
+
+        # At t≈100ms (original deadline): flush must NOT have fired yet.
+        # The watcher's first sleep expires here; seeing _batch_wakeup set it restarts
+        # for another full timeout rather than flushing immediately.
+        time.sleep(0.04)
+        self.assertFalse(
+            f._batch_flush_event.is_set(),
+            "flush fired at the original deadline; timer reset is not working",
+        )
+
+        # Watcher restarts at t≈100ms; expected flush at t≈200ms.  Wait up to
+        # 500ms so the test tolerates scheduler jitter on loaded CI runners.
+        fired = f._batch_flush_event.wait(timeout=0.5)
+        self.assertTrue(fired, "flush never fired after the reset deadline")
+
+        f._stop_batch_watcher()
+        f._batch_watcher.join(timeout=1.0)
+
+
+class TestFiniShutdownFlush(unittest.TestCase):
+    """Verify fini() drains _batch_pending_results before shutting down."""
+
+    def _make_filter(self, batch_size=2):
+        config = FilterConfig(
+            id="test-fini-flush",
+            batch_size=batch_size,
+            accumulate_timeout_ms=1000.0,
+        )
+        stop_evt = threading.Event()
+        f = CountingBatchFilter(config, stop_evt)
+        f.mq = MagicMock()
+        f.mq.metrics = {}
+        f.mq.sender = None
+        f.mq.send = MagicMock(return_value=True)
+        f.mq.destroy = MagicMock()
+        f.logger = MagicMock()
+        f.logger.enabled = False
+        f._is_last_filter = False
+        f._metadata_queue = MagicMock()
+        f._metadata_queue.put_nowait = MagicMock()
+        f.emitter = None
+        f.setup(config)
+        return f
+
+    def test_fini_drains_pending_results(self):
+        """fini() must send every result stashed in _batch_pending_results.
+
+        Regression guard for fini() lines:
+            if self._batch_pending_results:
+                for pending in self._batch_pending_results:
+                    self.mq.send(pending, POLL_TIMEOUT_MS)
+                self._batch_pending_results.clear()
+        Removing that block causes processed results to be silently dropped on shutdown.
+        """
+        f = self._make_filter(batch_size=2)
+
+        # Simulate state: two processed results are waiting to be sent (as would
+        # happen if loop_once was interrupted between _execute_batch and draining).
+        pending1 = {"main": Frame({"val": 10})}
+        pending2 = {"main": Frame({"val": 20})}
+        f._batch_pending_results.extend([pending1, pending2])
+
+        f.fini()
+
+        # Both pending results must have been handed to mq.send.
+        # _frame_buffer is empty, so no additional sends from the partial-batch path.
+        self.assertEqual(
+            f.mq.send.call_count,
+            2,
+            f"expected 2 send calls for pending results; got {f.mq.send.call_count}",
+        )
+        sent_frames = [call.args[0] for call in f.mq.send.call_args_list]
+        self.assertIn(pending1, sent_frames)
+        self.assertIn(pending2, sent_frames)
+
+        # List must be cleared after fini so results are not double-sent.
+        self.assertEqual(len(f._batch_pending_results), 0)
+
+
+class TestLoopOncePendingResultsDrain(unittest.TestCase):
+    """Verify loop_once sends all N results from a full batch, not just the last."""
+
+    def _make_filter(self, batch_size):
+        config = FilterConfig(
+            id="test-pending-drain",
+            batch_size=batch_size,
+            accumulate_timeout_ms=1000.0,
+        )
+        stop_evt = threading.Event()
+        f = CountingBatchFilter(config, stop_evt)
+        f.mq = MagicMock()
+        f.mq.metrics = {}
+        f.mq.sender = None
+        f.mq.send = MagicMock(return_value=True)
+        f.logger = MagicMock()
+        f.logger.enabled = False
+        f._is_last_filter = False
+        f._metadata_queue = MagicMock()
+        f._metadata_queue.put_nowait = MagicMock()
+        f.emitter = None
+        f.sources_timeout = 5000.0
+        f.outputs_timeout = 5000.0
+        f.exit_after_t = None
+        f.setup(config)
+        return f
+
+    def test_loop_once_sends_all_batch_results(self):
+        """loop_once must drain _batch_pending_results so every batch result reaches mq.
+
+        When _execute_batch returns N results for a full batch of size N,
+        _process_frames_batched stashes results[:-1] in _batch_pending_results and
+        returns results[-1]. loop_once must send the stashed results before sending
+        the final one.
+
+        Regression guard for loop_once lines:
+            if self._batch_pending_results:
+                for pending in self._batch_pending_results:
+                    self._send_frames(pending, outputs_timeout)
+                self._batch_pending_results.clear()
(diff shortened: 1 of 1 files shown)


tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,198 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        f.mq.metrics = {}
+        f.mq.sender = None
+        f.logger = MagicMock()
+        f.logger.enabled = False
+        f._is_last_filter = False
+        f._metadata_queue = MagicMock()
+        f._metadata_queue.put_nowait = MagicMock()
+        f.emitter = None
+        f.setup(config)
+        return f
```
stwilt, line 406: The PR adds three more copies of the mocked-`mq`/`logger`/`_metadata_queue`/`emitter` fixture (here, plus `TestFiniShutdownFlush._make_filter` at [`tests/test_frame_accumulation.py` L451](https://github.com/PlainsightAI/openfilter/blob/d04de2bf6ce8b86dc01814255eb4d756eee365c8/tests/test_frame_accumulation.py#L451) and `TestLoopOncePendingResultsDrain._make_filter` at [`tests/test_frame_accumulation.py` L511](https://github.com/PlainsightAI/openfilter/blob/d04de2bf6ce8b86dc01814255eb4d756eee365c8/tests/test_frame_accumulation.py#L511)), on top of the two copies already in the file. Worth extracting a module-level helper now before the fifth divergence — if the `Filter` constructor ever gains a new dependency, all five spots need to stay in sync.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,198 @@ def test_partial_batch_flushed_by_timeout(self):
...
+            "flush fired at the original deadline; timer reset is not working",
+        )
+
+        # Watcher restarts at t≈100ms; expected flush at t≈200ms.  Wait up to
+        # 500ms so the test tolerates scheduler jitter on loaded CI runners.
+        fired = f._batch_flush_event.wait(timeout=0.5)
+        self.assertTrue(fired, "flush never fired after the reset deadline")
+
+        f._stop_batch_watcher()
+        f._batch_watcher.join(timeout=1.0)
```
stwilt, line 447: Defense-in-depth suggestion (not a correctness issue — this ran fine locally under CPU stress and CI is green across 3.10–3.13): four `time.sleep` calls plus a 40ms guard window at line 433 is sensitive to future scheduler jitter on loaded runners. Commit fdbe7e6 set a precedent in this repo of replacing fixed sleeps with polling for similar timing tests. If `_batch_watcher_loop` ever exposed an injectable clock or a test-only hook, this would stop being a candidate for future flakes. Flagging for visibility, not as a blocker.

stwilt (dismissed):
Regression coverage looks solid — the three new classes target the exact `_batch_watcher_loop` wakeup branch, `fini()` drain block, and `loop_once()` pending-results drain that would silently drop frames if they regressed. All 28 pass locally (5 consecutive runs, plus runs under CPU stress with 8 busy cores) and CI is green on 3.10–3.13. Two small suggestions below; nothing blocking.

tests/test_frame_accumulation.py
shingonoide: Good point, the five copies are real duplication and any new `Filter` dependency will force updating all of them. Keeping this PR scoped to regression tests so the fixture lands first; tracked as [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896) along with the polling refactor below. Not applying the extraction here.

tests/test_frame_accumulation.py
shingonoide: Agreed the sleep-plus-guard pattern is a future flake risk and fdbe7e6 is the right precedent. Replacing this with polling requires exposing a test hook or injectable clock in `_batch_watcher_loop` (currently driven by `_batch_stop.wait(timeout)` with no external pulse), which is a production code change beyond the scope of this tests-only PR. Tracked as [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896); leaving the current test as-is since CI has been stable across 3.10-3.13.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,198 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        f = self._make_filter(batch_size=5, timeout_ms=100.0)
+
+        # Frame 1 at t≈0ms: starts the 100ms watcher timer
+        f.process_frames({"main": Frame({"val": 1})})
+
+        # Sleep ~60ms, well inside the original 100ms window
+        time.sleep(0.06)
+
+        # Frame 2 at t≈60ms: _reset_batch_timer() sets _batch_wakeup, watcher restarts
+        # The new flush deadline is ~60ms + 100ms = ~160ms from start
```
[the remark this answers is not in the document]
shingonoide: Good catch, the comment at line 427 contradicted the rest of the docstring. Traced the watcher loop: the inner sleep is indivisible for the full `timeout` before `_batch_wakeup` is observed, so Frame 2 at t≈60ms only takes effect at t≈100ms, and the restarted sleep pushes the flush to t≈200ms. Fixed in ed8d0fc.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,198 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        # Third loop_once fills the batch.  All three results must be sent.
+        f.mq.recv = MagicMock(return_value={"main": Frame({"val": 3})})
+        f.loop_once()
+
+        self.assertEqual(
+            f.mq.send.call_count,
+            3,
+            f"expected 3 sends (r1 + r2 via _batch_pending_results drain, r3 direct); "
+            f"got {f.mq.send.call_count}",
+        )
```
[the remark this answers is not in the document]
shingonoide: Agreed, pinning input order makes the regression guard stronger. One note on the suggested accessor: `_send_frames` calls `mq.send(frames, ...)` where `frames` is a dict of `{topic: Frame}` (from `_normalize_frame_result`), so `call.args[0]` is the dict, not the Frame. Applied a corrected form `call.args[0]["main"].data["val"]` in ed8d0fc.

stwilt (dismissed):

lucasmundim (commented):
Good PR — well-scoped tests-only change targeting real coverage gaps from PR #61. The test docstrings are excellent (citing the exact production code lines being guarded). Three observations below.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,208 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        f.mq.metrics = {}
+        f.mq.sender = None
+        f.logger = MagicMock()
+        f.logger.enabled = False
+        f._is_last_filter = False
+        f._metadata_queue = MagicMock()
+        f._metadata_queue.put_nowait = MagicMock()
+        f.emitter = None
+        f.setup(config)
+        return f
```
lucasmundim, line 406: Nit: this `_make_filter` is nearly identical to the ones in `TestBatchAccumulation` (line 96), `TestFiniShutdownFlush` (line 453), and `TestLoopOncePendingResultsDrain` (line 513) — four copies total, differing only in a few optional attributes (`mq.send`, `sources_timeout`, etc.).

Consider promoting the existing `TestBatchAccumulation._make_filter` to a module-level helper that accepts optional overrides, then calling it from all four classes. That would cut ~60 lines of duplication and make it easier to add new test classes later without copy-pasting setup.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,208 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        f.process_frames({"main": Frame({"val": 2})})
+
+        # At t≈100ms (original deadline): flush must NOT have fired yet.
+        # The watcher's first sleep expires here; seeing _batch_wakeup set it restarts
+        # for another full timeout rather than flushing immediately.
+        time.sleep(0.04)
+        self.assertFalse(
+            f._batch_flush_event.is_set(),
+            "flush fired at the original deadline; timer reset is not working",
+        )
```
lucasmundim, line 439: The negative assertion here (`assertFalse` at t≈100ms against a 100ms timer) is inherently racy. On a loaded CI runner, if the test's sleeps run slightly long or the watcher's first sleep expires slightly early, the watcher could restart its second timeout window before this line executes — making the assertion vacuously true rather than genuinely testing the reset path. Conversely, extreme scheduler delays could cause a false failure.

The 40ms margin is generous enough that this should rarely flake in practice, but it's worth noting — especially since PR #80 (commit fdbe7e6) recently moved other tests in this repo away from sleep-based timing toward polling. That said, this particular test is fundamentally harder to express without sleeps since it's asserting *when* the flush fires, not just *that* it fires.

tests/test_frame_accumulation.py
```diff
@@ -382,5 +382,208 @@ def test_partial_batch_flushed_by_timeout(self):
...
+        sent_vals = [
+            call.args[0]["main"].data["val"] for call in f.mq.send.call_args_list
+        ]
+        self.assertEqual(
+            sent_vals,
+            [1, 2, 3],
+            "batch results should reach mq.send in input order (r1, r2, r3)",
+        )
+
+        f._stop_batch_watcher()
```
lucasmundim, line 585: Minor inconsistency: `TestTimerReset` does `_stop_batch_watcher()` + `join(timeout=1.0)` (line 446-447), but here there's only `_stop_batch_watcher()` without a join. The watcher is a daemon thread so it won't block test exit, but the inconsistency could mask thread-leak issues. Consider adding `f._batch_watcher.join(timeout=1.0)` here too for consistency.

tests/test_frame_accumulation.py
shingonoide: Thanks, this is the same duplication stwilt flagged and it's already tracked in [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896) as a follow-up to promote `_make_filter` to a module-level helper with optional overrides. Keeping this PR scoped to coverage; the DRY pass will land in a separate tests-only change.

tests/test_frame_accumulation.py
shingonoide: Agreed on the brittleness, this `assertFalse` at t≈100ms is indeed the tightest margin in the suite. [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896) already tracks a polling-migration pass for these tests; I just added this spot in the ticket as the highest-risk hotspot. As you note, a 1:1 rewrite is non-trivial here (the assertion is about *when* flush fires, not just *that* it fires), so scope carefully.

tests/test_frame_accumulation.py
shingonoide: Good catch, fixed in 93c1482. One-line addition after `_stop_batch_watcher()` to match the teardown pattern in `TestTimerReset`. Note: `TestLoopOnceTimeoutFlush` line 382 (pre-existing) has the same inconsistency; leaving that for the DRY pass in [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896) where a shared helper can centralize teardown.

lucasmundim (approved):
LGTM. The watcher join inconsistency is fixed. The remaining comments (duplicated `_make_filter`, timing sensitivity) are minor and don't block merging.

shingonoide:
## Summary of review responses

Round 1 (push ed8d0fc, response to @stwilt and @copilot):
- Fixed Copilot's timing-comment contradiction at line 427 (docstring now matches `_batch_watcher_loop` actual behavior: reset extends deadline to ~200ms, not ~160ms).
- Added Copilot's suggested send-order assertion at line 573, with accessor corrected to match how `_normalize_frame_result` wraps bare Frames in `{"main": frame}`.
- Deferred @stwilt's two suggestions (fixture dedup + sleep to polling) into [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896).

Round 2 (push 93c1482, response to @lucasmundim):
- Applied L3: `f._batch_watcher.join(timeout=1.0)` after `_stop_batch_watcher()` in `TestLoopOncePendingResultsDrain` to match `TestTimerReset`'s teardown pattern.
- Bundled L1 (DRY across `_make_filter` copies) and L2 (race at `assertFalse` line 439) into [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896). Flagged line 439 as the highest-risk polling-migration target.

### Review record

- @stwilt approved on ed8d0fc at 00:31 BRT (03:31Z) on 2026-04-20.
- @lucasmundim approved on 93c1482 at 01:25 BRT (04:25Z) on 2026-04-20 with "LGTM".

@lucasmundim, the two items you flagged as minor (DRY across `_make_filter` and timing sensitivity at line 439) are tracked in [PLAT-896](https://plainsight-ai.atlassian.net/browse/PLAT-896) as follow-up, not blockers here.

Thanks @stwilt @lucasmundim for the reviews, and @copilot for the suggestions. CI is green on 93c1482; no further changes planned.


[PLAT-896]: https://plainsight-ai.atlassian.net/browse/PLAT-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLAT-896]: https://plainsight-ai.atlassian.net/browse/PLAT-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[PLAT-896]: https://plainsight-ai.atlassian.net/browse/PLAT-896?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #84, opened 2026-04-21
chore: use shared security scan workflow, bump actions and fix CVEs

lucasmundim:
## Summary

- Replace the standalone `security-scan.yaml` with a call to the shared `PlainsightAI/gh-actions-public` security scan workflow, which handles Grype configuration centrally — including CVE ignores for transitive dependencies like the ffmpeg bundled inside `opencv-python-headless`
- Bump all GitHub Actions to latest major versions to avoid Node.js <24 deprecation warnings
- Bump `requests` and `wheel` to fix reported CVEs

### Action version bumps

| Action | Before | After |
|--------|--------|-------|
| `actions/checkout` | v4 | v6 |
| `actions/setup-python` | v5 | v6 |
| `actions/upload-artifact` | v4 | v7 |
| `actions/download-artifact` | v4.1.3 | v8 |
| `docker/setup-buildx-action` | v3 | v4 |
| `docker/login-action` | v3 | v4 |
| `docker/build-push-action` | v5 | v6 |
| `dorny/paths-filter` | v3 | v4 |
| `mukunku/tag-exists-action` | v1.6.0 | v1.7.0 |

### Dependency CVE fixes

| Package | Before | After | CVE |
|---------|--------|-------|-----|
| `requests` | ~=2.32.5 | ~=2.33.0 | CVE-2026-25645 (Medium) |
| `wheel` | ~=0.45.1 | ~=0.46.2 | GHSA-8rrh-rw8j-w5fx (High) |
| `actions/download-artifact` | v4.1.3 | v8 | GHSA-cxww-7g56-2vh6 (High) |
| `opencv-python-headless` ffmpeg | 8.0.1 | n/a (no fix yet) | CVE-2026-40962 (Critical) — ignored via shared workflow |

## Test plan

- [ ] CI workflow passes (action version bumps)
- [ ] Security scan passes (shared workflow with CVE ignore)
- [ ] Create Release workflow still works on next release


.github/workflows/security-scan.yaml
```diff
@@ -9,40 +9,7 @@ on:
...
-      - name: Download Grype
-        uses: anchore/scan-action/download-grype@v3
-        id: grype
-        with:
-          grype-version: v0.103.0
-
-      - name: Run Grype security scan
-        run: ${{steps.grype.outputs.cmd}} --name openfilter -o table --fail-on critical --only-fixed dir:.
+  security-scan:
+    uses: PlainsightAI/gh-actions-public/.github/workflows/security-scan.yaml@main
```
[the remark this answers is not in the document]
lucasmundim: Intentional — `gh-actions-public` is our own org repo, and `@main` is how all ~24 filter repos pick up shared workflow updates (like the `ignore_cves` default we just added) without per-repo PRs. The private `gh-actions` repo follows the same pattern. Pinning to a SHA would require updating every consumer on every change, which is the opposite of what the shared workflow is for.

.github/workflows/create-release.yaml
```diff
@@ -315,7 +315,7 @@ jobs:
           echo "tags=${IMAGE}:${VERSION},${IMAGE}:latest" >> $GITHUB_OUTPUT
 
       - name: Build and push Docker image
-        uses: docker/build-push-action@v5
+        uses: docker/build-push-action@v6
```
shingonoide, line 318: Nit (optional): `docker/build-push-action@v6` enables provenance + SBOM attestations by default when `push: true`. Next openfilter release will therefore publish images with SLSA provenance attached to the manifest list. Generally an improvement — just worth a heads-up in `RELEASE.md` so downstream consumers pinning by digest know the image shape changed. If you want to preserve v5 behavior exactly, `provenance: false` and `sbom: false` would do it — but letting the new default ride is the reasonable choice.

.github/workflows/create-release.yaml
```diff
@@ -205,18 +205,18 @@ jobs:
...
+        uses: actions/setup-python@v6
         with:
           python-version: "3.11"
 
       - name: Install dependencies
         run: pip install twine
 
       - name: Download wheel artifact
-        uses: actions/download-artifact@v4.1.3
+        uses: actions/download-artifact@v8
```
shingonoide, line 219: Nit (optional): upload on v7 (line 194) + download on v8 here is safe because both sides live in the same workflow run under the same artifact name — the wire format matches across v5+. Just flagging that the v4.1.3 → v8 jump is what pulls in the `GHSA-cxww-7g56-2vh6` fix. Matching majors on both sides is the usual convention once the CVE-forcing constraint relaxes in a future bump.

shingonoide (approved):
Clean consumer-side migration to the shared `gh-actions-public` security-scan workflow (the #6 we approved), plus a well-scoped batch of action-major bumps and 3 advisory mitigations (`CVE-2026-25645` on requests, `GHSA-8rrh-rw8j-w5fx` on wheel, `GHSA-cxww-7g56-2vh6` on download-artifact). Full matrix is green across Py 3.10–3.13 for both unit tests and the new shared security scan. `@main` on the reusable workflow is the right call — that's how the full filter-repo fleet picks up shared workflow updates — Copilot's SHA-pin suggestion is already rebutted in-thread.

A couple of low-stakes observations below — all optional; none block.

One meta-note: `create-release.yaml` only triggers on `workflow_dispatch`, so the bumps to `upload-artifact@v7`, `download-artifact@v8`, `docker/login@v4`, `docker/build-push@v6`, and `mukunku/tag-exists@v1.7.0` aren't exercised by PR CI. Nothing here is expected to break (all Node-runtime bumps or fix-only minor bumps), but a `workflow_dispatch` dry-run against a throwaway tag before the next real release cut would catch any surprise without production risk. Entirely optional.

.github/workflows/create-release.yaml
lucasmundim: Good call on the provenance/SBOM attestations. Letting the new default ride is the right call — it is an improvement. Will add a note to RELEASE.md when we cut v0.1.30.

.github/workflows/create-release.yaml
lucasmundim: Ack — the v7/v8 mismatch is intentional for now since v8 is what fixes the GHSA. Will align majors on the next bump cycle.

PR #86, opened 2026-05-01
feat(filter_runtime): typed FilterConfigBase + emit_schema (FILTER-441)

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Introduces `FilterConfigBase(BaseSettings)` — the typed, introspectable successor to the legacy `adict`-based `FilterConfig` — and the surrounding machinery that lets a filter declare its config contract in Python and emit it as a JSON Schema (draft 2020-12) the platform can ingest without pulling the filter image.

Specifically:

- New `FilterConfigBase` in `openfilter/filter_runtime/config.py`. Subclasses get pydantic-settings env-var sourcing (`FILTER_*` prefix, `__` nested delimiter), declarative range/enum/pattern/format constraints, and a `model_json_schema()` export.
- New `openfilter/filter_runtime/formats.py` — `OpenfilterSource` and `VideoSource` Annotated aliases plus their Python-side validators (FC-1.5 py).
- FC-2 markers: `x-openfilter-managed`, `x-openfilter-resolve`, and `x-openfilter-preflight` JSON Schema extension keywords, plus `Managed(...)` and `Resolve(...)` field helpers. Managed fields are stripped from the operator-facing schema by default (filter exposes operator surface only; orchestrator handles managed fields).
- New authoring guide at `docs/declarative-config.md` covering the typed-config contract end to end (Tier 1/2/3 migration, env-var sourcing, tagged unions, named formats, the `Managed` / `Resolve` boundary).
- Tier-3 fallback intact: filters that don't subclass `FilterConfigBase` keep working unchanged.
- Adds `pydantic~=2.9` and `pydantic-settings~=2.6` as hard deps; lock file refreshed.

The build-time CLI half (`openfilter emit-schema`) is stacked on top of this PR as
[FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442).

---

## 🔍 Why is this needed?

Filter authors need a way to declare their config contract once, in Python, and have the platform pick it up at deploy time without side-loading filter images. Today the platform infers config from hand-maintained env-var docs that drift; the platform-side ingest work ([Declarative Configuration Validation](https://plainsight-ai.atlassian.net/browse/FILTER-440)) is blocked on having a real schema artifact to ingest.

This PR is the SDK-side scaffolding for that contract. It ships the typed base class, the named-format aliases, and the FC-2 boundary markers that distinguish operator-visible config from platform-managed / resolved / preflight fields. Once it lands, individual filters can opt in at their own cadence (Tier 1/2 per the design doc); unmigrated filters keep working through the openfilter 1.0 cut.

Design doc:
[Declarative Filter Configuration Contract — Goldenrod.2](https://plainsight-ai.atlassian.net/wiki/spaces/ENG/pages/2755919874).

---

### Unit tests (37, all passing)

`tests/test_filter_config_base.py` covers:

- env-var sourcing with the `FILTER_*` prefix
- tagged unions via `env_nested_delimiter="__"` (sam3 prompt-mode fitness test)
- managed-field exclusion in `emit_schema()` and `--include-managed` parity
- `Managed` / `Resolve` field helpers and parent/child override semantics
- `OpenfilterSource` / `VideoSource` validators (accept/reject + JSON Schema `format` keyword)

```
$ uv run --with pytest pytest tests/test_filter_config_base.py -q
.....................................                                    [100%]
37 passed in 0.30s
```

### Integration spike: sam3-shaped config

Validated FILTER-441's load-bearing acceptance criterion — sam3's three mutually-exclusive prompt modes as a tagged union — against a sam3-shaped `FilterConfigBase` subclass (text / multi-text / exemplars prompt modes, plus `Managed` `sources` and `device` fields, plus representative hyperparams). Results:

- ✅ `prompting` emits a discriminated union with `discriminator: {propertyName: "prompt_mode", mapping: {...}}` and `oneOf` referring to `$defs/{TextPrompt,MultiTextPrompt,Exemplars}`
- ✅ Managed fields (`sources`, `device`) stripped from operator-facing schema; `--include-managed` surfaces them with `x-openfilter-managed` and `x-openfilter-resolve` markers
- ✅ `sources.items.format == "video-source"` (FC-1.5 named format round-trips)
- ✅ Pydantic-settings env-var sourcing into the tagged union: `FILTER_PROMPTING__PROMPT_MODE=text FILTER_PROMPTING__TEXT_PROMPT='person walking'` parses into the `TextPrompt` variant; `FILTER_CONFIDENCE_THRESHOLD=0.7` overrides the top-level default
- ✅ Top-level `required: ["prompting"]` correctly surfaces fields with no default

This stacks on [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) for the CLI invocation; the spike was run via
`python -m openfilter.cli emit-schema` from the FILTER-442 worktree.

### Tier-3 fallback

No production filter touched in this PR. Filters that don't subclass `FilterConfigBase` (i.e. every existing one) are unaffected.

---

## 🔗 Related Issues

- [FILTER-441](https://plainsight-ai.atlassian.net/browse/FILTER-441) — this ticket
- [FILTER-440](https://plainsight-ai.atlassian.net/browse/FILTER-440) — Declarative Configuration Contract (parent epic)
- [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) — `openfilter emit-schema` CLI (stacked, follows this PR)

---

Sample emitted schema (managed field stripped by default):

```python
class MyFilter(FilterConfigBase):
    sources: list[VideoSource] = Managed([], resolve="orchestrator-generated")
    confidence_threshold: float = Field(default=0.5, ge=0, le=1)

>>> MyFilter.emit_schema()
{
  "title": "MyFilter",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "confidence_threshold": {
      "type": "number", "minimum": 0, "maximum": 1, "default": 0.5
    }
  },
  ...
}
```

---

## ✅ Checklist

- [x] I have added or updated **documentation** as needed (`docs/declarative-config.md` — authoring guide for the typed-config contract; the parent epic FC-12 row is closed by this addition)


lucasmundim (commented):
Solid PR — the design cleanly separates operator-facing from platform-managed config, the pydantic-settings integration is well done, and the 37 tests cover the core contracts well. The `_NamedFormat` pattern for injecting `format:` into JSON Schema while keeping the runtime validator is elegant, and the authoring guide is thorough.

A few items worth addressing before merge:

docs/declarative-config.md
```diff
@@ -0,0 +1,217 @@
...
+
+class MyFilterConfig(FilterConfigBase):
+    # Operator-facing fields
+    model_id: str = "facebook/sam3"
+    confidence_threshold: float = Field(default=0.5, ge=0, le=1)
+    mode: Literal["fast", "accurate"] = "fast"
+
+    # Platform-managed fields (see "The boundary" below)
+    sources: list[VideoSource] = Managed([], resolve="orchestrator-generated")
+    device: str = Managed("cuda", resolve="device-policy")
```
lucasmundim, line 59: `"device-policy"` is not a member of the `ResolveHint` literal defined in `config.py:43-49`:

```python
ResolveHint = Literal[
    "orchestrator-generated",
    "orchestrator-resolved",
    "tenant-provided",
    "secret-ref",
    "agent-env",
]
```

This won't blow up at runtime (Python doesn't enforce `Literal` on plain function params), but type checkers will flag it, and copy-paste from the docs will produce warnings. Same issue on line 153.

Either add `"device-policy"` to `ResolveHint`, or change the docs examples to use `"agent-env"` (matching the actual base class field).

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,208 @@
...
+for downstream consumers (the platform's autofill walker).
+
+Coexists with the legacy ``FilterConfig`` — unmigrated filters keep working
+unchanged. See ``openfilter.filter_runtime.formats`` for SDK-owned named
+formats.
+"""
+
+from __future__ import annotations
+
+from typing import Annotated, Any, Literal
```
lucasmundim: Nit: `Annotated` is imported but unused in this module — it's only needed in `formats.py`.

openfilter/filter_runtime/formats.py
```diff
@@ -0,0 +1,148 @@
...
+Additional formats land as bundled-filter migrations under FC-10 force them.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Annotated, Any
+from urllib.parse import urlparse
+
+from pydantic import AfterValidator, Field, GetJsonSchemaHandler
```
lucasmundim: Nit: `Field` is imported but unused in this module.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,208 @@
...
+                    continue
+                kept_props[prop_name] = _strip_managed(prop_schema)
+            out[key] = kept_props
+            if dropped and isinstance(schema.get("required"), list):
+                out["required"] = [r for r in schema["required"] if r not in dropped]
+        elif key == "required":
+            if "required" in out:
+                continue
+            out[key] = list(value) if isinstance(value, list) else value
+        elif key == "$defs" and isinstance(value, dict):
```
lucasmundim, line 240: When a managed property is removed from `properties`, any `$defs` entry it referenced via `$ref` is still emitted. The function recurses into `$defs` to strip managed fields *within* definitions (good), but doesn't prune definitions that are no longer referenced by any remaining property.

This produces valid but bloated schemas. Not a correctness bug, but the platform team ingesting these may wonder about unreferenced definitions. Worth a `# TODO` or a follow-up.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,208 @@
...
+        None, resolve="orchestrator-generated"
+    )
+    outputs_balance: bool | None = Managed(None)
+    outputs_timeout: int | None = Managed(None)
+    outputs_required: str | None = Managed(None)
+    outputs_metrics: str | bool | None = Managed(None)
+    outputs_filter: bool | None = Managed(None)
+    outputs_jpg: bool | None = Managed(None)
+
+    exit_after: float | str | None = None
```
lucasmundim, line 153: Is `exit_after` intentionally non-managed? Unlike `batch_size` / `accumulate_timeout_ms` (which arguably make sense for operator tuning), `exit_after` reads more like a debug/lifecycle parameter — the legacy class describes it as `'[[[days:]hrs:]mins:]secs[.subsecs]' or '@date/time/datetime'`.

If operators shouldn't touch it, it should be `Managed`. If it is intentionally operator-facing, ignore this.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,208 @@
...
+    ``sources`` with ``list[OpenfilterSource]`` (from ``formats``) is the
+    canonical worked example.
+    """
+
+    model_config = SettingsConfigDict(
+        env_prefix="FILTER_",
+        env_nested_delimiter="__",
+        extra="allow",
+        case_sensitive=False,
+        validate_default=False,
```
lucasmundim, line 131: With `validate_default=False`, a filter author can write `threshold: float = Field(default=1.5, ge=0, le=1)` and the class will import and instantiate fine — the constraint only fires when the field is set via env var.

This is the right call for backward compat, but it could silently accept invalid defaults. Consider a one-liner in the authoring guide: "defaults are not validated at import time — test your defaults explicitly."

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,208 @@
...
+    outputs_timeout: int | None = Managed(None)
+    outputs_required: str | None = Managed(None)
+    outputs_metrics: str | bool | None = Managed(None)
+    outputs_filter: bool | None = Managed(None)
+    outputs_jpg: bool | None = Managed(None)
+
+    exit_after: float | str | None = None
+
+    environment: str | None = Managed(None, resolve="agent-env")
+    log_path: str | bool | None = Managed(None)
```
lucasmundim: Minor type drift vs legacy `FilterConfig`:

- Legacy `log_path: str | Literal[False] | None` → here `str | bool | None` (widens `False` to `bool`)
- Legacy `extra_metrics: dict[str, JSONType] | ...` → here `dict[str, Any] | ...` (widens `JSONType` to `Any`)

Presumably intentional relaxations — just flagging in case they're accidental.

tests/test_filter_config_base.py
```diff
@@ -0,0 +1,320 @@
...
+        sources: list[OpenfilterSource] = Managed(
+            [], resolve="orchestrator-generated"
+        )
+
+    schema = EventSinkLike.emit_schema(include_managed=True)
+    sources = schema["properties"]["sources"]
+    assert sources[MANAGED_KEY] is True
+    assert sources["items"]["format"] == "openfilter-source"
+
+
```
lucasmundim, line 309: This test covers overriding a parent managed field with a *different* managed field. There's no test for the arguably more important Tier-1 case: a subclass overriding a parent managed field with a **plain (non-managed) field** to surface it to operators.

Worth adding a test that verifies the child's field appears in `emit_schema()` *without* `x-openfilter-managed`.

docs/declarative-config.md
stwilt: Good catch — added `"device-policy"` (and `"template-substitution"`, which I'd similarly slipped into the `Resolve()` example) to the `ResolveHint` literal in `config.py`. Docs examples now match the API. Done in b0175e7.

openfilter/filter_runtime/config.py
stwilt: Removed in b0175e7.

openfilter/filter_runtime/formats.py
stwilt: Removed in b0175e7 (also dropped an unused `Any` while I was there).

openfilter/filter_runtime/config.py
stwilt: Implemented in b0175e7 — `_strip_managed` now calls `_prune_orphan_defs` after the strip pass, walking the remaining schema to collect `$ref` targets and dropping any `$defs` entry not referenced. Iterates to fixed point so a def referenced only by another orphaned def is also dropped. Covered by `test_emit_schema_prunes_orphan_defs_transitively` in `tests/test_filter_config_base.py`.

openfilter/filter_runtime/config.py
stwilt: Keeping it operator-facing for now. Zero references to `exit_after` / `EXIT_AFTER` / `FILTER_EXIT` in `openfilter-pipelines-controller` or `plainsight-api`, and there's an existing operator-facing example at [`docs/image-in-filter.md` L1044](https://github.com/PlainsightAI/openfilter/blob/b0175e75c65b1a212d289a2c7cddda887f864d37/docs/image-in-filter.md#L1044) (`exit_after=300, # Exit after 5 minutes`). The controller's filter-config wiring just round-trips operator-set values to `FILTER_<NAME>` env vars — there's no orchestrator-owned lifecycle code that would benefit from owning this. If platform later wants time-bounded deployments, the env-var passthrough handles override and we can revisit.

openfilter/filter_runtime/config.py
stwilt: Added a "Caveat — defaults aren't validated at import time" callout to the Quick Start section in `docs/declarative-config.md` (b0175e7). Notes the workaround: cover defaults with an explicit unit test if they have constraints.

openfilter/filter_runtime/config.py
stwilt: Both narrowed in b0175e7 to match legacy: `log_path: str | Literal[False] | None`; `extra_metrics: dict[str, JSONType] | list[tuple[str, JSONType]] | None` (`JSONType` imported from `openfilter.filter_runtime.utils`). Drifts were unintentional.

tests/test_filter_config_base.py
stwilt: Added `test_child_can_surface_managed_parent_field_to_operators` in b0175e7. Subclasses `FilterConfigBase` overriding the parent's `device` (Managed) with a plain `Literal["cpu","cuda","mps"]` field; asserts neither `MANAGED_KEY` nor `RESOLVE_KEY` appear on the emitted property.

pyproject.toml
```diff
@@ -39,8 +39,11 @@ dependencies = [
   "opentelemetry-proto~=1.39.1",
   "opentelemetry-resourcedetector-gcp~=1.11.0a0",
   "opentelemetry-sdk~=1.39.1",
-  "opentelemetry-semantic-conventions~=0.60b1"
+  "opentelemetry-semantic-conventions~=0.60b1",
 
+  "pydantic~=2.9",
+  "pydantic-settings~=2.6",
+  "typing-extensions~=4.15",
```
lucasmundim: The PR only uses `typing_extensions.deprecated` (available since 4.5.0) and `Annotated` (much earlier). The `~=4.15` pin means `>=4.15, <5.0` — since openfilter is an SDK consumed as a library, this could cause dependency conflicts for downstream users who have typing-extensions pinned to e.g. 4.12 via another package. Pydantic 2.9 itself only requires `>=4.6`.

Consider relaxing to `~=4.6` or `>=4.6` to match pydantic's own floor and avoid unnecessarily constraining consumers.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,285 @@
...
+            kept_props: dict[str, Any] = {}
+            dropped: set[str] = set()
+            for prop_name, prop_schema in value.items():
+                if isinstance(prop_schema, dict) and prop_schema.get(MANAGED_KEY) is True:
+                    dropped.add(prop_name)
+                    continue
+                kept_props[prop_name] = _strip_managed(prop_schema)
+            out[key] = kept_props
+            if dropped and isinstance(schema.get("required"), list):
+                out["required"] = [r for r in schema["required"] if r not in dropped]
```
lucasmundim: When all properties in a sub-schema are managed and get dropped, the output contains `"properties": {}, "required": []`. While semantically equivalent to an absent `required` key per the spec, some schema consumers may trip on the empty array (e.g. a naive `if "required" in schema` check).

Consider omitting it when it filters down to empty:

```python
if dropped and isinstance(schema.get("required"), list):
    filtered = [r for r in schema["required"] if r not in dropped]
    if filtered:
        out["required"] = filtered
    # else: omit empty required entirely
```

Low priority — just a cleanliness nit.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,285 @@
...
+    def emit_schema(cls, *, include_managed: bool = False) -> dict[str, Any]:
+        """Return the filter's JSON Schema (draft 2020-12).
+
+        By default, fields tagged ``x-openfilter-managed: true`` are stripped
+        — that's the pipeline-operator surface, what the editor renders.
+        Pass ``include_managed=True`` for debugging or for the platform's
+        autofill walker, which needs the managed surface to know what it
+        owns.
+        """
+        schema = cls.model_json_schema(mode="validation")
```
lucasmundim, line 180: The docstring says "JSON Schema (draft 2020-12)" and the PR description sample output shows `"$schema": "https://json-schema.org/draft/2020-12/schema"`, but pydantic's `model_json_schema()` doesn't emit the `$schema` keyword by default.

If the platform ingest code dispatches on `$schema` to select the draft version, this could be a gap. Is this intentionally deferred to FILTER-442 where the CLI wraps the output? If so, worth a brief comment here noting that the CLI layer is responsible for adding `$schema`.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,285 @@
...
+    "PREFLIGHT_KEY",
+]
+
+
+MANAGED_KEY = "x-openfilter-managed"
+RESOLVE_KEY = "x-openfilter-resolve"
+PREFLIGHT_KEY = "x-openfilter-preflight"
+
+
+ResolveHint = Literal[
```
lucasmundim, line 46: `ResolveHint` is used in the signatures of `Managed` and `Resolve` but isn't in `__all__` or re-exported from `__init__.py`. Filter authors using type checkers who want to annotate a variable holding a resolve hint (e.g. a helper that constructs managed fields dynamically) would need to import from the internal module path.

Not blocking — most usage is via string literals — but consider adding it to the public surface for completeness.

openfilter/filter_runtime/config.py
```diff
@@ -0,0 +1,285 @@
...
+
+    ``resolve`` is optional. Pass it when there's a canonical resolution
+    category (orchestrator-side autofill, secret manager, agent-env lookup,
+    etc.) and the platform should know which dispatcher to use. Omit it when
+    the field is purely platform-owned with no further resolution semantics —
+    several inherited ``FilterConfigBase`` managed fields (e.g. ``log_path``,
+    ``metrics_interval``) are platform-owned plumbing without a canonical
+    resolver and intentionally leave ``resolve`` unset.
+    """
+    extra: dict[str, Any] = dict(json_schema_extra or {})
```
lucasmundim, line 79: The merge of caller-supplied `json_schema_extra` with the managed markers is a nice touch, but there's no test locking in the behavior. A filter author could reasonably do:

```python
field: str = Managed(None, json_schema_extra={"x-custom-ui-hint": "slider"})
```

and expect both `x-openfilter-managed: true` and `x-custom-ui-hint: "slider"` in the emitted schema. Worth a quick test to prevent accidental regressions if this function is refactored later.

lucasmundim (commented):
Second-pass review. All first-round items confirmed fixed in b0175e7. The PR is in good shape — none of these are blockers, mostly suggestions and edge-case hardening.

pyproject.toml
stwilt: Relaxed to `>=4.6` in d724eb5 to match pydantic's floor. `uv.lock` regenerated.

openfilter/filter_runtime/config.py
stwilt: `_strip_managed` now omits `required` entirely when filtering produces an empty list — both at the top level and for nested objects. New test `test_emit_schema_omits_required_when_all_managed` locks the cleanliness in. d724eb5.

openfilter/filter_runtime/config.py
stwilt: Added `setdefault("$schema", "https://json-schema.org/draft/2020-12/schema")` in `emit_schema()` after the strip/prune pipeline (d724eb5), so the contract holds without depending on the FILTER-442 CLI layer. The CLI's own `setdefault` becomes a no-op but stays as defensive belt-and-suspenders. New test `test_emit_schema_sets_dollar_schema_keyword`.

openfilter/filter_runtime/config.py
stwilt: Added `ResolveHint` to `config.py`'s `__all__` and re-exported it from `openfilter.filter_runtime.__init__.py` alongside the other FC-2 names. Filter authors can now `from openfilter.filter_runtime import ResolveHint`. d724eb5.

openfilter/filter_runtime/config.py
stwilt: Confirmed the merge already works (`Managed`/`Resolve` both do `dict(json_schema_extra or {})` then add the FC-2 markers), but the behavior wasn't covered. Added `test_managed_preserves_caller_supplied_json_schema_extra` in d724eb5 — asserts both the caller-supplied `x-custom-ui-hint` and the FC-2 markers (`MANAGED_KEY`, `RESOLVE_KEY`) survive the round-trip.

lucasmundim (approved):
Round-3 review — all round-2 items verified in d724eb5:

- typing-extensions floor relaxed to `>=4.6` to match pydantic
- `required: []` omitted when every required field is managed (top-level and nested); new `required_handled` flag handles dict-iteration order both ways
- `$schema: "https://json-schema.org/draft/2020-12/schema"` injected in `emit_schema()` via `setdefault`
- `ResolveHint` added to `__all__` and re-exported from `openfilter.filter_runtime`
- `json_schema_extra` merge behavior locked in by test

The new `@deprecated` decorator on legacy `FilterConfig` is well-scoped — verified third-party subclassers still see the warning under `python -W error::DeprecationWarning`, and openfilter-internal ones are correctly silenced via the module-name regex. The `append=False` choice is deliberate and well-commented.

46/46 tests in `test_filter_config_base.py` pass.

PR #87, opened 2026-05-05
feat(cli): openfilter emit-schema (FILTER-442)

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Adds the `openfilter emit-schema <module>[:Class]` CLI subcommand — the build-time producer half of the FC-3 schema-transport contract. The command imports a filter module, locates its `FilterConfigBase` subclass, and writes the emitted JSON Schema (draft 2020-12) to stdout (default) or `-o <path>`. Filter Dockerfiles invoke this at build time; the resulting JSON is pushed as a sibling OCI artifact (`application/vnd.openfilter.schema+json`) and the image manifest is stamped with a `com.plainsight.openfilter.schema` label carrying the artifact's sha256 digest.

Specifically:

- New `openfilter/cli/cmd_emit_schema.py` — argparse-driven subcommand with `module` / `module:Class` resolution, automatic candidate detection (errors out cleanly when a module declares 0 or 2+ `FilterConfigBase` subclasses), `--include-managed` opt-in to surface platform-managed fields, and `-o <path>` output redirection.
- Tweak the `openfilter` CLI dispatcher (`openfilter/cli/__main__.py`) to translate `cmd-with-hyphens` to `cmd_with_hyphens` so subcommands can use their canonical hyphenated spellings without breaking the underscore-named Python handler convention.
- 5 unit tests in `tests/test_emit_schema_cli.py` covering stdout output, `-o <path>`, `--include-managed`, single-class auto-pick, and unknown-class failure.

The wiring of the sibling-artifact upload + label-stamping into `cloudbuild-cascade.yaml` is tracked as a sub-step of [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442); it depends on PLAT-side FC-5 ingest and lands separately.

---

## 🔍 Why is this needed?

[FILTER-441](https://plainsight-ai.atlassian.net/browse/FILTER-441) landed `FilterConfigBase.emit_schema()` — the in-process serializer that produces each filter's JSON Schema. This PR is the CLI wrapper Dockerfiles invoke at build time so the artifact can ride alongside the image without a Python-runtime hop in the platform's ingest path.

Once both are in place, the transport contract closes: filter CI emits the schema, pushes it as a sibling OCI artifact, and stamps the image manifest's `com.plainsight.openfilter.schema` label with the sha256 digest. The platform fetches the artifact by digest from the same registry without pulling the filter image's layers. Integrity is transitively protected by whatever trust we already place in the registry serving the image bytes — no separate key infrastructure needed.

(Cosign-signed attestations were considered for the transport but deferred to a future-driver-only ticket — see
[FILTER-448](https://plainsight-ai.atlassian.net/browse/FILTER-448), closed Won't Do — the marginal threat surface didn't justify standing up KMS for the first internal use.)

---

## 🧪 How was it tested?

5 unit tests in `tests/test_emit_schema_cli.py`:

- `test_emit_schema_cli_stdout` — `module:Class` spec, stdout output, `$schema` keyword + managed-field stripping
- `test_emit_schema_cli_writes_file` — `-o <path>` redirection
- `test_emit_schema_cli_include_managed_surfaces_overrides` — `--include-managed` round-trips FC-2 markers
- `test_emit_schema_cli_auto_picks_single_class` — module-only spec finds the lone `FilterConfigBase` subclass
- `test_emit_schema_cli_fails_on_unknown_class` — typo'd class name errors out cleanly

```
$ uv run --with pytest pytest tests/test_emit_schema_cli.py -q
.....                                                                    [100%]
5 passed in 1.79s
```

Combined with [FILTER-441](https://plainsight-ai.atlassian.net/browse/FILTER-441)'s config-side suite (already merged in #86, 46 tests), the full FC-1 / FC-1.5 / FC-2 / FC-3-emission contract has 51 tests exercising it end-to-end.

---

## 🔗 Related Issues

- [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) — this ticket
- [FILTER-441](https://plainsight-ai.atlassian.net/browse/FILTER-441) — `FilterConfigBase.emit_schema()` (merged in #86)
- [FILTER-440](https://plainsight-ai.atlassian.net/browse/FILTER-440) — Declarative Configuration Contract (parent epic)
- [FILTER-448](https://plainsight-ai.atlassian.net/browse/FILTER-448) — cosign / KMS signing (closed Won't Do; revisit on concrete compliance/multi-tenancy driver)

---

```
$ openfilter emit-schema my_filter.filter --include-managed
{
  "title": "MyFilterConfig",
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "properties": {
    "sources": {
      "items": {"format": "video-source", "type": "string"},
      "type": "array",
      "x-openfilter-managed": true,
      "x-openfilter-resolve": "orchestrator-generated"
    },
    "confidence_threshold": {
      "type": "number", "minimum": 0, "maximum": 1, "default": 0.5
    },
    ...
  }
}
```

---

## ✅ Checklist

- [x] I have added or updated **documentation** as needed (`docs/declarative-config.md`'s "Schema emission" section, landed with FILTER-441 in #86, already describes this CLI's surface)

### DIFF
--- a/openfilter/cli/cmd_emit_schema.py
+++ b/openfilter/cli/cmd_emit_schema.py
@@ -1,0 +1,121 @@
+"""``openfilter emit-schema`` — Goldenrod.2 FC-1 / FC-3.
+
+Imports a filter module, locates its ``FilterConfigBase`` subclass, and writes
+the emitted JSON Schema to stdout (or ``-o <path>``). This is the build-time
+half of the schema-emission contract; FC-3's transport (manifest attestation /
+sigstore / ORAS) is layered on top — the CLI's only job is to produce the
+artifact.
+"""
+
+from __future__ import annotations
+
+import argparse
+import importlib
+import inspect
+import json
+import logging
+import os
+import sys
+from typing import Iterable
+
+from openfilter.filter_runtime.config import FilterConfigBase
+
+from .common import SCRIPT
+
+logger = logging.getLogger(__name__)
+logger.setLevel(int(getattr(logging, (os.getenv("LOG_LEVEL") or "INFO").upper())))
+
+
+def _candidate_classes(module: object) -> Iterable[type[FilterConfigBase]]:
+    for _, attr in inspect.getmembers(module, inspect.isclass):
+        if attr is FilterConfigBase:
+            continue
+        if not issubclass(attr, FilterConfigBase):
+            continue
+        if attr.__module__ != getattr(module, "__name__", None):
+            continue
+        yield attr
+
+
+def _resolve_config_class(spec: str) -> type[FilterConfigBase]:
+    """Resolve a module path or ``module:Class`` into a ``FilterConfigBase`` subclass."""
+    module_name, _, class_name = spec.partition(":")
+    module = importlib.import_module(module_name)
+
+    if class_name:
+        cls = getattr(module, class_name, None)
+        if cls is None:
+            raise SystemExit(f"{spec!r}: {class_name} not found in {module_name}")
+        if not (inspect.isclass(cls) and issubclass(cls, FilterConfigBase)):
+            raise SystemExit(
+                f"{spec!r}: {class_name} is not a FilterConfigBase subclass"
+            )
+        return cls
+
+    candidates = list(_candidate_classes(module))
+    if not candidates:
+        raise SystemExit(
+            f"{spec!r}: no FilterConfigBase subclass declared in {module_name}. "
+            f"Pass module:ClassName to disambiguate or check that the filter is migrated."
+        )
+    if len(candidates) > 1:
+        names = ", ".join(c.__name__ for c in candidates)
+        raise SystemExit(
+            f"{spec!r}: multiple FilterConfigBase subclasses found ({names}). "
+            f"Pass module:ClassName to disambiguate."
+        )
+    return candidates[0]
+
+
+def cmd_emit_schema(args: list[str]) -> None:
+    parser = argparse.ArgumentParser(
+        prog=f"{SCRIPT} emit-schema",
+        description=(
+            "Emit a filter's JSON Schema (draft 2020-12). "
+            "MODULE is an importable dotted path; optionally suffix with ':ClassName' "
+            "to disambiguate when a module declares more than one config class."
+        ),
+    )
+    parser.add_argument(
+        "MODULE",
+        help="Importable filter module (e.g. filter_template.filter) "
+        "or module:ConfigClass.",
+    )
+    parser.add_argument(
+        "-o",
+        "--output",
+        default="-",
+        help="Output path; '-' (default) writes to stdout.",
+    )
+    parser.add_argument(
+        "--include-managed",
+        action="store_true",
+        help="Include platform-managed fields in the emitted schema "
+        "(default: operator-facing surface only).",
+    )
+    parser.add_argument(
+        "--indent",
+        type=int,
+        default=2,
+        help="JSON indent (default 2; pass 0 for compact).",
+    )
+
+    opts = parser.parse_args(args)
+    cls = _resolve_config_class(opts.MODULE)
+
+    schema = cls.emit_schema(include_managed=opts.include_managed)
+    schema.setdefault("title", cls.__name__)
+    schema.setdefault("$schema", "https://json-schema.org/draft/2020-12/schema")
+
+    text = json.dumps(schema, indent=opts.indent or None, sort_keys=False)
+
+    if opts.output == "-":
+        sys.stdout.write(text)
+        if not text.endswith("\n"):
+            sys.stdout.write("\n")
+    else:
+        with open(opts.output, "w", encoding="utf-8") as f:
+            f.write(text)
+            if not text.endswith("\n"):
+                f.write("\n")
+        logger.info("wrote schema for %s to %s", cls.__qualname__, opts.output)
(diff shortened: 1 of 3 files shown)


lucasmundim (approved):
LGTM. Clean dispatcher tweak, correct use of the `FilterConfigBase.emit_schema(include_managed=...)` API from #86, and a reasonable test set. A few non-blocking notes for follow-up:

- **Non-atomic write to `-o <path>`**: `cmd_emit_schema.py` writes directly with `open(path, "w")`. If this ever feeds OCI artifact upload at build time, a `tempfile.NamedTemporaryFile` on the same filesystem + `os.replace()` would avoid leaving a partial file on interruption.
- **Unwrapped `importlib.import_module`** in `_resolve_config_class`: every other failure branch raises a clean `SystemExit(...)`, but a bad module path falls through as a raw `ModuleNotFoundError` traceback. Minor consistency nit.
- **Dead `setdefault` calls** for `title` and `$schema` in the CLI: both keys are already injected by pydantic and `FilterConfigBase.emit_schema()` respectively, so the CLI's `setdefault` calls are no-ops and can be removed.
- **Auto-detect edge cases** worth keeping in mind: module-level class aliases (`Alias = MyConfig`) would be yielded twice by `inspect.getmembers`, and package re-exports (`__init__.py` re-exporting a class defined in a submodule) would fail the `__module__` check. Neither pattern is used in the repo today — flagging in case third-party filters hit them.

openfilter/cli/cmd_emit_schema.py
```diff
@@ -0,0 +1,121 @@
...
+import logging
+import os
+import sys
+from typing import Iterable
+
+from openfilter.filter_runtime.config import FilterConfigBase
+
+from .common import SCRIPT
+
+logger = logging.getLogger(__name__)
```
shingonoide, line 25: Module-level `getattr(logging, NAME)` with no default raises `AttributeError` on any typo'd `LOG_LEVEL`, which crashes the import of this module — and since `__main__.py` imports it unconditionally, the entire CLI breaks for every subcommand, not just `emit-schema`. Suggested:

```python
logger.setLevel(getattr(logging, (os.getenv("LOG_LEVEL") or "INFO").upper(), logging.INFO))
```

(`int(...)` is also unneeded — `logging.INFO` is already an int.)

shingonoide (commented):
Echoing Lucas's APPROVE — adding one finding worth picking up alongside his follow-ups, plus a test-coverage note.

**`LOG_LEVEL` typo crashes the entire `openfilter` CLI at import time (`cmd_emit_schema.py:25`)**

```python
logger.setLevel(int(getattr(logging, (os.getenv("LOG_LEVEL") or "INFO").upper())))
```

`getattr(logging, "VERBOSE")` (or any other typo / unsupported level) raises `AttributeError` at module import. Because `openfilter/cli/__main__.py` imports `cmd_emit_schema` unconditionally, a bad `LOG_LEVEL` in the build environment doesn't just break `emit-schema` — it breaks `openfilter run`, `openfilter info`, and any future subcommand the dispatcher tries to load. Suggested fix:

```python
logger.setLevel(getattr(logging, (os.getenv("LOG_LEVEL") or "INFO").upper(), logging.INFO))
```

(The `int(...)` cast is also unneeded — `logging.INFO` is already an int.)

**Test coverage gaps tied to the import paths**

The 5 tests cover happy paths well; two negative-path gaps worth filling alongside Lucas's `import_module` wrap:
- No test for `ModuleNotFoundError` (e.g. `["emit-schema", "no.such.module"]`) — would lock in his suggested wrap.
- `test_emit_schema_cli_fails_on_unknown_class` exercises `getattr(...) is None`, but not the `inspect.isclass(cls) and not issubclass(cls, FilterConfigBase)` branch (`cmd_emit_schema.py:48-52`). Passing a non-Filter attribute (`module:dict`) would close that hole.

openfilter/cli/cmd_emit_schema.py
stwilt: This is the project-wide pattern, not a regression in this PR — the same unguarded `int(getattr(logging, LOG_LEVEL))` shape ships at:

- [`openfilter/cli/cmd_run.py:15`](https://github.com/PlainsightAI/openfilter/blob/a6dc1c39ffd059b4d532eb49e11565ed43844e37/openfilter/cli/cmd_run.py#L15)
- [`openfilter/cli/cmd_info.py:12`](https://github.com/PlainsightAI/openfilter/blob/a6dc1c39ffd059b4d532eb49e11565ed43844e37/openfilter/cli/cmd_info.py#L12)
- [`openfilter/filter_runtime/filter.py:123,160`](https://github.com/PlainsightAI/openfilter/blob/a6dc1c39ffd059b4d532eb49e11565ed43844e37/openfilter/filter_runtime/filter.py#L123-L160)
- ~10 test modules under `tests/`

A typo'd `LOG_LEVEL` already breaks `openfilter run` and `openfilter info` at import — `__main__.py` would have hit this with or without `emit-schema`. Switching only this file to a `getattr(..., logging.INFO)` fallback would make it the odd one out. Worth doing, but as a uniform sweep (probably folded into `cli/common.py`'s `level_name` resolution so every subcommand inherits it) — out of scope for the FC-1/FC-3 schema-emission work here. Happy to take a follow-up PR if you want to drive it.

PR #88, opened 2026-05-06
feat(filter_runtime): FilterOutputSchema + frame.data shape catalog (…

stwilt:
…FILTER-444)

## What does this PR do?

Adds the SDK affordance for filters to declare their `frame.data` output as a build-time JSON Schema (draft 2020-12), parallel to `FilterConfigBase` from [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442). Ships:

- `FilterOutputSchema(BaseModel)` in `openfilter.filter_runtime.output` with `__schema_id__` (`$id`) and `__frame_data_key__` (dotted path on `frame.data`, surfaced as the `x-openfilter-frame-data-key` extension). No managed/resolve filtering — output schemas describe data, not config.
- `openfilter.filter_runtime.shapes` catalog: `BoundingBox`, `Polygon`, `Mask`, `Keypoint`, `Detection`, `DetectionSet`, `Track`, `TrackSet`, `Pose`, `PoseSet`, `KeypointSet`, `OCRSpan`, `OCRSpanSet`, `ClassificationResult`. Each with stable `$id` at `https://schemas.plainsight.ai/shapes/<kebab>/v1` and per-shape coordinate conventions documented inline (pixel xyxy for boxes / polygons / masks / OCR quads; normalized [0,1] for keypoints).
- `openfilter emit-schema --kind {config,output}` flag, defaulting to `config` to preserve the [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) contract. Auto-detection respects the selected base class; `--include-managed` is rejected with `--kind output`.

## Why is this needed?

Consumers of `frame.data` (a shared visualizer, webvis SSE clients, GT label visualization, downstream filters) currently negotiate shapes out-of-band — every consumer hard-codes its own dialect against every producer. Declaring output as a schema lets all consumers bind against one interop type.

The catalog is **normative, not descriptive** — it defines what filters *should* emit, with per-filter migrations as separate follow-ups (catalog shapes were seeded from the superset of what `filter-sam3-detector`, `filter-protege-model`, `filter-optical-character-recognition`, and `filter-pose-estimation` currently emit on `frame.data`). Filters that don't want to use the catalog can ship their own bespoke `FilterOutputSchema` with a filter-owned `$id`; catalog `$ref`s are an opt-in convenience, never a gate.

## How was it tested?

- 26 new unit tests in `tests/test_output_schema.py` covering base-class semantics (`$id` / frame-data-key emission, draft 2020-12 conformance, catalog `$ref` resolution into `$defs`, bespoke schema with no catalog dep), every catalog shape's `$id`, and shape-specific validation (BoundingBox pixel xyxy, Keypoint normalized range, Detection open-set label-only, Track inheritance, OCRSpan quad arity, Pose skeleton convention, ClassificationResult multilabel toggle).
- 5 new CLI tests in `tests/test_emit_schema_cli.py` covering `--kind output` happy path, auto-detection of a single output class, cross-base rejection (config kind on output class and vice versa), and `--include-managed` incompatibility with `--kind output`.
- Full suite: 559 passing (36 new + 523 unchanged); 42 pre-existing failures in `test_filter_video_out.py` / `test_filter_rest.py` / `test_filter_webvis.py` are gated on `av` / `fastapi` extras absent from `--extra dev` and unrelated to this change.

## Related Issues

- Implements [FILTER-444](https://plainsight-ai.atlassian.net/browse/FILTER-444) — SDK half.
- Workflow plumbing for emit + OCI sibling-artifact attach (config + output) is split into [FILTER-450](https://plainsight-ai.atlassian.net/browse/FILTER-450).
- Sibling to [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) (config schema CLI, shipped in #87).

## Screenshots or Logs

N/A — pure SDK module; no UI surface.

## Checklist


shingonoide (commented):
Solid SDK half of FILTER-444. Schema emission is symmetric with FILTER-442 where it should be (default `--kind config` preserved, `--include-managed` mutex with `--kind output` covered by tests, draft 2020-12 stamped) and intentionally divergent where it should be (no managed/resolve filtering on the output path). Catalog coverage matches the production filters cited. A handful of non-blocking items, none gating.

### Non-blocking

- **Public-API asymmetry with `FilterConfigBase`** — `openfilter/filter_runtime/__init__.py` re-exports `FilterConfigBase`, `MANAGED_KEY`, `Managed`, `Resolve`, but the FILTER-444 surface (`FilterOutputSchema`, `FRAME_DATA_KEY`, catalog shapes) is reachable only via the deeper `openfilter.filter_runtime.output` / `.shapes` paths. Filters following the FILTER-442 import idiom (`from openfilter.filter_runtime import FilterConfigBase`) hit `ImportError` on the corresponding output line. Both `output.py` and `shapes.py` already declare `__all__`; just add the imports to `filter_runtime/__init__.py`.

- **`--kind output` auto-detect breaks with helper subclasses in the same module** — `cmd_emit_schema.py:40-48` collects every `FilterOutputSchema` subclass defined in the target module. If a filter declares its top-level output class plus one or more bespoke nested shapes in the same file, the CLI errors with "multiple found" and forces `module:Class`. Consider preferring a class with non-empty `__frame_data_key__` when disambiguating, or document the constraint.

- **`ClassificationResult.classes` / `.confidences` parallel-array length not validated** — `shapes.py:357-358` declares both as independent `list[]`s; the docstring (line 348) calls them "parallel arrays" but no `model_validator` enforces `len(classes) == len(confidences)`. `test_classification_result_supports_multilabel` doesn't exercise mismatch. A `model_validator(mode="after")` closes this.

- **`__schema_id__` / `__frame_data_key__` not validated on subclass override** — `output.py:55-56` declares them as `ClassVar[str | None]` / `ClassVar[str]` but the base never checks subclass overrides. A subclass setting `__schema_id__ = 42` or `__frame_data_key__ = None` produces a broken schema / silent dotted-path bug with no helpful error. An `__init_subclass__` validating type + non-empty `__schema_id__` URI would fail-fast at class-construction time.

- **`BoundingBox` xyxy ordering documented but not enforced** — docstring at `shapes.py:65-71` promises `(x1,y1)` top-left and `(x2,y2)` bottom-right, but no validator rejects an inverted box. `BoundingBox(x1=100, y1=100, x2=10, y2=10)` constructs cleanly. Either add a `model_validator` or weaken the docstring to "ordering by convention; consumers normalize."

### Nits

- `cmd_emit_schema.py:31` — `int(getattr(logging, (os.getenv("LOG_LEVEL") or "INFO").upper()))` raises `TypeError`/`AttributeError` at import on unknown `LOG_LEVEL` (e.g. `LOG_LEVEL=trace`). Same pattern as FILTER-442; if it's already filed, ignore.

- `Keypoint.z` (`shapes.py:144-147`) is unbounded while x/y/visibility are bounded `[0,1]`. Docstring notes "MediaPipe emits it; RTMPose sets 0"; consider documenting the valid range or leaving `Field(ge=...)` off intentionally with a comment.

- `Pose.keypoints` declares `skeleton: Literal["coco-17"]` but does not validate `len(keypoints) == 17`; downstream consumers can't trust the ordering convention without arity enforcement.

- Catalog classes set `model_config = ConfigDict(json_schema_extra={"$id": __schema_id__})` and `emit_schema()` also injects `$id` from `__schema_id__`. For top-level emission both write the same value, so it's idempotent, but two sources of truth for the same field is fragile. Consider relying solely on `emit_schema()`'s injection. (Confirmed harmless under draft 2020-12: emitted `$ref`s are JSON-Pointer fragments, not URI-relative, so nested `$id`s in `$defs` don't re-base resolution in practice.)

openfilter/filter_runtime/shapes.py
```diff
@@ -0,0 +1,362 @@
...
+    """Axis-aligned box in **pixel** coordinates, ``xyxy`` ordering.
+
+    ``(x1, y1)`` is the top-left corner; ``(x2, y2)`` is the bottom-right.
+    Pixel-space because the production detectors (``filter-sam3-detector``,
+    ``filter-protege-model``) emit pixel xyxy. Detectors that emit normalized
+    or ``cxcywh`` declare a bespoke ``FilterOutputSchema`` instead.
+    """
+
+    __schema_id__: ClassVar[str] = _shape_id("bounding-box")
+    model_config = ConfigDict(json_schema_extra={"$id": __schema_id__})
```
lucasmundim: Every catalog shape repeats this two-line pattern (`__schema_id__` plus `model_config = ConfigDict(json_schema_extra={"$id": __schema_id__})`). The `model_config` line is load-bearing — without it, `__schema_id__` only stamps the top-level schema in `emit_schema()`, but pydantic doesn't propagate it into `$defs[<Name>]` when the shape is referenced from another schema (e.g. `Detection` inside a filter-author output). So a future shape author who forgets the line silently breaks `$ref` resolution for downstream consumers.

Consider auto-merging in `FilterOutputSchema.__init_subclass__` so subclasses only need to set `__schema_id__`:

```python
def __init_subclass__(cls, **kwargs):
    super().__init_subclass__(**kwargs)
    if cls.__schema_id__:
        extra = dict(cls.model_config.get("json_schema_extra") or {})
        extra.setdefault("$id", cls.__schema_id__)
        cls.model_config["json_schema_extra"] = extra
```

Eliminates the 14x duplication across the catalog and the foot-gun for filter authors writing bespoke `FilterOutputSchema` subclasses.

openfilter/filter_runtime/shapes.py
```diff
@@ -0,0 +1,362 @@
...
+
+    x1: float = Field(description="Left edge, pixel coordinates.")
+    y1: float = Field(description="Top edge, pixel coordinates.")
+    x2: float = Field(description="Right edge, pixel coordinates.")
+    y2: float = Field(description="Bottom edge, pixel coordinates.")
```
lucasmundim, line 81: No invariant on `x2 >= x1` / `y2 >= y1`. A filter emitting an inverted or zero-area box validates fine, which undercuts the catalog being normative ("what filters should emit"). Consider a `@model_validator(mode='after')` rejecting `x2 < x1` or `y2 < y1` — cheap, and catches a real production bug class for downstream consumers binding against this shape.

openfilter/filter_runtime/output.py
```diff
@@ -0,0 +1,76 @@
...
+      whole ``frame.data`` namespace. Surfaced in the emitted schema as the
+      ``x-openfilter-frame-data-key`` extension so consumers know where to
+      bind without inspecting filter source.
+
+    Catalog shapes leave ``__frame_data_key__`` empty — they're nested types
+    referenced via ``$ref``, not standalone ``frame.data`` declarations.
+    """
+
+    __schema_id__: ClassVar[str | None] = None
+    __frame_data_key__: ClassVar[str] = ""
```
lucasmundim: `__frame_data_key__ = ""` is documented as "the whole `frame.data` namespace" (see the docstring at line 47), but `emit_schema()` (lines 74-75) elides the `x-openfilter-frame-data-key` marker entirely when empty. Consumers can't distinguish "schema describes the whole `frame.data`" from "author forgot to declare a key". Two options:

- default to `None`, always emit the extension when set (explicit "unset");
- always emit `""` for the whole-namespace case so consumers can dispatch on presence.

Either makes consumer logic less brittle. Catalog shapes leave it empty by design (they're nested types), so this only matters for filter-author subclasses — but those are exactly the ones consumers bind against.

lucasmundim (commented):
Schema-emission half of FILTER-444 is well-structured: `FilterOutputSchema` mirrors the `FilterConfigBase` pattern cleanly (no managed/resolve filtering since it's data, not config), the catalog `$id`s are stable and slug-conventional, and coordinate semantics are documented inline per shape. Three things worth a closer look — boilerplate that invites drift, a missing geometric invariant, and one frame-data-key sentinel ambiguity. None are blockers.

openfilter/filter_runtime/shapes.py
stwilt: Done in 1305477. `FilterOutputSchema.__init_subclass__` now auto-merges `$id` into `model_config["json_schema_extra"]` via `setdefault`, so subclasses only set `__schema_id__` and bespoke shape authors can't silently break `$ref` resolution. Dropped the 14× redundant `model_config` lines across the catalog and added a regression test that asserts `defs["Detection"]["$id"] == Detection.__schema_id__` when a filter `$ref`s the catalog. `setdefault` semantics keep explicit class-level overrides winning (covered by `test_init_subclass_explicit_model_config_wins`).

openfilter/filter_runtime/shapes.py
stwilt: Done — added a `model_validator(mode="after")` rejecting `x2 < x1 || y2 < y1` with a structured error message. Zero-area boxes (`x2 == x1`) are still allowed because production NNs do emit them and rejecting would force every consumer to filter — covered by `test_bounding_box_allows_zero_area`. Inversion regression covered by `test_bounding_box_rejects_inverted_xyxy`.

openfilter/filter_runtime/output.py
stwilt: Went with option 1 (default to `None`, always emit when set):

* `__frame_data_key__: ClassVar[str | None] = None` — `None` means unset (marker absent), which is what catalog shapes inherit since they're nested types.
* `""` now unambiguously means "whole `frame.data` namespace" and is always emitted as `"x-openfilter-frame-data-key": ""` so consumers can dispatch on presence vs. value.
* `__init_subclass__` validates the type (rejects non-string non-None overrides at class-construction time).

Tests: `test_empty_frame_data_key_emits_marker`, `test_catalog_shapes_leave_frame_data_key_unset`, `test_init_subclass_rejects_non_string_frame_data_key`.

stwilt:
Thanks for the careful read, both — addressed in 1305477.

### Non-blocking (shingonoide)

* **Public-API asymmetry** — added `FilterOutputSchema`, `FRAME_DATA_KEY`, `SHAPE_ID_BASE`, and the catalog shapes to `openfilter.filter_runtime.__init__`. FILTER-442 import idiom now works for outputs too.
* **`--kind output` auto-detect with helper subclasses** — `cmd_emit_schema._resolve_class` now prefers the candidate with a non-None `__frame_data_key__` (filter-author top-level output anchors a frame.data path; helpers leave it unset). Two anchored candidates remain ambiguous and fall back to the multi-class error — `test_emit_schema_cli_output_kind_fails_when_multiple_anchored`. Added a `_SynthHelperShape` to the existing CLI fixture so the auto-pick-anchored happy path is exercised too.
* **`ClassificationResult` parallel-array length** — `model_validator(mode="after")` enforces `len(classes) == len(confidences)`. `test_classification_result_rejects_parallel_array_mismatch`.
* **`__schema_id__` / `__frame_data_key__` subclass-override validation** — `__init_subclass__` raises `TypeError` for non-string or empty `__schema_id__`, and non-string `__frame_data_key__`. Tests: `test_init_subclass_rejects_non_string_schema_id`, `test_init_subclass_rejects_empty_schema_id`, `test_init_subclass_rejects_non_string_frame_data_key`.
* **`BoundingBox` ordering** — covered in the inline thread.

### Nits

* **`Keypoint.z` unbounded** — tightened the description to call out that the range is backend-defined and intentionally unbounded (MediaPipe vs. RTMPose). Left `Field(ge=…)` off intentionally; the docstring now signals intent.
* **`Pose.keypoints` arity for `coco-17`** — `model_validator(mode="after")` enforces 17 entries when `skeleton == "coco-17"`. Skeleton-unset cases skip the check (`test_pose_without_skeleton_skips_arity_check`); wrong arity covered by `test_pose_coco17_rejects_wrong_arity`.
* **Two sources of truth for `$id`** — resolved by the auto-merge in `__init_subclass__` (see the inline thread). One declaration site (`__schema_id__`); `model_config` is populated automatically; `emit_schema()` still stamps the top-level `$id` for standalone retrieval.
* **`LOG_LEVEL` brittleness in `cmd_emit_schema.py:31`** — leaving as-is here since it's the same pattern as FILTER-442 and out of scope for the FILTER-444 surface. Happy to file a follow-up that hardens both call sites in one go if you want a ticket.

openfilter/filter_runtime/output.py
```diff
@@ -0,0 +1,109 @@
...
+                    )
+                # Pydantic does not propagate the top-level ``$id`` into
+                # ``$defs[<Name>]`` when this class is referenced from another
+                # schema; stamping it via ``json_schema_extra`` does. Use
+                # ``setdefault`` so an explicit class-level override wins.
+                existing = cls.model_config.get("json_schema_extra") or {}
+                if isinstance(existing, dict):
+                    extra = dict(existing)
+                    extra.setdefault("$id", schema_id)
+                    cls.model_config["json_schema_extra"] = extra
```
lucasmundim: **Silent `$id` inheritance when subclassing a catalog shape.**

`__init_subclass__` only stamps `json_schema_extra["$id"]` when `__schema_id__` is in `cls.__dict__` (correct — only set on classes that explicitly declare it). But `cls.__schema_id__` resolves through MRO, and pydantic copies the parent's `model_config` into the subclass. Net effect: a filter author who writes

```python
class MyExtendedDetection(Detection):
    extra_field: int = 0
```

gets `MyExtendedDetection.emit_schema()["$id"] == "https://schemas.plainsight.ai/shapes/detection/v1"` — the parent's id stamped on a different class. I confirmed this locally:

```
MyExtendedDetection model_config: {'json_schema_extra': {'$id': '.../detection/v1'}}
emit $id: https://schemas.plainsight.ai/shapes/detection/v1
```

JSON Schema `$id` is a stable identity for `$ref`, so this collides on the wire — two distinct schemas claiming to be `Detection`. It's the kind of bug that'll surface only when a consumer's `$ref` cache hands back the wrong schema.

Two options:

1. **Detect at subclass time** — in `__init_subclass__`, if `__schema_id__` is not in `cls.__dict__` *and* a parent has `$id` in its `model_config.json_schema_extra`, raise a `TypeError` telling the author to override `__schema_id__` (or set it to `None` to opt out).
2. **Document only** — add a docstring note that subclassing a `__schema_id__`-bearing class requires overriding the id.

I'd lean toward (1) — silent identity collisions are nasty to debug, and the runtime check is cheap. A test for the inherit-without-override case would lock in whichever decision is made.

openfilter/cli/cmd_emit_schema.py
```diff
@@ -19,52 +23,72 @@
...
+    # subclass with one or more bespoke nested helpers in the same file. The
+    # top-level one anchors a frame.data path (non-None __frame_data_key__);
+    # helpers leave it None. When exactly one candidate is anchored, prefer it.
+    if base is FilterOutputSchema:
+        anchored = [
+            c for c in candidates
+            if getattr(c, "__frame_data_key__", None) is not None
+        ]
+        if len(anchored) == 1:
+            return anchored[0]
```
lucasmundim, line 85: **Asymmetric auto-detect between `--kind config` and `--kind output`.**

Output kind has the helper-vs-anchored heuristic above; config kind keeps the original strict "multiple = error". The asymmetry is justified by the inline comment ("filter authors commonly co-locate"), but a config module that ships a base + a subclass (legitimate pattern) hits the error and forces `module:Class`.

Not blocking — just worth a note in the help string, or a parallel future improvement so authors who learn the looser output behavior first aren't surprised by config.

openfilter/cli/cmd_emit_schema.py
```diff
@@ -101,9 +132,18 @@ def cmd_emit_schema(args: list[str]) -> None:
...
+    base = _BASES[opts.kind]
+    cls = _resolve_class(opts.MODULE, base)
+
+    if opts.kind == "config":
+        schema = cls.emit_schema(include_managed=opts.include_managed)
+    else:
+        schema = cls.emit_schema()
+
     schema.setdefault("title", cls.__name__)
     schema.setdefault("$schema", "https://json-schema.org/draft/2020-12/schema")
```
lucasmundim: **Redundant `setdefault`.**

Both `title` and `$schema` are already set inside the `emit_schema()` implementations (pydantic stamps `title`; both `FilterConfigBase.emit_schema` and `FilterOutputSchema.emit_schema` set `$schema`). Defensive but dead. Either drop these two lines or add a comment that this is belt-and-suspenders for someone who overrides `emit_schema` and forgets.

openfilter/filter_runtime/shapes.py
```diff
@@ -0,0 +1,387 @@
...
+    @model_validator(mode="after")
+    def _validate_parallel_arrays(self) -> "ClassificationResult":
+        if len(self.classes) != len(self.confidences):
+            raise ValueError(
+                f"ClassificationResult.classes and confidences must be "
+                f"parallel arrays of equal length; got "
+                f"{len(self.classes)} classes and "
+                f"{len(self.confidences)} confidences"
+            )
+        return self
```
lucasmundim, line 387: **Runtime-only invariant not expressed in the emitted schema.**

`_validate_parallel_arrays` enforces `len(classes) == len(confidences)` at pydantic construction time, but the schema returned by `emit_schema()` has no `allOf`/`if` to express it. Confirmed:

```json
"properties": {
  "classes": {"items": {"type": "string"}, "type": "array"},
  "confidences": {"items": {"type": "number"}, "type": "array"},
  ...
}
```

A consumer validating raw JSON against the schema alone won't catch a length mismatch. Same gap exists for `BoundingBox._validate_xyxy_ordering` and `Pose._validate_skeleton_arity`.

Acceptable as long as it's documented — consumers using pydantic to deserialize get the validators for free. Worth a one-liner in either the module docstring or each affected class's docstring saying "JSON Schema underspecifies; validate via pydantic for full constraints.

lucasmundim (commented):
## Review summary

Solid PR — design is sound, docstrings carry their weight, and the fixup commit shows real responsiveness to earlier review (auto-stamp via `__init_subclass__`, type validation, sentinel semantics for `__frame_data_key__`). I ran the new tests locally (49 passed) and stress-tested a few edge cases.

**Verdict:** Approve with minor suggestions. The only thing I'd lobby to address before merge is the silent `$id` inheritance when subclassing a catalog shape (inline) — that's a foot-gun. The rest are documentation/polish.

### Strengths

- Excellent docstrings — every catalog shape documents its coordinate convention and the production filter that motivated each choice.
- The `__init_subclass__` comment explaining *why* it stamps `json_schema_extra` (pydantic doesn't propagate top-level `$id` into `$defs[<Name>]` otherwise) is load-bearing context future readers will need.
- 36 new tests covering base-class semantics, every catalog shape's `$id`, validators, and CLI happy/unhappy paths. The parametrized `test_catalog_shape_has_expected_id` is a nice sweep.
- Auto-detect heuristic (prefer the single anchored class) is well-motivated and tested both ways.

### Smaller observations (no inline)

- **Asymmetry between `--kind config` and `--kind output` auto-detect.** Output kind has the helper-vs-anchored heuristic; config kind keeps the strict "multiple = error". The asymmetry is justified by your comment, but worth noting in the help string or as a parallel future improvement so users who learn the looser output behavior first aren't surprised by config.
- **Schema doesn't express runtime-only invariants.** `BoundingBox` xyxy ordering, `ClassificationResult` parallel arrays, and `Pose` `coco-17` arity are pydantic-side validators only; consumers validating raw JSON against `emit_schema()` won't catch these. Acceptable (consumers using pydantic to deserialize get the validator for free), but worth acknowledging in the module docstring so consumers know to validate via pydantic, not just the schema.
- **`Detection.label_id`** lacks a closed-set/open-set discriminator in the schema — only the docstring mentions the convention. If downstream consumers need to dispatch on it, an extension key would make it observable. Out of scope here; raising for a future PR.

### Security / performance

No concerns. Pure type/schema declaration; `__init_subclass__` runs once per class definition; `importlib.import_module` was already present in #87.

openfilter/filter_runtime/output.py
stwilt: Good catch — confirmed locally too. Fixed in [bd5f8ea](https://github.com/PlainsightAI/openfilter/pull/88/commits/bd5f8ea) (filed and closing as [FILTER-452](https://plainsight-ai.atlassian.net/browse/FILTER-452) for traceability). Went with your option 1: `__init_subclass__` raises `TypeError` when a subclass inherits an `$id`-bearing `model_config` without overriding `__schema_id__`, with `__schema_id__ = None` as the explicit opt-out. While I was in there I also flipped the inner stamp from `setdefault` to direct assignment so a declared `__schema_id__` overrides any inherited parent `$id` rather than being shadowed by it. Three new tests cover the inheritance-raises path, the `None` opt-out, and the explicit-override case. Resolving.

openfilter/cli/cmd_emit_schema.py
stwilt: Agreed it's an asymmetry. Holding the loose heuristic to `--kind output` for now because that's where the co-located-helper pattern shows up in practice (the catalog itself is the proof: `Detection`/`DetectionSet` ship in the same module). `--kind config` rarely hits the same shape — config bases tend to be one-per-module — and `FilterConfigBase` doesn't have a clean discriminator analog to `__frame_data_key__` to anchor on. Will mirror the heuristic into config-kind if filter authors start tripping on it. Resolving.

openfilter/cli/cmd_emit_schema.py
stwilt: You're right — dead code. Dropped both lines in [bd5f8ea](https://github.com/PlainsightAI/openfilter/pull/88/commits/bd5f8ea) as a drive-by alongside the FILTER-452 fix. Resolving.

openfilter/filter_runtime/shapes.py
stwilt: You're right that JSON Schema 2020-12 can't express two of the three (cross-property arithmetic, array-length equality), and `Pose` `coco-17` *could* be expressed via `if`/`then` but pydantic doesn't auto-emit. Filed as [FILTER-454](https://plainsight-ai.atlassian.net/browse/FILTER-454) — defines a `https://schemas.plainsight.ai/meta/openfilter/v1` vocabulary URI with `x-openfilter-xyxy-ordering` / `x-openfilter-parallel-arrays` / `x-openfilter-conditional-arity` keywords, plus reference Ajv (TS) and santhosh-tekuri (Go) validators in a future shared package. The `Pose` `if`/`then` lands alongside that work for free interop with stock validators. Module-docstring note about the pydantic-vs-JSON-Schema gap rides with FILTER-454 since it'd be misleading to land it without the vocabulary. Resolving.

stwilt:
Thanks for the second pass — addressed the substantive items in two new commits:

- **[bd5f8ea](https://github.com/PlainsightAI/openfilter/pull/88/commits/bd5f8ea)** — [FILTER-452](https://plainsight-ai.atlassian.net/browse/FILTER-452) silent `$id` inheritance fix (your option 1: raise on inherited `$id` without override; `__schema_id__ = None` is the opt-out). Also drops the redundant `setdefault("title"/"$schema", ...)` in the CLI as a drive-by.
- **[d2ce68e](https://github.com/PlainsightAI/openfilter/pull/88/commits/d2ce68e)** — Bumps to `v0.2.0` with a RELEASE.md entry covering [FILTER-441](https://plainsight-ai.atlassian.net/browse/FILTER-441) / [FILTER-442](https://plainsight-ai.atlassian.net/browse/FILTER-442) / [FILTER-444](https://plainsight-ai.atlassian.net/browse/FILTER-444) and the FILTER-452 fix together. Minor bump matches the public-API surface added across the three sibling tickets.

Two follow-up tickets filed under the [FILTER-440](https://plainsight-ai.atlassian.net/browse/FILTER-440) epic:

- **[FILTER-453](https://plainsight-ai.atlassian.net/browse/FILTER-453)** (P2 story, Goldenrod.2) — FC-3 output-schema transport extension. The "ship the schemas without consumers downloading the container" piece, mirroring the FILTER-442 cloudbuild spike. Lives in `cloudbuild-cascade.yaml`, not openfilter-side.
- **[FILTER-454](https://plainsight-ai.atlassian.net/browse/FILTER-454)** (P3 story) — JSON Schema vocabulary URI for the runtime invariants stock validators can't represent. Defers until consumers in Go/TS show up.

Re your "smaller observations": asymmetric auto-detect is acknowledged inline (no clean discriminator analog on the config side; will mirror if it becomes friction). `Detection.label_id` closed/open-set discriminator I'm leaving for a future PR per your suggestion — real schema gap but downstream consumers haven't asked for it yet.

If the four threads + this summary look resolvable, ready to merge.

[FILTER-452]: https://plainsight-ai.atlassian.net/browse/FILTER-452?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FILTER-441]: https://plainsight-ai.atlassian.net/browse/FILTER-441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FILTER-444]: https://plainsight-ai.atlassian.net/browse/FILTER-444?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FILTER-440]: https://plainsight-ai.atlassian.net/browse/FILTER-440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FILTER-453]: https://plainsight-ai.atlassian.net/browse/FILTER-453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

openfilter/filter_runtime/output.py
```diff
@@ -0,0 +1,139 @@
...
+
+    * ``__schema_id__`` — the ``$id`` URI for the schema. First-party filters
+      use ``https://schemas.plainsight.ai/filters/<name>/v1``; third-party
+      filters stamp whatever URI they own. Catalog shapes use
+      ``https://schemas.plainsight.ai/shapes/<kebab>/v1``. Optional but
+      strongly recommended; without it the emitted schema has no stable
+      identity for ``$ref`` resolution. ``__init_subclass__`` propagates the
+      value into ``model_config.json_schema_extra`` so pydantic stamps it on
+      ``$defs[<Name>]`` when this class is referenced from another schema —
+      subclasses only need to set ``__schema_id__``.
```
lucasmundim, line 45: Worth surfacing the FILTER-452 contract here while you're in the file: subclassing a `$id`-bearing parent (catalog shape or filter-author output) without overriding `__schema_id__` is a `TypeError` at class-definition time, and `__schema_id__ = None` is the explicit opt-out. The error message itself documents this at the failure point, but readers reading top-down won't know the rule until they trip it. Optional / nit.

lucasmundim (approved):
Re-review of `bd5f8ea` and the v0.2.0 bump. The FILTER-452 fix is clean: `__init_subclass__` correctly refuses silent inheritance, `__schema_id__ = None` is a viable opt-out, and direct assignment (vs `setdefault`) ensures a child's declared `__schema_id__` wins over an inherited parent `$id`. CLI cleanup is correct — both `setdefault`s were dead since pydantic / `emit_schema` already stamp `title` and `$schema`. Verified manually that subclassing without override raises with a clear message, opt-out emits no `$id`, and the parent class is not mutated by subclass operations.

Held items (auto-detect asymmetry, runtime-only invariants) — rationale on both is reasonable.

v0.2.0 bundling FILTER-441 / -442 / -444 / -452 into one release roll-up makes sense; no breaking changes, additive only.

One optional doc nit inline. LGTM otherwise.

PR #90, opened 2026-05-08
refactor(observability): infer OTLP TLS from endpoint scheme

stwilt:
## 📋 What does this PR do?

Replaces the always-`insecure=True` default for the OTLP gRPC span exporter and metrics exporter with TLS inferred from the endpoint scheme via `urllib.parse.urlparse`:

- `http://...` → `insecure=True` (plaintext)
- `https://...` → `insecure=False` (TLS)
- bare `host:port` → `insecure=False` (TLS — secure default, matches the OTel SDK)
- Explicit `insecure=` in `exporter_config` / kwargs always wins

Also drops the broken `OTLP_GRPC_ENDPOINT_SECURITY` env-var lookup in the metrics factory, which never bool-parsed strings and so was effectively `insecure=True` regardless of the env var's value.

## 🔍 Why is this needed?

The previous default would silently send telemetry over plaintext gRPC even when the configured endpoint was an `https://` URL — a quiet security footgun. Inferring from the URL scheme matches what most OTel SDKs (and the Go exporters this code originally mirrored) do, and gives `https://` and bare-host endpoints a secure-by-default behavior. The default `http://localhost:4317` still infers `insecure=True`, so local collectors keep working with no config change.

## 🧪 How was it tested?

- `pytest tests/test_tracing.py` — 24/24 pass
- New cases cover `http://`, `https://`, uppercase scheme, bare `host:port`, empty/None, and explicit `insecure=True` override against a bare-host endpoint
- `python -c "from openfilter.filter_runtime.open_telemetry.open_telemetry_exporter_factory import ExporterFactory"` — imports clean

## 🔗 Related Issues

None — no ticket for this change.

## ✅ Checklist

- [x] I have added or updated **documentation** as needed (no user-facing docs touched — internal exporter default)


openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -20,11 +22,16 @@ def build(exporter_type: str, **kwargs) -> MetricExporter:
 
         elif exporter_type == "otlp_grpc":
             try:
-
-                return OTLPGrpcExporter(
-                    endpoint=kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT"),
-                    insecure=kwargs.get("insecure", os.getenv("OTLP_GRPC_ENDPOINT_SECURITY",True))
-                )
+                endpoint = kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
```
lucasmundim: Behavior change risk for the metrics path: unlike `build_span_exporter` in `tracing.py`, this branch has no `"http://localhost:4317"` fallback. When neither kwarg nor env var is set, `endpoint` is `None`, `infer_otlp_insecure(None)` returns `False`, and the exporter falls back to its own default endpoint with **TLS** — which will break against a plaintext local collector that previously worked.

Before: `endpoint=None`, `insecure=True` (truthy default in old code) → plaintext OK.
After: `endpoint=None`, `insecure=False` → TLS handshake against a plaintext collector.

Suggest mirroring the tracing factory:

```python
endpoint = (
    kwargs.get("endpoint")
    or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
    or "http://localhost:4317"
)
```

That keeps the local-dev experience unchanged while preserving the security improvement for `https://`.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -20,11 +22,16 @@ def build(exporter_type: str, **kwargs) -> MetricExporter:
...
+                # Infer TLS from the endpoint scheme (http://=plaintext,
+                # https://=TLS, bare host:port=TLS). Explicit insecure=
+                # always wins. Replaces the prior OTLP_GRPC_ENDPOINT_SECURITY
+                # env-var lookup, which never parsed strings to bool and so
+                # was effectively always insecure=True.
+                insecure = kwargs.get("insecure")
+                if insecure is None:
+                    insecure = infer_otlp_insecure(endpoint)
+                return OTLPGrpcExporter(endpoint=endpoint, insecure=insecure)
```
lucasmundim, line 49: No tests exercise the metrics factory path here, even though this is where the most behavior-impactful change lives (the broken `OTLP_GRPC_ENDPOINT_SECURITY` env var was on this branch). All new tests are in `test_tracing.py`. Consider adding a small `tests/test_open_telemetry_exporter_factory.py` covering: `http://` endpoint, `https://` endpoint, explicit `insecure=True` against a bare host, and the unset-env-var path (which would surface the issue in my other comment as a failing test).

tests/test_tracing.py
```diff
@@ -52,20 +53,40 @@ def test_otlp_grpc_default_endpoint(self):
...
         with patch.dict("os.environ", {}, clear=True):
             exporter = build_span_exporter("otlp_grpc")
-        # _client is the OTLP exporter's underlying TraceServiceStub; the
-        # private _channel attribute is what we actually care about, but the
-        # public surface that survives upgrades is hard to come by. Instead
-        # we re-call build_span_exporter with insecure=False and confirm a
-        # different exporter object comes back, demonstrating the knob works.
-        secure = build_span_exporter("otlp_grpc", insecure=False)
-        self.assertIsNotNone(secure)
+        self.assertIsNotNone(exporter)
```
lucasmundim: This assertion is weaker than the one it replaces. The old `test_otlp_grpc_defaults_to_insecure` at least documented intent to inspect channel credentials; this one only checks the call doesn't return `None`, which won't catch a regression where the default scheme stops inferring `insecure=True`.

Since `test_infer_otlp_insecure_from_scheme` already covers the helper directly, this test is mostly a smoke check. Either tighten it (introspect the underlying gRPC channel for `_credentials`) or drop it as redundant — `assertIsNotNone` adds little here.

openfilter/observability/tracing.py
```diff
@@ -42,6 +43,19 @@ def force_flush(self, timeout_millis=30000):
         return True
 
 
+def infer_otlp_insecure(endpoint: Optional[str]) -> bool:
+    """Infer the OTLP gRPC ``insecure`` flag from the endpoint URL scheme.
+
+    ``http://`` is plaintext; ``https://`` is TLS; bare ``host:port`` is
+    treated as TLS (secure default, matches the OTel SDK). Uses
+    :func:`urllib.parse.urlparse` so the scheme comparison handles case
+    normalization without pattern matching.
```
lucasmundim: Worth one line on the `None` / empty-string contract — currently it's only inferable from the `if not endpoint: return False` body. Something like: "Returns `False` (TLS) for `None` or empty input, since the OTel SDK falls back to its own default endpoint and we should err on the side of TLS."

openfilter/observability/tracing.py
```diff
@@ -42,6 +43,19 @@ def force_flush(self, timeout_millis=30000):
...
+    """
+    if not endpoint:
+        return False
+    return urlparse(endpoint).scheme == "http"
```
lucasmundim: Nice — using `urlparse` here means case normalization comes for free (the test for `HTTP://` proves it). One thing worth confirming: `urlparse("collector.monitoring:4317").scheme` returns `"collector.monitoring"` on some Python versions because of how the `host:port` form gets parsed as a `scheme:path`. That happens to still work for the `== "http"` comparison, but it's a subtle gotcha — the test at line 64 covers it, so this is just a note for future readers.

lucasmundim (commented):
Solid security fix and the bug call-out on the env-var lookup is correct. A few items before merging — flagged inline. Two non-inline items not in the diff that also want attention:

- `RELEASE.md:352` and `docs/overview.md:935-952` still document `OTLP_GRPC_ENDPOINT_SECURITY` / don't mention the new scheme-based inference. Worth a docs pass in this PR or a follow-up.

shingonoide (commented):
LGTM on the security direction; +1 to Mundão's metrics-factory regression — that's the merge blocker.

Two complementary notes:

- **`OTLP_GRPC_ENDPOINT_SECURITY` was a full no-op, not a half-working knob — worth being precise in `RELEASE.md` beyond just deleting the stale doc.** `os.getenv(name, True)` returns the literal `True` only when unset; any *set* value came back as a string, and every non-empty string is truthy, so `insecure=True` regardless of what an operator wrote. No one was ever getting TLS through this var. The actual behavior change is narrower: operators with bare `host:port` endpoints flip from plaintext to TLS. That's the line worth in the changelog.
- **New hard import coupling.** `from openfilter.observability.tracing import infer_otlp_insecure` at the top of the metrics factory means the metrics module now fails to import if `tracing` does. Low-probability but easy to avoid with a lazy import inside the `otlp_grpc` branch, or by moving the helper to a small `openfilter/observability/_otlp.py` so neither side imports the other's heavy module.

Otherwise clean — `urlparse` over regex is the right call, and the gRPC path now infers transport the same way `otlp_http` already does, which is good consistency.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
stwilt: Good catch — fixed in b9405dd by mirroring `build_span_exporter`'s explicit `or "http://localhost:4317"` fallback. Without it, `endpoint=None` reached the SDK and combined with the inferred `insecure=False` to silently break plaintext local collectors. Added [`tests/test_open_telemetry_exporter_factory.py` L27](https://github.com/PlainsightAI/openfilter/blob/b9405dd99ba9cc2ba787c34a9024db98c71e717e/tests/test_open_telemetry_exporter_factory.py#L27) as a regression guard.

stwilt:
Updated in b9405dd — see inline thread replies for the per-comment items.

**@lucasmundim** — five inline points addressed; on the docs you flagged: the dead `OTLP_GRPC_ENDPOINT_SECURITY` line under v0.1.6 of `RELEASE.md` is gone, and the new `v0.1.31` entry incorporates @shingonoide's precision wording. `docs/overview.md` got a TLS-inference note + an annotation on the gRPC endpoint row in the env-var table.

**@shingonoide** — both points taken:
- *Changelog precision*: the v0.1.31 entry now spells out that `os.getenv(name, True)` only ever returned the literal `True` when unset, every set value came back as a truthy string, and the actual operator-visible change is narrow (bare `host:port` endpoints flip from plaintext to TLS).
- *Import decoupling*: helper moved to `openfilter/observability/_otlp.py`; both factories import from there. `tracing.py` re-exports `infer_otlp_insecure` for backward compat with any existing importers (the test file was the only in-tree caller).

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
stwilt: Done — added tests/test_open_telemetry_exporter_factory.py covering the four cases you listed: `http://`, `https://`, explicit `insecure=True` against a bare host, and the unset-everything path that surfaces the issue from your other comment as a regression guard.

tests/test_tracing.py
stwilt: Dropped. [`tests/test_tracing.py` L56](https://github.com/PlainsightAI/openfilter/blob/b9405dd99ba9cc2ba787c34a9024db98c71e717e/tests/test_tracing.py#L56) (`test_infer_otlp_insecure_from_scheme`) already pins the helper's contract directly so the wrapper smoke check wasn't earning its keep.

openfilter/observability/tracing.py
stwilt: Folded into the helper's docstring at [`openfilter/observability/_otlp.py` L19](https://github.com/PlainsightAI/openfilter/blob/b9405dd99ba9cc2ba787c34a9024db98c71e717e/openfilter/observability/_otlp.py#L19) (the helper moved per @shingonoide's decoupling comment): the `None` / empty branch now reads "when the caller has no endpoint to give us, the OTel SDK falls back to its own default endpoint and we err on the side of TLS rather than silently downgrading to plaintext."

openfilter/observability/tracing.py
stwilt: Acknowledged, leaving the inference as-is per your "just a note for future readers" framing. The `collector.monitoring:4317` case is covered explicitly in [`tests/test_tracing.py` L56](https://github.com/PlainsightAI/openfilter/blob/b9405dd99ba9cc2ba787c34a9024db98c71e717e/tests/test_tracing.py#L56) so a future urllib change would surface there.

lucasmundim (dismissed):
All five inline points addressed and the two follow-ups from @shingonoide too. Verified locally:

- Metrics-factory `or "http://localhost:4317"` fallback lands correctly — with empty env and no kwargs, the built exporter opens an **insecure** channel against `localhost:4317`. The original regression I flagged would have produced a secure channel against the SDK's default endpoint here.
- End-to-end TLS behavior verified by intercepting `grpc.{secure,insecure}_channel`: `http://` → insecure, `https://` → TLS, bare `host:port` → TLS, and explicit `insecure=True` against bare host correctly overrides to insecure.
- 27/27 tests pass across `tests/test_tracing.py` and the new `tests/test_open_telemetry_exporter_factory.py`.
- Decoupling clean: `openfilter/observability/_otlp.py` is the single source, metrics factory imports from there (no `tracing` import), and `tracing.py` re-exports `infer_otlp_insecure` in `__all__` for any external callers.
- `RELEASE.md` v0.1.31 entry is precise about the actual behavior change (only bare `host:port` operators flip plaintext → TLS), and the stale v0.1.6 line is gone. `docs/overview.md` got the TLS-inference callout.

One minor observation, not blocking: `https://...` + explicit `insecure=True` ends up as a secure channel — the underlying OTel SDK preserves TLS when the URL scheme is `https://` regardless of the kwarg. The factory passes the kwarg through faithfully; this is SDK behavior, not a PR bug, and the realistic bare-host override case works correctly.

LGTM. Thanks for the careful follow-through on every point.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -20,11 +22,24 @@ def build(exporter_type: str, **kwargs) -> MetricExporter:
...
-                    endpoint=kwargs.get("endpoint") or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT"),
-                    insecure=kwargs.get("insecure", os.getenv("OTLP_GRPC_ENDPOINT_SECURITY",True))
+                # Mirror the tracing factory's localhost fallback so an unset
+                # endpoint stays plaintext-against-localhost rather than
+                # falling through to the SDK's TLS default and breaking local
+                # `docker run otel/opentelemetry-collector` setups.
+                endpoint = (
+                    kwargs.get("endpoint")
+                    or os.getenv("OTEL_EXPORTER_OTLP_GRPC_ENDPOINT")
+                    or "http://localhost:4317"
```
lucasmundim: The comment "Mirror the tracing factory's localhost fallback" implies behavioral parity with `tracing.py`, but only the literal fallback string is shared — the env-var chain is not. `tracing.py:build_span_exporter` resolves the endpoint through:

`config kwarg → TELEMETRY_EXPORTER_OTLP_ENDPOINT → OTEL_EXPORTER_OTLP_ENDPOINT → OTEL_EXPORTER_OTLP_GRPC_ENDPOINT → http://localhost:4317`

(see https://github.com/PlainsightAI/openfilter/blob/610df4d96e70a131241c3ec71839135ab21b62e2/openfilter/observability/tracing.py#L80-L86)

This factory only checks `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT`. So an operator who sets `TELEMETRY_EXPORTER_OTLP_ENDPOINT` (the Plainsight convention used by `tracing.py` and `client.py:build_exporter`) will get traces exported correctly but metrics will silently fall through to `http://localhost:4317`.

The tracing factory's own docstring already flags this as "a latent inconsistency worth aligning in a follow-up on the metrics side" (https://github.com/PlainsightAI/openfilter/blob/610df4d96e70a131241c3ec71839135ab21b62e2/openfilter/observability/tracing.py#L52-L57). Two options here: (a) align the chain in this PR (a 3-line change), or (b) reword the comment to make the divergence explicit so the next reader doesn't have to grep the tracing factory to learn it.

lucasmundim (commented):
### Code review

Found 1 issue:

1. The "Mirror the tracing factory's localhost fallback" comment overstates parity — only the fallback is mirrored, not the env-var precedence chain.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
stwilt: Done in [`3cecd64`](https://github.com/PlainsightAI/openfilter/pull/90/commits/3cecd645). The metrics factory now walks the same 4-step chain as `build_span_exporter`, on both transports:

- [`otlp_grpc` endpoint resolution L34-L40](https://github.com/PlainsightAI/openfilter/blob/3cecd645/openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py#L34-L40) — `kwarg → TELEMETRY_EXPORTER_OTLP_ENDPOINT → OTEL_EXPORTER_OTLP_ENDPOINT → OTEL_EXPORTER_OTLP_GRPC_ENDPOINT → http://localhost:4317`
- [`otlp_http` endpoint resolution L55-L60](https://github.com/PlainsightAI/openfilter/blob/3cecd645/openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py#L55-L60) — same chain, terminating at the HTTP-specific var instead of the gRPC one (no localhost fallback since the SDK's default is `http://localhost:4318`)
- [Replaced the "Mirror the tracing factory's localhost fallback" comment](https://github.com/PlainsightAI/openfilter/blob/3cecd645/openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py#L25-L33) with the explicit precedence list to match `build_span_exporter`'s style
- [Dropped the latent-inconsistency paragraph](https://github.com/PlainsightAI/openfilter/blob/3cecd645/openfilter/observability/tracing.py#L47-L51) from `build_span_exporter`'s docstring — the env-var chain is no longer asymmetric across factories. (The `client.py:build_exporter` demo path's separate type-name asymmetry — `"otlp"`-only vs `otlp_grpc`/`otlp_http` — wasn't in scope here; flag if you want it on the same follow-up.)
- New tests: [`test_telemetry_exporter_otlp_endpoint_takes_precedence`](https://github.com/PlainsightAI/openfilter/blob/3cecd645/tests/test_open_telemetry_exporter_factory.py#L52-L64) and [`test_otel_exporter_otlp_endpoint_used_when_telemetry_var_unset`](https://github.com/PlainsightAI/openfilter/blob/3cecd645/tests/test_open_telemetry_exporter_factory.py#L66-L75). 6/6 pass.

tests/test_open_telemetry_exporter_factory.py
```diff
@@ -0,0 +1,79 @@
...
+    def test_otel_exporter_otlp_endpoint_used_when_telemetry_var_unset(self):
+        # When the Plainsight-convention var is unset, the standard OTel
+        # spec var should take precedence over the legacy GRPC-specific one.
+        env = {
+            "OTEL_EXPORTER_OTLP_ENDPOINT": "https://otel.example:4317",
+            "OTEL_EXPORTER_OTLP_GRPC_ENDPOINT": "https://grpc.example:4317",
+        }
+        with patch.dict("os.environ", env, clear=True):
+            exporter = ExporterFactory.build("otlp_grpc")
+        self.assertIsInstance(exporter, OTLPGrpcExporter)
```
lucasmundim: Both new tests assert only `assertIsInstance(exporter, OTLPGrpcExporter)`. The exporter class is `OTLPGrpcExporter` regardless of which env var wins, so these assertions would pass identically if the precedence were reversed — the test names make a specific claim (`_takes_precedence`, `_used_when_telemetry_var_unset`) that the body doesn't actually verify.

This is essentially the same shape as the old `assertIsNotNone` smoke check that got dropped in the prior round for not earning its keep.

Concrete suggestion — patch the exporter constructor in the factory module and assert the call:

```python
with patch.dict("os.environ", env, clear=True), \
     patch(
         "openfilter.filter_runtime.open_telemetry."
         "open_telemetry_exporter_factory.OTLPGrpcExporter"
     ) as mock_exporter:
    ExporterFactory.build("otlp_grpc")
    mock_exporter.assert_called_once_with(
        endpoint="https://plainsight.example:4317",
        insecure=False,
    )
```

Same pattern would tighten `test_unset_endpoint_falls_back_to_localhost_plaintext` (L43-50) to actually pin the `http://localhost:4317` literal, and the earlier `http://` / `https://` / `insecure=True` cases to confirm what was passed to the exporter.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
```diff
@@ -20,18 +22,44 @@ def build(exporter_type: str, **kwargs) -> MetricExporter:
...
                 logging.error(f"Failed to set OTLP_GRPC exporter {e}")
 
         elif exporter_type == "otlp_http":
             try:
+                endpoint = (
+                    kwargs.get("endpoint")
+                    or os.getenv("TELEMETRY_EXPORTER_OTLP_ENDPOINT")
+                    or os.getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+                    or os.getenv("OTEL_EXPORTER_OTLP_HTTP_ENDPOINT")
+                )
```
lucasmundim, line 65: The comment at line 25 says precedence "mirrors the tracing factory." The `otlp_grpc` branch above does, including the `"http://localhost:4317"` fallback. This `otlp_http` branch matches the env-var ordering but drops the `"http://localhost:4318"` fallback that `build_span_exporter` has at `tracing.py:104`.

Not a behavior bug — the OTel HTTP exporter has its own endpoint default when passed `endpoint=None` — but the parity claim in the gRPC-branch comment is louder than what this branch delivers. Either add the `or "http://localhost:4318"` fallback for true symmetry, or drop a one-line note here that this branch intentionally defers to the SDK default for the unset case.

lucasmundim (commented):
Re-review on 3cecd64.

The env-var precedence chain is the right shape and the comment block at lines 25-33 clearly documents the ordering. Two items inline — neither blocking, but the test one is the same shape as the `assertIsNotNone` weakness we dropped in the prior round and worth tightening before this merges.

tests/test_open_telemetry_exporter_factory.py
stwilt: Good call — fixed in `20c0c65`. Every test now patches the exporter constructor in the factory module's namespace ([`tests/test_open_telemetry_exporter_factory.py` L21-L23](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L21-L23)) and asserts the actual call args, so the test names match what's being verified.

- `test_unset_endpoint_falls_back_to_localhost_plaintext` now pins the literal at [`tests/test_open_telemetry_exporter_factory.py` L65-L67](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L65-L67) — `assert_called_once_with(endpoint="http://localhost:4317", insecure=True)` — so any regression that lets `endpoint=None` (or TLS) reach the SDK fails loudly.
- `test_telemetry_exporter_otlp_endpoint_takes_precedence` ([`tests/test_open_telemetry_exporter_factory.py` L81-L83](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L81-L83)) and `test_otel_exporter_otlp_endpoint_used_when_telemetry_var_unset` ([`tests/test_open_telemetry_exporter_factory.py` L94-L96](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L94-L96)) pin the winning endpoint string + inferred `insecure` value, so reversing the precedence would now make these fail.
- The earlier http:// / https:// / explicit-insecure cases are pinned the same way at [`tests/test_open_telemetry_exporter_factory.py` L31-L33](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L31-L33), [`tests/test_open_telemetry_exporter_factory.py` L42-L44](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L42-L44), and [`tests/test_open_telemetry_exporter_factory.py` L54-L56](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/tests/test_open_telemetry_exporter_factory.py#L54-L56).

6/6 still pass.

openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py
stwilt: Went with the one-line note option in `20c0c65` rather than the literal fallback — the OTLP HTTP SDK already defaults `endpoint=None` to `http://localhost:4318/v1/metrics`, and a bare-host `or "http://localhost:4318"` fallback in our chain would silently drop the `/v1/metrics` path suffix. So the asymmetry vs the gRPC branch is intentional, and [`openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py` L55-L60](https://github.com/PlainsightAI/openfilter/blob/20c0c652bf853a0ac39f7be93370dce8d781cf61/openfilter/filter_runtime/open_telemetry/open_telemetry_exporter_factory.py#L55-L60) now spells that out inline so the next reader doesn't have to compare against `build_span_exporter` to figure out why.

lucasmundim (approved):
**Re-review on `20c0c65`.**

Both inline items from the prior round are addressed:

- **Test assertions are now load-bearing.** Every case patches `OTLPGrpcExporter` in the factory module's namespace and asserts the exact `(endpoint, insecure)` kwargs. `test_unset_endpoint_falls_back_to_localhost_plaintext` pins `("http://localhost:4317", True)`; the two precedence tests pin the winning endpoint and inferred `insecure`. Reversing precedence or letting `endpoint=None` reach the SDK would now fail loudly.
- **`otlp_http` asymmetry documented.** The new comment at L55-60 makes the intentional divergence vs. the gRPC branch explicit, and the SDK-default claim checks out — `OTLPMetricExporter(endpoint=None)._endpoint == "http://localhost:4318/v1/metrics"`, so a bare-host fallback would have silently dropped the path suffix.

29/29 tests pass locally. LGTM.

One non-blocking follow-up worth noting: the `otlp_http` branch grew the same env-var precedence chain but the new test file only covers `otlp_grpc`. Reasonable scope cut for this PR; if symmetric coverage is wanted later, the existing tests are a cheap template.

openfilter/observability/client.py
```diff
@@ -69,7 +70,15 @@ def shutdown(self, timeout: float = 30000) -> None:
             or os.getenv("TELEMETRY_EXPORTER_OTLP_ENDPOINT")
```
lucasmundim, line 70: Pre-existing, not introduced here, but worth a follow-up: this env-var chain is shorter than the other two builders we aligned in 3cecd64. `ExporterFactory.otlp_grpc` and `tracing.build_span_exporter.otlp_grpc` both also consult `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_GRPC_ENDPOINT` before falling back to localhost. An operator setting only the standard OTel env vars would get them honored by traces and by `filter_runtime`'s metrics factory, but not by this builder. Out of scope for this PR (the diff doesn't touch the chain) — flagging for a follow-up that fans out the same 4-step chain here.

RELEASE.md
```diff
@@ -2,6 +2,16 @@
 
 OpenFilter Library release notes
 
+## v0.2.1 - 2026-05-09
+
+### Changed
+
+- **OTLP gRPC `insecure` flag now inferred from endpoint scheme**: The exporter factory and tracing builder now use `urllib.parse.urlparse` on the configured endpoint — `http://` infers plaintext, `https://` infers TLS, and bare `host:port` infers TLS (secure default, matches the OTel SDK). Explicit `insecure=` in `exporter_config` / kwargs always wins. The default `http://localhost:4317` endpoint still infers `insecure=True`, so local collectors keep working with no config change. The metrics factory's missing localhost fallback was also fixed in this change — an unset endpoint no longer falls through to the SDK's TLS default against a plaintext local collector.
```
lucasmundim, line 9: Worth a small tightening: this entry says "the exporter factory and tracing builder," but 440a52f adds the same inference to a third site — `observability/client.py:build_exporter` — which is the metrics path `OpenTelemetryClient` uses in the live runtime and the one that produced the `WRONG_VERSION_NUMBER` symptom operators would search the changelog for. Suggest something like "the exporter factory, tracing builder, and `OpenTelemetryClient`'s metrics builder now use `urllib.parse.urlparse`..." so the changelog matches symptom to fix.

lucasmundim (approved):
Re-review on 440a52f. The new commit closes the third call site — client.py:build_exporter is the metrics builder OpenTelemetryClient wires up in the live runtime (line 157), so this is the path that was producing the WRONG_VERSION_NUMBER handshake error in filter pods against the in-cluster collector.

35/35 tests pass locally across the three relevant files. The new test file pins exact (endpoint, insecure) kwargs — same load-bearing shape as test_open_telemetry_exporter_factory.py from the prior round.

LGTM. Two non-blocking notes inline, both about scope-symmetry vs. the other two builders, neither introduced by this PR.

PR #93, opened 2026-05-15
feat(cascade): widen `<` / `<=` upper bounds that exclude OF_VERSION

stwilt:
## 📋 What does this PR do?

Augments the openfilter → filter-* cascade so it can rewrite a consumer's pyproject upper bound when the new `OF_VERSION` exceeds it, instead of silently skipping the filter as a no-op.

Three coordinated changes in `scripts/cascade/`:

- **`bump-strategy.sh`** (the embedded `_bump.py` helper): adds upper-bound widening to `rewrite_requirement()`. Rule: `target.major == 0` → `<target.major>.<target.minor + 1>.0` (next minor, mirrors the org's existing "track minor" 0.X pin pattern); `target.major >= 1` → `<target.major + 1>.0.0` (standard SemVer next-major). `!=` exclusions, `>` lower bounds, clause order, surrounding whitespace, and environment markers are still preserved verbatim.
- **`check_constraint.py`**: emits a new `widen:<spec>` state for constraints whose only blocking clauses are widenable upper bounds (`<` / `<=`). `skip:<spec>` is now reserved for non-widenable blocks: lower-bound exclusions (downgrades) and explicit `!=` exclusions. Splitting `widen:` from `skip:` is what lets discover.sh consider the filter eligible while keeping a hard boundary against silent semantic regressions.
- **`discover.sh`**: treats `widen:` as eligible with a separate diagnostic line ("eligible — pin `<spec>` will be widened for `OF_VERSION`") so operators can tell at a glance that this cascade PR is more than a routine lower-bound bump.

## 🔍 Why is this needed?

Every `filter-*` repo in the org currently pins `openfilter[all]>=0.1.30,<0.2.0`. Under the pre-existing cascade behavior, a v0.2.0 (or v1.0.0) release would cascade to **zero** of 55 filters — the cascade refused to rewrite upper bounds, so it could only handle 0.1.x → 0.1.y bumps within the existing range. Without this augmentation, every non-patch openfilter release requires manually fanning out a constraint-widening PR across ~55 repos before the cascade can do anything useful.

Smoke test against the org-canonical case (`<0.2.0,>=0.1.30` → `OF_VERSION=0.2.0`) before this change:

```
filter-sam3-detector: skip (constraint <0.2.0,>=0.1.30 excludes 0.2.0)
Discovery complete: 0 eligible, 1 skipped
```

After this change:

```
filter-sam3-detector: eligible — pin <0.2.0,>=0.1.30 will be widened for 0.2.0
Discovery complete: 1 eligible, 0 skipped
[…]
pyproject.toml: rewrote 1 openfilter pin(s) → 0.2.0
```

Humans still review every cascade PR; widening is automation-floor behavior. The bot proposes the new bound; the reviewer can adjust it before merge if a non-default widening rule is wanted for that filter.

## 🧪 How was it tested?

**Unit tests** (`tests/test_cascade_bump_strategy.py`): 7 new tests covering the org-canonical 0.X → 0.Y widening, 1.0+ widening (next-major rule), `<=` widening, no-op-when-upper-bound-already-admits-target, inline-comment preservation through tomlkit, the `requirements.txt` rewrite path, and idempotency on re-run. All 16 tests in the file pass (9 pre-existing + 7 new).

```
$ pytest tests/test_cascade_bump_strategy.py -v
…
============================== 16 passed in 0.95s ==============================
```

**`check_constraint.py` manual validation** against the five representative cases (`>=0.1.30,<0.2.0` pin):

| `OF_VERSION` | Expected | Got |
|---|---|---|
| `0.1.30`           | `ok:`    | `ok:<0.2.0,>=0.1.30` |
| `0.2.0`            | `widen:` | `widen:<0.2.0,>=0.1.30` |
| `0.1.0`            | `skip:`  | `skip:<0.2.0,>=0.1.30` (downgrade) |
| `0.2.0` + `!=0.2.0` | `skip:` | `skip:!=0.2.0,<1.0,>=0.1.0` |
| no openfilter dep  | `none`   | `none` |

**End-to-end smoke test** via `gh workflow run cascade-on-tag.yaml --ref feat/dt-145-cascade-widen-constraint -f dry_run=true -f single_filter=filter-sam3-detector` (run
[25894063159](https://github.com/PlainsightAI/openfilter/actions/runs/25894063159)): discover correctly classified sam3-detector as `widen:`, bump-strategy cloned the consumer, ran the rewriter, and reported `pyproject.toml: rewrote 1 openfilter pin(s) → 0.2.0`. `open-mechanical-pr` was correctly skipped per `dry_run=true`.

## 🔗 Related Issues

[DT-145](https://plainsight-ai.atlassian.net/browse/DT-145) — cascade infrastructure.

## ✅ Checklist

- [x] I have added or updated **documentation** as needed (in-tree docstrings on `discover.sh`, `check_constraint.py`, and `_bump.py`)

[DT-145]: https://plainsight-ai.atlassian.net/browse/DT-145?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/scripts/cascade/bump-strategy.sh
+++ b/scripts/cascade/bump-strategy.sh
@@ -14,18 +14,30 @@ set -euo pipefail
 BUMP_HELPER_DIR=$(mktemp -d)
 trap 'rm -rf "${BUMP_HELPER_DIR}"' EXIT
 cat > "${BUMP_HELPER_DIR}/_bump.py" <<'HELPER'
-"""rewrite_requirement: bump ==/>=/~= clauses in a PEP 508 requirement.
+"""rewrite_requirement: bump ==/>=/~= clauses in a PEP 508 requirement,
+and widen `<` / `<=` upper bounds that would otherwise exclude OF_VERSION.
 
-Lower-bound clauses go to OF_VERSION; upper-bound clauses (<, <=, !=),
-clause order, surrounding whitespace, and any environment marker are
-preserved verbatim.
+Lower-bound clauses (==, >=, ~=) go to OF_VERSION. Upper-bound clauses
+(<, <=) are widened to admit OF_VERSION when they currently exclude it;
+otherwise they stay verbatim. != exclusions, `>` lower bounds, clause
+order, surrounding whitespace, and any environment marker are preserved
+verbatim regardless.
+
+Widening rule for the new upper bound:
+    * target.major == 0 (pre-1.0 / unstable): `<target.major>.<target.minor + 1>.0`
+      (e.g. 0.2.0 -> 0.3.0). Matches the org's existing "track next minor"
+      pin pattern for 0.X filters.
+    * target.major >= 1: `<target.major + 1>.0.0` (e.g. 1.0.0 -> 2.0.0).
+      Standard SemVer "track next major" for post-1.0 filters.
 """
 from __future__ import annotations
 
 from packaging.requirements import Requirement
 from packaging.specifiers import Specifier
+from packaging.version import Version
 
 _BUMPABLE_OPS = ("==", ">=", "~=")
+_UPPER_BOUND_OPS = ("<", "<=")
 
 
 def _specifier_span(text: str, req: Requirement) -> tuple[int, int]:
@@ -54,10 +66,20 @@ def _specifier_span(text: str, req: Requirement) -> tuple[int, int]:
     return spec_start, spec_end
 
 
+def next_upper_bound(target: Version) -> str:
+    """Compute the widened upper bound to admit `target` and a reasonable
+    forward range. See module docstring for the rule.
+    """
+    if target.major == 0:
+        return f"0.{target.minor + 1}.0"
+    return f"{target.major + 1}.0.0"
+
+
 def rewrite_requirement(old: str, of_version: str) -> str:
     req = Requirement(old)
     if not req.specifier:
         return old
+    target = Version(of_version)
     spec_start, spec_end = _specifier_span(old, req)
     spec_text = old[spec_start:spec_end]
     new_clauses: list[str] = []
@@ -68,13 +90,25 @@ def rewrite_requirement(old: str, of_version: str) -> str:
             new_clauses.append(raw)
             continue
         s = Specifier(stripped)
+        leading = raw[: len(raw) - len(raw.lstrip())]
+        trailing = raw[len(raw.rstrip()) :]
         if s.operator in _BUMPABLE_OPS:
-            leading = raw[: len(raw) - len(raw.lstrip())]
-            trailing = raw[len(raw.rstrip()) :]
             new_clauses.append(f"{leading}{s.operator}{of_version}{trailing}")
             bumped = True
-        else:
-            new_clauses.append(raw)
+            continue
+        if s.operator in _UPPER_BOUND_OPS:
+            bound = Version(s.version)
+            needs_widen = (
+                (s.operator == "<" and target >= bound)
+                or (s.operator == "<=" and target > bound)
+            )
+            if needs_widen:
+                new_clauses.append(
+                    f"{leading}<{next_upper_bound(target)}{trailing}"
+                )
+                bumped = True
+                continue
+        new_clauses.append(raw)
     if not bumped:
         return old
     return old[:spec_start] + ",".join(new_clauses) + old[spec_end:]
--- a/scripts/cascade/check_constraint.py
+++ b/scripts/cascade/check_constraint.py
@@ -12,10 +12,14 @@ silent "no openfilter dep" line.
 
 `widen:<spec>` — OF_VERSION is outside the current constraint and every
 blocking clause is one bump-strategy.sh's rewriter can rewrite in place
-(`<`, `<=`, `==`, `~=`, `>=`). `skip:<spec>` is reserved for blocks the
-rewriter can't handle: `!=X` exclusions and `>X` strict lower bounds.
+without downgrading: `<`, `<=`, `==`, `~=`. `skip:<spec>` covers the rest:
+`!=X` and `>X` (rewriter has no rule), and `>=X` blocking (rewriter has
+`>=` in `_BUMPABLE_OPS`, but a `>=X` clause only blocks when target < X,
+so applying that rule would downgrade the consumer's floor — different
+semantic category from widening, kept as a skip-class guardrail).
 Cascade PRs are human-reviewed before merge, so the split exists to filter
-mechanically-unhandled clauses, not to gate operator-intent decisions.
+mechanically-unhandled-or-unsafe clauses, not to gate operator-intent
+decisions.
 
 When openfilter appears in MULTIPLE PEP 621 tables (e.g. [project.dependencies]
 AND [project.optional-dependencies.gpu]), all matching specifiers are
@@ -36,7 +40,7 @@ def _is_widenable_block(combined: SpecifierSet, target: Version) -> bool:
     for s in combined:
         if target in SpecifierSet(str(s)):
             continue
-        if s.operator in ("<", "<=", "==", "~=", ">="):
+        if s.operator in ("<", "<=", "==", "~="):
             continue
         return False
     return True
--- a/scripts/cascade/check_constraint.py
+++ b/scripts/cascade/check_constraint.py
@@ -1,7 +1,7 @@
 """Check openfilter version constraint compatibility.
 
 Reads OF_VERSION from env, scans pyproject.toml for openfilter dependency.
-Outputs: none | none:poetry-format | ok:<spec> | skip:<spec> | error:<msg>
+Outputs: none | none:poetry-format | ok:<spec> | widen:<spec> | skip:<spec> | error:<msg>
 
 `none:poetry-format` is emitted when no openfilter dep was found in PEP 621
 [project.dependencies] / [project.optional-dependencies] but one IS present
@@ -9,6 +9,14 @@ under Poetry's [tool.poetry.dependencies] (or dev-dependencies). The cascade
 only knows how to rewrite PEP 621 pins, so Poetry consumers are still
 skipped — but the operator gets a distinguishable diagnostic instead of a
 silent "no openfilter dep" line.
+
+`widen:<spec>` means OF_VERSION is outside the current constraint, BUT the
+only blocking clauses are upper bounds (`<` / `<=`) that bump-strategy.sh
+can safely widen. `skip:<spec>` is reserved for blocks the cascade refuses
+to rewrite automatically: lower-bound exclusions (would be a downgrade) and
+explicit `!=` exclusions (deliberate gating). Splitting `widen` from `skip`
+lets the cascade discover this filter as eligible while keeping a hard
+boundary against silent semantic regressions.
 
 When openfilter appears in MULTIPLE PEP 621 tables (e.g. [project.dependencies]
 AND [project.optional-dependencies.gpu]), all matching specifiers are
@@ -20,8 +28,30 @@ the optional specifier rejects.
 import sys, os, tomllib
 
 from packaging.requirements import Requirement
-from packaging.specifiers import SpecifierSet
+from packaging.specifiers import Specifier, SpecifierSet
 from packaging.version import Version
+
+
+def _is_widenable_block(combined: SpecifierSet, target: Version) -> bool:
+    """True when every clause that excludes `target` is a widenable upper bound.
+
+    Widenable: `<X` or `<=X` — bump-strategy.sh's rewriter handles these
+    in place.
+
+    Not widenable: `>=X` / `>X` lower bounds (widening would be a downgrade),
+    `!=X` exclusions (deliberate gating), and `==X` / `~=X` pins (the
+    rewriter can bump these in place, but if they're still excluding
+    target after intersection it means another clause makes the situation
+    unsalvageable; operator should inspect).
+    """
+    for s in combined:
+        if target in SpecifierSet(str(s)):
+            continue
+        if s.operator in ("<", "<="):
+            continue
+        return False
+    return True
+
 
 try:
     if not os.path.exists("pyproject.toml"):
@@ -65,6 +95,11 @@ try:
         sys.exit(0)
     v = Version(v_str)
     spec_str = str(combined)
-    print(("ok:" if v in combined else "skip:") + spec_str)
+    if v in combined:
+        print("ok:" + spec_str)
+    elif _is_widenable_block(combined, v):
+        print("widen:" + spec_str)
+    else:
+        print("skip:" + spec_str)
 except Exception as e:
     print("error:" + str(e))
--- a/scripts/cascade/check_constraint.py
+++ b/scripts/cascade/check_constraint.py
@@ -10,13 +10,12 @@ only knows how to rewrite PEP 621 pins, so Poetry consumers are still
 skipped — but the operator gets a distinguishable diagnostic instead of a
 silent "no openfilter dep" line.
 
-`widen:<spec>` means OF_VERSION is outside the current constraint, BUT the
-only blocking clauses are upper bounds (`<` / `<=`) that bump-strategy.sh
-can safely widen. `skip:<spec>` is reserved for blocks the cascade refuses
-to rewrite automatically: lower-bound exclusions (would be a downgrade) and
-explicit `!=` exclusions (deliberate gating). Splitting `widen` from `skip`
-lets the cascade discover this filter as eligible while keeping a hard
-boundary against silent semantic regressions.
+`widen:<spec>` — OF_VERSION is outside the current constraint and every
+blocking clause is one bump-strategy.sh's rewriter can rewrite in place
+(`<`, `<=`, `==`, `~=`, `>=`). `skip:<spec>` is reserved for blocks the
+rewriter can't handle: `!=X` exclusions and `>X` strict lower bounds.
+Cascade PRs are human-reviewed before merge, so the split exists to filter
+mechanically-unhandled clauses, not to gate operator-intent decisions.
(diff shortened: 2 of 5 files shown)


scripts/cascade/bump-strategy.sh
```diff
@@ -68,13 +90,25 @@ def rewrite_requirement(old: str, of_version: str) -> str:
...
+            continue
+        if s.operator in _UPPER_BOUND_OPS:
+            bound = Version(s.version)
+            needs_widen = (
+                (s.operator == "<" and target >= bound)
+                or (s.operator == "<=" and target > bound)
+            )
+            if needs_widen:
+                new_clauses.append(
+                    f"{leading}<{next_upper_bound(target)}{trailing}"
```
lucasmundim, line 107: Operator normalization: the widened clause is always emitted as `<NEXT`, even when the input was `<=X`. Tested in `test_pyproject_widens_le_upper_bound` so the behavior is intentional, but the module docstring above doesn't call this out — a reader might expect `<=` to round-trip. Worth a one-liner in the `_bump.py` docstring noting that widening normalizes the operator to `<`.

scripts/cascade/check_constraint.py
```diff
@@ -20,9 +28,31 @@
...
+
+def _is_widenable_block(combined: SpecifierSet, target: Version) -> bool:
+    """True when every clause that excludes `target` is a widenable upper bound.
+
+    Widenable: `<X` or `<=X` — bump-strategy.sh's rewriter handles these
+    in place.
+
+    Not widenable: `>=X` / `>X` lower bounds (widening would be a downgrade),
+    `!=X` exclusions (deliberate gating), and `==X` / `~=X` pins (the
+    rewriter can bump these in place, but if they're still excluding
```
lucasmundim: Minor docstring nit: the reasoning here ("another clause makes the situation unsalvageable") only holds when there *is* another clause. A consumer pinned as just `openfilter~=0.1.30` upgrading to `0.3.0` lands here with no other clause — the rewriter would in fact handle it correctly via `_BUMPABLE_OPS`, but `_is_widenable_block` returns False, so it's classified as `skip:`.

Two options:
- Drop the misleading reasoning from the docstring and just say `~=` / `==` pins are skipped conservatively (operator inspection preferred).
- Or add `~=` to the widenable set so the rewriter's bumping behavior is exposed through discover.sh. I'd lean toward the first — `~=` pins are rare for `openfilter` in practice and conservative skip is safer.

scripts/cascade/discover.sh
```diff
@@ -170,7 +174,17 @@ while IFS= read -r REPO; do
...
+      ;;
+    widen:*)
+      # Eligible — target outside the current pin range but only by an
+      # upper bound (< / <=). bump-strategy.sh's rewriter will widen it
+      # using next_upper_bound() (next-minor-after-target for 0.X,
+      # next-major-after-target for 1.0+). Surfaced separately in
+      # diagnostics so the operator knows this PR is more than a
+      # lower-bound bump.
+      CONSTRAINT="${COMPAT#widen:}"
+      echo "  ${REPO}: eligible — pin ${CONSTRAINT} will be widened for ${OF_VERSION}" >&2
```
lucasmundim, line 189: Follow-up suggestion (not blocking): the current diagnostic says the pin *will* be widened but not *to what*. An operator reading the discovery log has to know the widening rule (next-minor for 0.X, next-major for 1.0+) to predict the result. Computing and displaying the proposed new bound here (e.g. `pin <0.2.0,>=0.1.30 will be widened to <0.3.0 for 0.2.0`) would make the discovery log self-describing.

Could reuse `next_upper_bound()` from `_bump.py`, though that means lifting the helper out of the heredoc — probably a separate refactor.

scripts/cascade/check_constraint.py
```diff
@@ -20,9 +28,31 @@
 import sys, os, tomllib
 
 from packaging.requirements import Requirement
-from packaging.specifiers import SpecifierSet
+from packaging.specifiers import Specifier, SpecifierSet
 from packaging.version import Version
 
+
+def _is_widenable_block(combined: SpecifierSet, target: Version) -> bool:
```
lucasmundim, line 38: Suggestion for a follow-up: `_is_widenable_block` is pure (takes a `SpecifierSet` + `Version`, returns bool) and would be easy to unit-test in isolation. The PR description has a 5-case manual validation table — lifting those into a small `tests/test_cascade_check_constraint.py` would lock in the contract and catch regressions if/when the widenable set grows (e.g. if `~=` gets added later per the other comment).

Not blocking — the smoke-test path via `bump-strategy.sh` covers the end-to-end behavior.

lucasmundim (commented):
Nice piece of work — the `widen:` / `skip:` split is the right design, the rewriter preserves what it should, and the 7 new tests cover the right axes (I reran them locally, all 16 pass).

A few non-blocking suggestions inline. Higher-signal items are the `~=X` docstring (it mentions a multi-clause scenario that doesn't apply to single `~=X` pins) and worth-considering follow-ups: unit tests for `_is_widenable_block`, and surfacing the proposed bound in the `discover.sh` diagnostic.

Also: the rewriter unconditionally emits `<` regardless of whether the input was `<` or `<=` — functionally correct and tested, but worth one line in `_bump.py`'s docstring noting the operator normalization.

Approving with the above as follow-up suggestions.

scripts/cascade/bump-strategy.sh
stwilt: Skipping the docstring tweak — keeping prose surface minimal in the cascade infra. Behavior is unchanged and covered by `test_pyproject_widens_le_upper_bound`.

scripts/cascade/check_constraint.py
stwilt: Right that the framing was wrong — but digging in turned up the policy itself was wrong too. Local survey across `~/work`: 9 of 20 openfilter consumers pin with `~=` (filter-global-position, filter-pose-estimation, filter-protege-model, filter-scene-caption, filter-template) or `==` (filter-event-sink, filter-sam3-detector-020, filter-quality-gates, filter-sweet-green-subject-data-aggregator). The conservative-skip was stranding ~45% of the fleet out of the cascade for no semantic reason — including filter-template, which seeds every new filter.

The rewriter already handles `~=` and `==` via `_BUMPABLE_OPS` (emits `~=target` / `==target`). The classifier was just refusing to dispatch them. Cascade PRs are human-reviewed before merge, so the "auto-rewrite without operator review" framing was a strawman — the review IS the operator review.

b559222 broadens `_is_widenable_block` ([`scripts/cascade/check_constraint.py` L34](https://github.com/PlainsightAI/openfilter/blob/b559222a631d76ecf9e4efb64f4d0dc5964287ec/scripts/cascade/check_constraint.py#L34)) to accept any operator the rewriter handles: `<`, `<=`, `==`, `~=`, `>=`. `skip:` is now reserved for `!=X` exclusions and `>X` strict lower bounds — clauses the rewriter literally can't touch. The "operator intent" framing is gone.

scripts/cascade/discover.sh
stwilt: Noted — leaving for a follow-up rather than expanding scope here.

scripts/cascade/check_constraint.py
stwilt: Done — [`tests/test_cascade_check_constraint.py` L1](https://github.com/PlainsightAI/openfilter/blob/b559222a631d76ecf9e4efb64f4d0dc5964287ec/tests/test_cascade_check_constraint.py#L1) in b559222 has 11 classifier tests covering the new dispatch boundary (`<` / `<=` / `~=` / `==` → `widen:`; `!=` / `>` → `skip:`; `ok:` and `none*` paths). Two companion rewriter tests in [`tests/test_cascade_bump_strategy.py` L269-L296](https://github.com/PlainsightAI/openfilter/blob/b559222a631d76ecf9e4efb64f4d0dc5964287ec/tests/test_cascade_bump_strategy.py#L269-L296) lock in the `~=` and `==` cascade-style point bumps end-to-end.

lucasmundim (commented):
Re-review covering b559222 and bbd011a.

bbd011a (Python <3.11 skipif) is fine — mirrors the existing pattern in `test_cascade_bump_strategy.py`, and the CI matrix does still include 3.10, so the gate is load-bearing.

b559222 (broaden `_is_widenable_block` to include `==`, `~=`, `>=`) is mostly good — adding `==` / `~=` is reasonable given the fleet-survey data and the human-review safety net. One concern on `>=` inclusion (inline), one nit on the test-file docstring attribution (inline), and the discover.sh module/inline comments are now stale w.r.t. the broader policy (separate inline).

scripts/cascade/check_constraint.py
```diff
@@ -20,9 +27,21 @@
...
+    """True when every clause excluding `target` is one bump-strategy.sh rewrites."""
+    for s in combined:
+        if target in SpecifierSet(str(s)):
+            continue
+        if s.operator in ("<", "<=", "==", "~=", ">="):
```
lucasmundim: `>=` in this set silently enables downgrade cascades. When `>=X` is the blocker, by definition `target < X` — i.e., OF_VERSION is older than the consumer's stated minimum. The rewriter has `>=` in `_BUMPABLE_OPS` and will emit `>=of_version`, turning a `>=0.5.0` pin into `>=0.3.0` if a 0.3.0 cascade fires.

The commit message frames the new skip-set as "clauses the rewriter literally can't touch" — but the rewriter touches `>=` *by downgrading*, which is a different category from `==` / `~=` (point bumps where the consumer never expressed a floor) and from `<` / `<=` (true widening). The previous policy explicitly cited "widening would be a downgrade" as the skip rationale for lower bounds; this commit removes that without acknowledging the trade-off.

Also note the asymmetry between `>X` (still skip) and `>=X` (now widen) — both represent the same operator intent ("at least version X required"). Worth either dropping `>=` from this set (keeping the downgrade guardrail) or adding a `test_ge_lower_bound_blocking_*` case to `test_cascade_check_constraint.py` that locks in whichever behavior is intentional. As-is, the only `>=` test is `test_target_already_admitted_is_ok`, which doesn't exercise the blocking path.

tests/test_cascade_check_constraint.py
```diff
@@ -0,0 +1,151 @@
...
+The classifier emits `ok:` / `widen:` / `skip:` / `none*` / `error:` against a
+consumer pyproject.toml in CWD. discover.sh reads this and decides whether to
+dispatch a bump PR. Anything classified `skip:` is stranded — the cascade
+ignores it — so the boundary between `widen:` and `skip:` is the load-bearing
+policy: which pin shapes get auto-cascaded vs. left to operators.
+
+Coverage:
+- `<X` / `<=X` upper bounds — widenable (historical behavior).
+- `~=X` / `==X` pins — widenable. Almost half the filter fleet uses `~=` per
+  org convention (Mundim's stylistic preference, no semantic intent claim);
```
lucasmundim: Nit: please drop the "Mundim's stylistic preference" parenthetical. Two reasons: (1) embedding a reviewer name in an OSS test docstring is unusual and the attribution will outlive the context, and (2) I don't actually have an expressed preference for `~=` — the rationale for accepting `~=` is the fleet-survey data in the b559222 commit message, which is sufficient on its own. A line like "almost half the filter fleet uses `~=` per org convention" without the attribution captures the same fact.

scripts/cascade/discover.sh
```diff
@@ -6,8 +6,12 @@
...
 #   2. No `.github/no-cascade-openfilter` opt-out marker
 #   3. Has a VERSION file at the default branch tip
-#   4. Declares an `openfilter` PEP 621 dependency whose constraint allows
-#      ${OF_VERSION} (delegated to scripts/cascade/check_constraint.py)
+#   4. Declares an `openfilter` PEP 621 dependency whose constraint either
+#      already permits ${OF_VERSION} (check_constraint.py emits `ok:`) or
+#      is only blocked by a widenable upper bound (`<` / `<=`) that
+#      bump-strategy.sh's rewriter will rewrite in place (emits `widen:`).
+#      Lower-bound exclusions (downgrades) and `!=` exclusions still
+#      `skip:`.
```
lucasmundim: Stale after b559222: this header still says "widenable upper bound (`<` / `<=`)" and explicitly that "Lower-bound exclusions (downgrades) and `!=` exclusions still `skip:`". After the broadening, `==` / `~=` / `>=` are also widenable, and (if `>=` stays — see other comment) lower-bound downgrades are no longer skipped. The same staleness exists in the inline `widen:*` case comment around lines 180–181 ("only by an upper bound (`<` / `<=`)"). Worth a one-pass update so the module docs match the policy.

scripts/cascade/check_constraint.py
stwilt: Right — the asymmetry was real. Took option (a) and dropped `>=` from `_is_widenable_block` so blocking floors stay `skip:`; the rewriter's `>=` rule no longer fires on the discovery path. Docstring at [`scripts/cascade/check_constraint.py` L13-L22](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/scripts/cascade/check_constraint.py#L13-L22) now spells out *why* `>=` is excluded despite being in `_BUMPABLE_OPS`. `test_ge_lower_bound_blocking_is_skip` at [`tests/test_cascade_check_constraint.py` L118-L127](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/tests/test_cascade_check_constraint.py#L118-L127) locks the guardrail in (`>=0.5.0` blocking 0.3.0 → `skip:`).

tests/test_cascade_check_constraint.py
stwilt: Done — parenthetical dropped at [`tests/test_cascade_check_constraint.py` L11-L12](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/tests/test_cascade_check_constraint.py#L11-L12). Extended the Coverage block at [`tests/test_cascade_check_constraint.py` L13-L16](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/tests/test_cascade_check_constraint.py#L13-L16) to document `>=X`-blocking → skip while I was in there.

scripts/cascade/discover.sh
stwilt: Done — header at [`scripts/cascade/discover.sh` L9-L15](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/scripts/cascade/discover.sh#L9-L15) and inline `widen:*` comment at [`scripts/cascade/discover.sh` L181-L187](https://github.com/PlainsightAI/openfilter/blob/e5d026cfba924fe9220ae23d45313aed1e528ffd/scripts/cascade/discover.sh#L181-L187) both refreshed to reflect the actual widenable set (`<` / `<=` / `==` / `~=`) and call out `>=`-blocking explicitly.

lucasmundim (approved):
Re-review covering e5d026c.

All three prior comments addressed cleanly:

- `>=` removed from `_is_widenable_block` — the asymmetry with `_BUMPABLE_OPS` is now intentional and documented at `scripts/cascade/check_constraint.py` L13-L22 (`>=X` only blocks when target < X, so applying the rewriter's `>=` rule would downgrade, not widen). `test_ge_lower_bound_blocking_is_skip` locks it in.
- Test docstring attribution dropped; Coverage block extended to document `>=X`-blocking → skip.
- `discover.sh` module header (L9-L15) and inline `widen:*` comment (L181-L187) refreshed to match the actual policy.

Ran the full cascade suite locally against e5d026c — 30 passed (12 check_constraint + 18 bump_strategy). LGTM.

shingonoide (commented):
Lucas covered the main design moves; one additive finding and a defensive note.

### Important

**`~=` widening changes the spec's arity, which can narrow the compatible range**
`scripts/cascade/bump-strategy.sh` — `_BUMPABLE_OPS` (line 39), `rewrite_requirement` (line 96)

`_BUMPABLE_OPS = ("==", ">=", "~=")` and the rewrite emits `f"{leading}{s.operator}{of_version}{trailing}"`, so `~=<input>` becomes `~=<of_version>` with whatever arity `of_version` carries. For a two-component input like `~=0.1` (desugars to `>=0.1,<0.2`), rewriting with `of_version=0.2.5` produces `~=0.2.5` (`>=0.2.5,<0.2.6`) — narrower than the original compatible range, not wider. The existing tests only cover the three-component case (`~=0.1.30` → `~=0.3.0`) where arity happens to match.

Two paths forward: normalize the rewritten clause to the input's arity, or treat sub-three-component `~=` as non-widenable in `_is_widenable_block` so the cascade falls back to `skip:`. A test for two-component `~=` input would lock the chosen behavior in either way.

### Non-blocking

**`SpecifierSet.__contains__` defaults to `prereleases=False`**
`scripts/cascade/check_constraint.py:79-91`

`combined = SpecifierSet()` is built without `prereleases=True`, and the membership check at line 91 (`v in combined`) uses the same default. A pre-release `OF_VERSION` (e.g. `0.3.0a1`) would test as `False` against a constraint that logically admits it, mis-classifying as `widen:` or `skip:` instead of `ok:`. Under the current pipeline this can't trigger — `.github/workflows/cascade-on-tag.yaml:85-90` rejects any `OF_VERSION` not matching `^[0-9]+\.[0-9]+\.[0-9]+$` before the script runs — so this is latent risk, only active if the script is ever invoked directly. A `prereleases=True` on the SpecifierSet construction or a docstring note saying pre-release `OF_VERSION` is unsupported would close it cheaply.

PR #95, opened 2026-05-15
feat(filter_runtime): windowed async batch dispatch (FILTER-457/458/459/460)

stwilt:
# 📦 Pull Request

## 📋 What does this PR do?

Extends `process_batch()` (from openfilter#61) with four composable additions that together let filters with the "ring buffer + curated dispatch + async network call" pattern (every remote-VLM / retrieval / captioning filter today) use the runtime-owned accumulator without reimplementing dispatch machinery per filter:

1. **[FILTER-457](https://plainsight-ai.atlassian.net/browse/FILTER-457) — deferred-callable `process_batch()`.** Each result slot in `process_batch()`'s returned list may now be a `Callable[[], dict[str, Frame] | None]` that `mq.send` invokes lazily — same shape as `process()`'s already-supported `Callable` return. Lets a filter hand the batch's actual work to a worker thread and return a poll closure instead of parking the loop thread on the batch's wall-clock duration. Each `Callable` slot is wrapped in a `_timed_batch_slot` closure that records its own per-slot wall-clock on resolution; pure-synchronous batches preserve existing EMA semantics. (`eefb995`)

2. **[FILTER-459](https://plainsight-ai.atlassian.net/browse/FILTER-459) — sliding window + `select_batch` hook.** New `accumulate_window: int` config decouples the runtime ring buffer from `batch_size`. When set, the buffer becomes a `deque(maxlen=window)`, frames are NOT removed on dispatch (the deque's maxlen handles eviction), and the dispatch trigger fires every `batch_size` new frames via a `_frames_since_last_dispatch` counter. The full ring is passed to a new public `Filter.select_batch()` hook that picks which frames go to `process_batch()`. Default `select_batch` returns the most recent `batch_size` entries, so unset or `window == batch_size` produces no behavior change. (`2f243a8`)

3. **[FILTER-458](https://plainsight-ai.atlassian.net/browse/FILTER-458) — filter-driven flush + `batch_trigger`.** Public `Filter.flush_batch()` schedules an immediate dispatch on the next opportunity (next frame append OR next `loop_once` idle poll). New `batch_trigger: "auto" | "manual"` config: `"manual"` suppresses both the count-based trigger AND the `accumulate_timeout_ms` watcher, leaving `flush_batch()` as the only drain path. Safety cap at `batch_size * 4` force-flushes the full buffer with a warning to surface forgotten flushes. (`4b9584a`)

4. **[FILTER-460](https://plainsight-ai.atlassian.net/browse/FILTER-460) — `batch_workers` pool.** New `batch_workers: int` config (default 1) runs `process_batch()` on a `ThreadPoolExecutor` so the loop accumulates the next batch while previous batches are in flight. Each pool submission returns one `Callable` per slot; the existing `mq.send` `Callable`-polling path resolves them in submission order (loop_once strictly feeds them to `mq.send` in order). Backpressure via `Semaphore(batch_workers)`: when all workers are busy, `_execute_batch_pool` blocks on `acquire()` and the loop stops `recv`'ing — natural upstream backpressure rather than silent drops. `batch_workers=1` (default) keeps the synchronous inline path. Shutdown drain via `concurrent.futures.wait` with `batch_shutdown_timeout_s` (default 60s). (`c33ab38`)

The four commits are independent in spec but they all touch the same `_execute_batch` / `_process_frames_batched` / `loop_once` cluster, so they land in 457 → 459 → 458 → 460 order on one branch rather than as four parallel PRs. Each is left as a separate commit for reviewability; final PR-level squash is reviewer's call.

---

## 🔍 Why is this needed?

openfilter#61's `process_batch()` covered the GPU-bound case: a single forward pass over a batched tensor, where the synchronous dispatch fits the runtime loop fine. It does **not** cover the "ring buffer + curated dispatch + async network call" pattern that every remote-VLM / retrieval / captioning filter re-implements with its own `deque` + `threading.Thread` + dispatch queue + keepalive-during-shutdown wait.

The reference offender is `filter-scene-caption` (`filter.py:305-326`, `642-660`, `427-467`): it keeps a `frame_count * 4` ring buffer, fires inference on `(media-time interval elapsed AND active_key gate open AND focus session valid)`, dispatches via a `threading.Thread` + `deque(maxlen=32)` queue, and emits keepalive frames during shutdown until the queue drains. Every comparable filter (remote-VLM caption, embedding-distance retrieval, re-ID over the last N detections) reimplements the same shape.

Promoting this pattern to a runtime affordance is the same playbook openfilter#82 followed for zero-copy local IPC: take a thing every filter reimplements, fold it into the runtime, delete the per-filter machinery. A follow-up PR in `filter-scene-caption` will validate the migration end-to-end by removing its `_inference_thread` + `_dispatch_queue` + `_wait_with_keepalive` and switching to `batch_workers=1` + `batch_trigger="manual"` + `accumulate_window=frame_count*4` + a `select_batch` override that mirrors its `_sample_frames` strategy.

Goldenrod.3 milestone candidate under epic [FILTER-367 — Inference Performance & Scaling](https://plainsight-ai.atlassian.net/browse/FILTER-367).

---

## 🧪 How was it tested?

**Unit tests:** 28 new tests added to `tests/test_frame_accumulation.py` (5 + 8 + 9 + 6 across the four tickets), covering:

- **Deferred callables:** all-callable batches return `Callable` slots, mixed dict/callable batches preserve slot order, `Callable` slots resolving to `None` route through `mq.send`'s existing None handling, batch-level EMA is skipped when any slot is a `Callable` (matching `_process_frames_single`).
- **Sliding window:** default `accumulate_window` uses the original list buffer; setting it switches to a bounded `deque`; ring persists across dispatches in sliding mode and gets evicted when maxlen is reached; custom `select_batch` overrides route through every dispatch path (count-driven, timeout-driven, shutdown drain); invalid `accumulate_window < batch_size` raises at config validation.
- **Manual trigger:** `batch_trigger="manual"` suppresses count + timeout triggers, the `_batch_watcher` never starts, `flush_batch()` from `process()` includes the just-appended frame, safety cap at `batch_size * 4` force-flushes the FULL buffer (bypasses `select_batch` because the cap is a bailout for a bug), `flush_batch()` at `batch_size=1` is a no-op, manual+sliding compose correctly.
- **Worker pool:** `batch_workers=1` stays on the inline path (no pool created), pool mode returns `Callable` slots, two batches submitted back-to-back actually overlap in wall-clock (verified via `threading.Barrier`), third submission to a `workers=2` pool blocks on the semaphore until a worker frees, `fini()` drains in-flight batches within `batch_shutdown_timeout_s`.

**Final counts:** 56/56 tests in `test_frame_accumulation.py` pass (28 original + 28 new). 536/536 of the broader runnable suite passes. The 37 pre-existing `test_filter_rest.py` failures and 41 pre-existing `test_filter_webvis.py` failures are missing-`fastapi` collection errors, unrelated to this branch.

**Not tested in this PR:** real-pipeline validation against a remote VLM. The pool's overlap, ordering, and backpressure are verified via unit tests with explicit barriers/semaphores, but a `filter-scene-caption` migration PR (follow-up) is what proves the end-to-end shape against Gemini Flash. Each ticket includes that migration as an acceptance criterion.

---

## 🔗 Related Issues

- [FILTER-367](https://plainsight-ai.atlassian.net/browse/FILTER-367) — Inference Performance & Scaling (parent epic, Goldenrod.3)
- [FILTER-457](https://plainsight-ai.atlassian.net/browse/FILTER-457) — Allow `process_batch()` to return a deferred callable
- [FILTER-458](https://plainsight-ai.atlassian.net/browse/FILTER-458) — Filter-driven flush trigger (`Filter.flush_batch`)
- [FILTER-459](https://plainsight-ai.atlassian.net/browse/FILTER-459) — Separate accumulation window from batch size (`Filter.select_batch` hook)
- [FILTER-460](https://plainsight-ai.atlassian.net/browse/FILTER-460) — Multiple in-flight batches via opt-in worker pool (`batch_workers`)

Builds on [FILTER-369](https://plainsight-ai.atlassian.net/browse/FILTER-369) (openfilter#61, frame accumulation foundation).
Same runtime-affordance playbook as openfilter#82 (zero-copy local IPC).

---

N/A — runtime change, no UI. Test output:

```
$ uv run --extra dev python -m pytest tests/test_frame_accumulation.py -q
...
56 passed, 202 warnings in 4.94s
```

---

## ✅ Checklist

- [x] I have added or updated **tests** as needed (28 new tests in `tests/test_frame_accumulation.py`)
- [x] I have added or updated **documentation** as needed — N/A

[FILTER-457]: https://plainsight-ai.atlassian.net/browse/FILTER-457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[FILTER-459]: https://plainsight-ai.atlassian.net/browse/FILTER-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/openfilter/filter_runtime/filter.py
+++ b/openfilter/filter_runtime/filter.py
@@ -575,6 +575,10 @@ class Filter:
             flush-trigger order even when workers complete out of order. Submission BLOCKS when all workers are
             busy (natural upstream backpressure rather than silent drops). Useful for network-bound batches
             (remote VLM/retrieval calls) where call latency > flush interval.
+            A process_batch() exception in pool mode (batch_workers > 1) is logged and the batch is dropped —
+            the filter keeps running. This is intentionally non-fatal, and DIVERGES from the synchronous
+            batch_workers == 1 path, which re-raises and exits the run loop. Call self.exit() inside
+            process_batch() if a batch failure must stop the filter regardless of mode.
 
         batch_shutdown_timeout_s:
             Maximum seconds to wait for in-flight batch workers to drain during fini(). Default 60. Batches still
@@ -1676,8 +1680,16 @@ class Filter:
         with self._batch_inflight_lock:
             self._batch_inflight.add(future)
 
+        # Bind the semaphore by value. fini() nulls self._batch_submit_sem
+        # right after shutdown(wait=False), so a not_done future completing in
+        # the background afterwards would hit None.release() via self. The
+        # semaphore object itself stays valid; releasing it post-shutdown is
+        # harmless (nothing acquires it again). _batch_inflight_lock is never
+        # nulled, so referencing it through self stays safe.
+        submit_sem = self._batch_submit_sem
+
         def _on_done(f: concurrent.futures.Future) -> None:
-            self._batch_submit_sem.release()
+            submit_sem.release()
             with self._batch_inflight_lock:
                 self._batch_inflight.discard(f)
 
@@ -1694,9 +1706,28 @@ class Filter:
         """Pool-thread entry point. Runs process_batch() and returns the raw
         result list bracketed by start/end timestamps. Per-slot normalization
         + EMA + metadata enqueue all happen on resolution, not here, so
-        results emerge with timing tied to when downstream actually pulls."""
+        results emerge with timing tied to when downstream actually pulls.
+
+        A process_batch() exception is caught here and treated as a dropped
+        batch (every slot resolves to None), logged once. This is deliberately
+        NON-fatal and DIVERGES from the synchronous path (batch_workers == 1),
+        where _execute_batch re-raises and the run loop exits: pool mode exists
+        for network-bound batches (remote VLM / retrieval calls) where a
+        transient failure should not kill a long-running filter. A filter that
+        needs a batch failure to be fatal should detect it inside
+        process_batch() and call self.exit() — exit() sets stop_evt, which ends
+        the run loop regardless of this catch.
+        """
         t_in = time.time()
-        results = self.process_batch(batch)
+        try:
+            results = self.process_batch(batch)
+        except Exception:
+            logger.exception(
+                "process_batch() raised in pool worker on a batch of %d "
+                "frame(s); dropping the batch and continuing",
+                len(batch),
+            )
+            results = None
         t_out = time.time()
         if results is None:
             results = [None] * len(batch)
@@ -1720,12 +1751,18 @@ class Filter:
             try:
                 t_in, t_out, results = future.result()
             except Exception:
+                # Defensive: _batch_pool_worker catches process_batch()
+                # failures itself (logging once, dropping the batch), so this
+                # only fires on an unexpected pool-infrastructure failure
+                # (e.g. a cancelled future). Drop the slot rather than crash
+                # the loop — consistent with the pool path's non-fatal
+                # contract documented on _batch_pool_worker.
                 logger.exception(
-                    "Pool batch worker raised; slot %d/%d will not emit",
+                    "Pool future failed unexpectedly; slot %d/%d dropped",
                     slot_idx,
                     batch_len,
                 )
-                raise
+                return None
             if slot_idx >= len(results):
                 return None
             slot = results[slot_idx]
@@ -1739,9 +1776,16 @@ class Filter:
                     return None
             frames = self._normalize_frame_result(slot)
             process_time_ms = (t_out - t_in) * 1000
-            self._filter_time_in = t_in
-            self._filter_time_out = t_out
-            self._update_process_time_ema(process_time_ms)
+            # All slots of a pool batch share one process_batch() wall-clock
+            # (the future's t_in/t_out). Fold it into the EMA exactly once —
+            # on slot 0 — so a batch of N slots does not inflate the EMA
+            # N-fold. Mirrors the once-per-batch update in the synchronous
+            # _execute_batch path. _inject_timings still runs per slot: every
+            # frame needs its own timing metadata stamped.
+            if slot_idx == 0:
+                self._filter_time_in = t_in
+                self._filter_time_out = t_out
+                self._update_process_time_ema(process_time_ms)
             self._inject_timings(frames, t_in, t_out, process_time_ms)
             try:
                 self._metadata_queue.put_nowait((frames, self.emitter))
@@ -1797,21 +1841,29 @@ class Filter:
         # Timeout in batch mode — flush any partial batch and return early.
         # Only runs when frames is empty (timeout case), not when real frames are received.
         if batch_mode and not frames:
-            with self._batch_lock:
-                if self._frame_buffer:
-                    window_snapshot = list(self._frame_buffer)
-                    self._frames_since_last_dispatch = 0
-                    if not self._sliding_window:
-                        self._frame_buffer.clear()
-                    self._batch_flush_event.clear()
-                    self._cancel_batch_timer()
-                else:
-                    window_snapshot = None
-            if window_snapshot is not None:
-                batch = self.select_batch(window_snapshot)
-                flushed_results = self._execute_batch(batch)
-                for r in flushed_results:
-                    self._send_frames(r, outputs_timeout)
+            # Skip the timeout-driven drain in manual mode: batch_trigger="manual"
+            # contracts that ONLY flush_batch() drains the buffer, so a slow or
+            # idle source must not dispatch a partial batch here. A pending
+            # flush_batch() request is still honored — the recv-loop above picks
+            # up _batch_flush_event via _process_batch_flush before reaching this
+            # point — so manual flushes are not lost, only timeout flushes are
+            # suppressed.
+            if self._batch_trigger != "manual":
+                with self._batch_lock:
+                    if self._frame_buffer:
+                        window_snapshot = list(self._frame_buffer)
+                        self._frames_since_last_dispatch = 0
+                        if not self._sliding_window:
+                            self._frame_buffer.clear()
+                        self._batch_flush_event.clear()
+                        self._cancel_batch_timer()
+                    else:
+                        window_snapshot = None
+                if window_snapshot is not None:
+                    batch = self.select_batch(window_snapshot)
+                    flushed_results = self._execute_batch(batch)
+                    for r in flushed_results:
+                        self._send_frames(r, outputs_timeout)
             # In batch mode, timeout means "flush partial" — already done above.
             # No need to call process_frames({}) which would just return None.
             if (
@@ -2376,6 +2428,11 @@ class Filter:
             The default implementation calls process() per frame and evaluates any
             Callable returns eagerly to preserve pre-existing behavior; override
             process_batch() to opt into deferred-callable semantics.
+
+            Exception handling depends on batch_workers. With batch_workers == 1 (synchronous) a
+            process_batch() exception re-raises and exits the run loop. With batch_workers > 1 (pool) it
+            is logged once and the batch is dropped (every slot resolves to None) so the filter keeps
+            running. Call self.exit() from within process_batch() to stop the filter regardless of mode.
         """
         results = []
         for frames in batch:
@@ -2448,6 +2505,15 @@ class Filter:
     def _stop_batch_watcher(self) -> None:
         self._batch_stop.set()
         self._batch_wakeup.set()  # unblock the wait
+        # Join so shutdown sequencing is explicit: fini() proceeds straight to
+        # pool drain + mq.destroy() after this, and joining here closes the
+        # window where a late _batch_flush_event.set() from the watcher could
+        # race MQ teardown. daemon=True already rules out a leak; the join just
+        # makes the ordering deterministic. Bounded so a wedged watcher cannot
+        # block shutdown indefinitely.
+        watcher = self._batch_watcher
+        if watcher is not None:
+            watcher.join(timeout=1.0)
 
     def _reset_batch_timer(self) -> None:
         """Reset the accumulation timeout by waking the watcher."""
--- a/openfilter/filter_runtime/filter.py
+++ b/openfilter/filter_runtime/filter.py
@@ -9,6 +9,8 @@ import sys
 import threading
 import time
 import warnings
+from collections import deque
+from collections.abc import Sequence
 from datetime import datetime
 from multiprocessing import synchronize
 from typing import Any, Callable, Literal, List
@@ -293,6 +295,9 @@ class FilterConfig(
     accumulate_timeout_ms: (
         float | None
     )  # Max ms to wait for a full batch before flushing partial (default 100.0)
+    accumulate_window: (
+        int | None
+    )  # Ring buffer size in sliding-window mode; must be >= batch_size. Unset = one-shot buffer (current behavior).
 
     def clean(self):  # -> Self:
         """Return a clean instance of this config without any hidden items starting with '_'."""
@@ -534,6 +539,14 @@ class Filter:
             Maximum time in milliseconds to wait for a full batch before flushing a partial batch. Only used when
             batch_size > 1. Default 100.0. If frames arrive slower than this window the partial batch is flushed
             rather than waiting indefinitely.
+
+        accumulate_window:
+            Sliding-window ring buffer size; must be >= batch_size. When set, the runtime keeps the most recent
+            accumulate_window frames in a bounded ring rather than draining the buffer on each dispatch. The full
+            ring is passed to Filter.select_batch() which picks which frames go to process_batch(). Lets filters
+            with temporal sampling needs (caption/VLM, re-ID, audio-visual sync) reach across a wider window than
+            what fits in a single batch. Unset (the default) preserves the original one-shot drain-on-dispatch
+            behavior — equivalent to accumulate_window=batch_size.
 
     Environment variables:
         LOG_LEVEL:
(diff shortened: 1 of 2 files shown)


openfilter/filter_runtime/filter.py
```diff
@@ -1436,12 +1587,160 @@ def _execute_batch(self, batch: list[dict[str, Frame]]) -> list[dict[str, Frame]
...
+                # FILTER-457: per-slot Callable returned by process_batch.
+                # Resolve inline so this acts as a single deferred slot.
+                slot = slot()
+                if slot is None:
+                    return None
+            frames = self._normalize_frame_result(slot)
+            process_time_ms = (t_out - t_in) * 1000
+            self._filter_time_in = t_in
+            self._filter_time_out = t_out
+            self._update_process_time_ema(process_time_ms)
```
lucasmundim: **EMA double-counted in pool mode.**

Each slot in a batch independently calls `_update_process_time_ema(process_time_ms)` with the same per-batch duration, so a batch of N slots updates the EMA N times. The inline batch path at `_execute_batch` updates it once per batch. For `batch_workers > 1` and `batch_size > 1` the EMA will be systematically inflated.

Suggest guarding the timing updates with `if slot_idx == 0:` (matching the once-per-batch semantics of the sync path).

openfilter/filter_runtime/filter.py
```diff
@@ -1490,13 +1789,16 @@ def loop_once(self) -> None:
         if batch_mode and not frames:
```
lucasmundim, line 1843: **Manual `batch_trigger` silently bypassed here.**

The `batch_trigger` docstring (lines 561-568) states that in manual mode "dispatch only happens when the filter explicitly calls self.flush_batch()". This flush path unconditionally drains the buffer when `mq.recv` returns no frames (timeout or `sources_timeout`), with no `self._batch_trigger` check. A manual-mode filter with any source timeout will have partial batches dispatched without an explicit `flush_batch()` call, violating the documented contract.

`TestManualBatchTrigger` doesn't exercise this path. Suggest guarding with `if self._batch_trigger != "manual":`.

openfilter/filter_runtime/filter.py
```diff
@@ -1436,12 +1587,160 @@ def _execute_batch(self, batch: list[dict[str, Frame]]) -> list[dict[str, Frame]
...
+                    len(batch),
+                )
+                return []
+
+        future = pool.submit(self._batch_pool_worker, batch)
+        with self._batch_inflight_lock:
+            self._batch_inflight.add(future)
+
+        def _on_done(f: concurrent.futures.Future) -> None:
+            self._batch_submit_sem.release()
```
lucasmundim: **Race on `fini()` teardown for abandoned futures.**

`fini()` (around line 2154-2173) calls `concurrent.futures.wait(pending, timeout=...)`, then `self._batch_pool.shutdown(wait=False)` and sets `self._batch_submit_sem = None`. Any `not_done` future that completes after `fini()` returns will fire this `_on_done` callback, hitting `NoneType.release()` and `AttributeError`. The comment at L2169-2171 explicitly anticipates these futures finishing in the background, but the callback isn't guarded.

Suggest either checking `if self._batch_submit_sem is not None:` here or removing the done-callback before shutdown.

openfilter/filter_runtime/filter.py
```diff
@@ -1436,12 +1587,160 @@ def _execute_batch(self, batch: list[dict[str, Frame]]) -> list[dict[str, Frame]
...
+
+        def _resolve():
+            try:
+                t_in, t_out, results = future.result()
+            except Exception:
+                logger.exception(
+                    "Pool batch worker raised; slot %d/%d will not emit",
+                    slot_idx,
+                    batch_len,
+                )
```
lucasmundim, line 1764: **Pool exception surfaces asymmetrically vs the sync batch path.**

When `process_batch` raises in the pool worker, `future.result()` re-raises here for every slot of that batch. The exception then propagates up through `mq.send`'s callback closure into `loop_once` and `_send_frames`, neither of which catches it — so a pool batch failure crashes the loop many frames later, at send time, rather than at submission time as in the inline `_execute_batch` path.

Consider catching here, logging, and returning `None` (treat the slot as a dropped result), or at least documenting the asymmetric semantics.

tests/test_frame_accumulation.py
```diff
@@ -586,5 +586,630 @@ def test_loop_once_sends_all_batch_results(self):
...
+        f._execute_batch([{"main": Frame({"a": 2})}])
+        # Fini should wait for both to finish (< 5s timeout).
+        t0 = time.time()
+        f.fini()
+        elapsed = time.time() - t0
+        self.assertLess(elapsed, 5.0)
+        # All futures should have completed (no warning would fire).
+        with f._batch_inflight_lock:
+            # add_done_callback fires async — give it a tick to drain.
+            time.sleep(0.05)
```
lucasmundim: **Fixed-sleep anti-pattern reintroduced.**

PR #94 / commit d936bc8 (FILTER-461) specifically replaced the "single-step assert after a fixed sleep" pattern with a polling helper to eliminate CI flakes. This test holds `_batch_inflight_lock`, sleeps 50ms, then asserts — under loaded CI the 50ms window may be insufficient and the test will flake.

Suggest a poll-with-timeout style instead, e.g.:
```python
deadline = time.time() + 2.0
while time.time() < deadline:
    with f._batch_inflight_lock:
        if not f._batch_inflight:
            break
    time.sleep(0.01)
with f._batch_inflight_lock:
    self.assertEqual(len(f._batch_inflight), 0)
```

lucasmundim (commented):
### Code review

Found 5 issues (3 with high confidence, 2 moderate).

**High confidence**

1. EMA is double-counted in pool mode — every slot in a batch updates `_process_time_ema` independently, skewing it by a factor of N for `batch_workers > 1`.
2. `batch_trigger="manual"` is silently bypassed by the `sources_timeout` flush path in `loop_once` — violates the documented contract that only `flush_batch()` drains. Not covered by `TestManualBatchTrigger`.
3. `test_pool_drains_on_fini` reintroduces the fixed-sleep anti-pattern that PR #94 / commit d936bc8 (FILTER-461) explicitly removed.

**Moderate confidence**

4. The `_on_done` callback has no None-guard on `_batch_submit_sem`/`_batch_inflight_lock`, while `fini()` calls `shutdown(wait=False)` and immediately nulls them. Abandoned `not_done` futures completing later will `AttributeError` on `NoneType.release()`.
5. Pool-mode `process_batch` exceptions re-raise through `mq.send` into `loop_once` (asynchronously, many frames later), whereas the sync path raises at submission time on the loop thread — asymmetric and harder to handle.

See inline comments for specifics.

shingonoide (commented):
The pool-mode dispatch is well-structured: `_batch_inflight` is lock-guarded, the daemon watcher only signals via `_batch_flush_event`, and the loop thread preserves submission order by stashing `_batch_pending_results` and consuming Callables in order. Three observations.

**1. Sleep-then-assert pattern in five batch-accumulation tests** — `tests/test_frame_accumulation.py`

This is the same shape that was rewritten in #80 for topology tests. The same fix applies here:

- `test_timeout_sets_flush_flag` (`:180-182`): `time.sleep(0.1)` → `assertTrue(f._batch_flush_event.is_set())`. Replace with `self.assertTrue(f._batch_flush_event.wait(timeout=2.0))`.
- `test_flush_returns_partial_batch` (`:190-192`): same positive-side fix.
- `test_pool_drains_on_fini` (`:1208-1211`): `time.sleep(0.05)` then `assertEqual(len(f._batch_inflight), 0)`. Replace with a polling loop with deadline: `deadline = time.monotonic() + 2.0; while time.monotonic() < deadline and f._batch_inflight: time.sleep(0.005)`. (Also: the sleep is currently held inside `with f._batch_inflight_lock:` — by the time `fini()` has returned, `concurrent.futures.wait` has already drained the futures and the done-callbacks have completed, so the sleep is inert in practice. But holding the same lock the done-callback needs would be a deadlock if the assertion ever moves earlier; the polling form removes both issues.)
- `test_timer_reset_extends_accumulation_window` (`:424,:435-439`): `time.sleep(0.06)` + `time.sleep(0.04)` then `assertFalse(f._batch_flush_event.is_set())`. The positive assertion at `:443` already uses `wait(timeout=0.5)` — good. The `assertFalse` at `:436` is the negative regression guard and is structurally racy on a loaded CI runner. The cleanest fix is to make the watcher's clock injectable for this test (e.g. swap `time.monotonic` for a test fake) so the timer-reset behavior can be asserted without wall-clock dependency. Failing that, raise the margin substantially (e.g. 200ms timer-reset, 50ms negative-side assertion) so noise is well below the design margin.
- `test_pool_backpressure_blocks_submission_when_full` (`:1164-1168`): `blocked.wait(timeout=1.0)` is fine, but the following `time.sleep(0.1)` → `assertFalse(released.is_set())` is the racy part. The third submission's blocking is already observable: it does not return from `_execute_batch_pool` until a semaphore slot frees. A deterministic alternative is to run the third submission in a thread and assert that thread is still alive after a short window, then unblock and assert it joins — that uses thread liveness rather than timing.

**2. No coverage for pool-worker exception propagation** — `openfilter/filter_runtime/filter.py:1709-1718`

`_wrap_pool_slot_callable._resolve()` logs the future exception, then `raise`s. In pool mode `_execute_batch_pool` returns the wrapped Callables, and `mq.send` invokes each Callable inline when it's time to dispatch. If `process_batch()` raised in the pool worker, the re-raise from `_resolve()` propagates out through `mq.send` → `_send_frames` (no try/except at `:1744-1750`) → `loop_once`. The run loop's `except loop_exc` defaults to `YesLoopException`, which a plain `Exception` from the pool worker does not match, so the filter exits via `fini()` and re-raises.

Whether that's the intended behavior (crash the filter loudly) or a regression risk (one bad batch kills the filter) deserves a test that submits a batch where `process_batch()` raises and asserts the resulting filter state. Either way, the contract should be locked in before someone refactors `mq.send`'s Callable handling.

**3. `_stop_batch_watcher` does not join the watcher thread** — `filter.py:2435-2437`

`fini()` signals `_batch_stop` + `_batch_wakeup` and immediately proceeds to pool shutdown and `mq.destroy()`. The watcher is `daemon=True` so there is no leak, and the lock around `_batch_flush_event` mutation makes the practical race narrow. But the test at `:222` (`test_watcher_stopped_on_cleanup`) already does `f._batch_watcher.join(timeout=1.0)` to verify the watcher exits — production should do the same. Adding `self._batch_watcher.join(timeout=1.0)` after the two `set()` calls makes shutdown sequencing explicit and removes the window where a late `_batch_flush_event.set()` could race with the MQ teardown.

Nothing else flagged. `_batch_submit_sem`, `_batch_inflight`, and `_batch_flush_event` are initialized before any worker thread can observe them; `_batch_inflight` operations are properly lock-guarded; ordering of pool callables preserves submission order through `_batch_pending_results`; the FILTER-457/458/459/460 work splits cleanly across four commits on the branch.

openfilter/filter_runtime/filter.py
stwilt: Fixed in `0362f4a`. `_wrap_pool_slot_callable._resolve()` now folds the shared per-batch wall-clock into the EMA only on `slot_idx == 0` — once per batch, matching the synchronous `_execute_batch` path. `_inject_timings` still runs per slot, since each frame needs its own timing metadata stamped.

openfilter/filter_runtime/filter.py
stwilt: Fixed in `0362f4a`. The `loop_once` source-timeout drain is now guarded by `if self._batch_trigger != "manual"`. A pending `flush_batch()` is still honored — the recv-loop picks up `_batch_flush_event` via `_process_batch_flush` before this point — so manual flushes aren't lost, only timeout flushes are suppressed. Added `test_manual_mode_not_drained_by_source_timeout` to cover the path.

openfilter/filter_runtime/filter.py
stwilt: Fixed in `0362f4a`. `_execute_batch_pool` now binds the semaphore by value (`submit_sem = self._batch_submit_sem`) and `_on_done` releases that local. A `not_done` future completing after `fini()` nulls `self._batch_submit_sem` releases the still-valid object — harmless, nothing acquires it again — instead of hitting `None.release()`. Went with the captured-local over a `None`-check to avoid the check-then-null race window.

openfilter/filter_runtime/filter.py
stwilt: Addressed in `0362f4a` — went with catch-and-drop. `_batch_pool_worker` now catches the `process_batch()` exception, logs it once, and drops the batch (every slot resolves to `None`), so the filter keeps running. Intentionally non-fatal and divergent from the synchronous path: pool mode targets network-bound batches (remote VLM/retrieval) where a transient failure shouldn't kill a long-running filter. A filter that needs a fatal batch failure calls `self.exit()` inside `process_batch()`. The divergence is documented on `_batch_pool_worker` and the `batch_workers` / `process_batch` docstrings, and locked in by `test_pool_worker_exception_drops_batch_and_continues`.

tests/test_frame_accumulation.py
stwilt: Fixed in `0362f4a`. `test_pool_drains_on_fini` now polls with a 2s deadline instead of the fixed sleep, and the wait is no longer held inside `_batch_inflight_lock` (the done-callback needs that lock). Ran the same de-flake pass across the other four sleep-then-assert tests shingonoide flagged — `test_timeout_sets_flush_flag` / `test_flush_returns_partial_batch` now use `Event.wait()`, `test_pool_backpressure_blocks_submission_when_full` uses event-waits + thread-liveness, and `test_timer_reset_extends_accumulation_window` widened its margins to a 200ms watcher window.

tests/test_frame_accumulation.py
```diff
@@ -586,5 +593,731 @@ def test_loop_once_sends_all_batch_results(self):
...
+        # any slot is a Callable — matches _process_frames_single's behavior.
+        self.assertEqual(f._process_time_ema, prior_ema)
+
+        resolved = closure()
+        self.assertIsInstance(resolved, dict)
+        self.assertIn("main", resolved)
+        self.assertTrue(resolved["main"].data["resolved"])
+
+        # Closure resolution should have updated the EMA and queued metadata.
+        self.assertNotEqual(f._process_time_ema, prior_ema)
```
lucasmundim: **Flaky EMA assertion.**

The closure body (`lambda: {"main": Frame({"resolved": True})}`) can resolve faster than the wall-clock resolution, so `(t_out - t_in) * 1000` rounds to `0.0` and the EMA stays equal to `prior_ema = 0.0`. Reproduces ~1/10 locally.

Same root cause and same fix shape as the de-flake pass already in this commit; suggest either making the lambda do a tiny `time.sleep(0.001)`, or asserting on `f._metadata_queue.put_nowait.assert_called()` + `f._filter_time_in != 0` instead of the EMA delta.

lucasmundim (approved):
Re-review on `0362f4a`: all five prior issues fixed and verified by the new tests. One additional flake found below — pre-existing in the FILTER-457 commit, not introduced by the fix-up, but worth catching in the same de-flake sweep.

PR #85, opened 2026-04-26
feat(cascade): tag-triggered bump-PR cascade

stwilt:
Replaces `cloudbuild-cascade.yaml`'s central image rebuild with a GitHub Actions workflow on `push: tags: 'v*'` that walks `filter-*` org repos and opens mechanical bump PRs via `plainsight-bot`. Each PR writes `VERSION` + a `RELEASE.md` stanza + the `openfilter` pin in `pyproject.toml` (and `requirements*.txt` if present) for one consumer.

PR-opening mechanics (supersede stale + commit-as-bot + push + `gh pr create` + `gh pr merge --auto --squash`) are delegated to the new `open-mechanical-pr` composite action in `gh-actions-public` (separate PR — needs to merge first for this cascade's `@main` ref to resolve at runtime).

Bump fan-out is a `fail-fast: false` matrix with `max-parallel: 6` — one consumer's failure does not abort siblings; cascading is best-effort across the org. Slack alert on workflow failure mirrors the Block Kit shape from `client-portal/.github/workflows/update-api-client.yaml`.

**Deletions:**
- `cloudbuild-cascade.yaml` (Cloud Build cascade, replaced)
- `scripts/build-filters.sh` (per-image rebuild orchestrator)
- `cloud.cascade*` Makefile targets
- The `cascade-dry-run` CI job whose paths-filter referenced the removed files

**Move:** `scripts/check_constraint.py` → `scripts/cascade/check_constraint.py` (now invoked by the new discovery script as the source of truth on openfilter constraint semantics).

The Cloud Build trigger removal is a separate Terraform PR (deletion-only, in the `infrastructure` repo).

---

[FILTER-306](https://plainsight-ai.atlassian.net/browse/FILTER-306) delivered the centralized rebuild loop. It had two structural problems:

* **No consumer release ever gets cut.** Each consumer's `VERSION` / `RELEASE.md` / tag does not advance — only `:latest` image subscribers see updates. Customers pinning specific filter versions stay on the old runtime indefinitely.
* **Ownership is inverted.** The central cascade mutates images owned by other teams' repos without their review.

This PR flips the model: openfilter doesn't rebuild anything. It opens mechanical PRs that each consumer's own release flow handles. Each consumer's image publish is authored by its own release workflow, gated by its own codeowners.

Full design context (including the DT-146 prevent/cleanup safety nets that gate auto-merge correctly):
[DT-145](https://plainsight-ai.atlassian.net/browse/DT-145).

---

[...]

PR #97, opened 2026-05-16
fix(release): use GH_BOT_USER_PAT for gh release create so cascade-on-tag fires

stwilt:
## 📋 What does this PR do?

Swaps `secrets.GITHUB_TOKEN` → `secrets.GH_BOT_USER_PAT` on the `Create GitHub Release` step's `GH_TOKEN` env in
`.github/workflows/create-release.yaml`, so the tag-push event registers as an actor-attributed push (plainsight-bot via the org PAT) and triggers `cascade-on-tag.yaml`'s `on: push: tags: ['v[0-9]+.[0-9]+.[0-9]+']` listener. Adds an 11-line block comment explaining the rationale inline so a future reader doesn't "simplify" it back to `GITHUB_TOKEN` and re-introduce the bug.

One-line behavioral change; one comment block.

## 🔍 Why is this needed?

`cascade-on-tag.yaml` (added in #85) listens for release-tag pushes and fans out bump PRs to every eligible `filter-*` consumer. v0.2.1 was the first real release since that workflow merged — and the cascade silently did not auto-fire. Per [GitHub's docs on `GITHUB_TOKEN`](https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow):

> "When you use the repository's GITHUB_TOKEN to perform tasks, events
> triggered by the GITHUB_TOKEN, with the exception of
> `workflow_dispatch` and `repository_dispatch`, will not create a new
> workflow run."

The `Create GitHub Release` step's `gh release create` was running with `GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}`. The tag push *did* happen, but no `push: tags` event registered against the workflow listener.

Workaround for v0.2.1: manually `gh workflow run cascade-on-tag.yaml --ref main` ([run 25972753116](https://github.com/PlainsightAI/openfilter/actions/runs/25972753116)). That fired under `workflow_dispatch` (exempt from the anti-recursion clause) and ran clean: 30/30 bumps succeeded. Without this PR, every release would require the same manual nudge.

## 🧪 How was it tested?

**Empirical verification** that PAT-attributed tag pushes do trigger downstream `push: tags` workflows in this org, with this exact secret:

`PlainsightAI/openfilter-mcp/.github/workflows/auto-tag.yml` pushes a release tag via `git push https://x-access-token:${GH_BOT_USER_PAT}@github.com/...`. The sibling workflow `release-tag-bump-gitops.yaml` listens on `push: tags` and successfully fired for v0.2.2 at 2026-05-16T20:06:57Z ([run 25971651045](https://github.com/PlainsightAI/openfilter-mcp/actions/runs/25971651045)): `event: push`, `actor: plainsight-bot`, `triggering_actor: stwilt`, `conclusion: success`.

This is direct same-org, same-secret proof. The `gh release create` REST call and `git push` git-protocol path both authenticate as the token owner; GitHub's anti-recursion clause is scoped to GITHUB_TOKEN, not to which API method is used.

Post-merge verification (covered by Monday's v1.0 cut):

- [ ] Dispatch Create Release on main
- [ ] Observe `cascade-on-tag.yaml` run appear in the run list with `event=push` within ~minutes of the tag push, without manual nudging

## 🔗 Related Issues

- [FILTER-462](https://plainsight-ai.atlassian.net/browse/FILTER-462) — root-cause analysis and fix proposal (this PR implements the proposed fix).
- Discovered live during the v0.2.1 cut on 2026-05-16. Cascade did not auto-fire; had to be manually `workflow_dispatch`'d. Once dispatched, ran clean (30/30 bumps succeeded).
- Adjacent finding tracked separately: cascade-opened PRs across the `filter-*` repos all have `autoMergeRequest: null` despite `AUTO_MERGE='true'` being passed to `open-mechanical-pr` — likely a per-repo "Allow auto-merge" setting issue, not a cascade bug.

## ✅ Checklist

- [x] I have added or updated **tests** as needed (no test impact; release-pipeline-only change verified empirically against analogous PAT-tag-push pattern in `openfilter-mcp`)
- [x] I have added or updated **documentation** as needed (added an 11-line inline rationale comment so the choice survives future readers)

[FILTER-462]: https://plainsight-ai.atlassian.net/browse/FILTER-462?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -90,7 +90,18 @@ jobs:
             --title "Release ${{ env.CHANGELOG_VERSION }}" \
             --notes-file release-notes.md
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          # GH_BOT_USER_PAT (org-level PAT) rather than secrets.GITHUB_TOKEN:
+          # events created by the intrinsic GITHUB_TOKEN do not trigger
+          # downstream workflows (GitHub's anti-recursion safeguard). The
+          # tag-push from this step is what cascade-on-tag.yaml listens for
+          # (`on: push: tags: ['v[0-9]+.[0-9]+.[0-9]+']`); with GITHUB_TOKEN
+          # the tag is created but no workflow run is registered for the
+          # event, so the cascade silently fails to auto-fire on releases
+          # and has to be manually dispatched every time. Using the bot PAT
+          # attributes the push to plainsight-bot, which is a real user
+          # identity from GitHub's perspective and does trigger downstream
+          # workflows. See FILTER-462.
+          GH_TOKEN: ${{ secrets.GH_BOT_USER_PAT }}
           RELEASE_NOTES: ${{ steps.changelog.outputs.description }}
         shell: bash
 
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -4,7 +4,13 @@ on:
   workflow_dispatch:
 
 permissions:
-  contents: write
+  # contents: read (not write): tag creation + GH release creation moved
+  # to GH_BOT_USER_PAT in the Create GitHub Release step (FILTER-462), so
+  # the workflow's intrinsic GITHUB_TOKEN no longer needs write on repo
+  # contents. The remaining GITHUB_TOKEN consumers are read-only
+  # (mukunku/tag-exists-action queries the tags API; everything else
+  # reads code or writes to external registries — PyPI, Docker Hub).
+  contents: read
   pull-requests: read
 
 jobs:


.github/workflows/create-release.yaml
```diff
@@ -90,7 +90,18 @@ jobs:
...
+          # downstream workflows (GitHub's anti-recursion safeguard). The
+          # tag-push from this step is what cascade-on-tag.yaml listens for
+          # (`on: push: tags: ['v[0-9]+.[0-9]+.[0-9]+']`); with GITHUB_TOKEN
+          # the tag is created but no workflow run is registered for the
+          # event, so the cascade silently fails to auto-fire on releases
+          # and has to be manually dispatched every time. Using the bot PAT
+          # attributes the push to plainsight-bot, which is a real user
+          # identity from GitHub's perspective and does trigger downstream
+          # workflows. See FILTER-462.
+          GH_TOKEN: ${{ secrets.GH_BOT_USER_PAT }}
```
lucasmundim, line 110: **Operational note (non-blocking):** this expands `GH_BOT_USER_PAT`'s blast radius — it's now load-bearing for the *release itself*, not just the downstream cascade. Before this PR, a missing/expired PAT meant "cascade silently doesn't fire" (recoverable via manual `workflow_dispatch`); after this PR, the same condition means `gh release create` fails outright and blocks `prepare-pypi-publish` / `publish-docker-images` too. The stricter failure mode is arguably better (visible vs. silent), but worth making sure a PAT-rotation runbook exists somewhere on-call can find it. Not something to fix here.

lucasmundim (commented):
LGTM. Minimal-surface fix; the empirical proof from `openfilter-mcp` (`release-tag-bump-gitops` firing on a PAT-attributed tag push) is convincing, and the inline rationale comment is the right altitude to keep future readers from "simplifying" it back. Two non-blocking observations — one inline, one general:

**Optional follow-up (out of scope here):** with the release step now using the PAT, the workflow's top-level `permissions: contents: write` (line 7) is no longer needed for the release step itself — the only remaining consumer of the default `GITHUB_TOKEN` with that scope is the `Checking Tag` step (line 83), which is read-only. Could be tightened to `contents: read` for least-privilege in a quick follow-up PR. Flagging only; happy to leave as-is.

stwilt:
Thanks for the read.

1. **Permissions tightening**: applied in 7c38dcc. Top-level `permissions: contents: write` → `contents: read`. Every job verified — only `Create GitHub Release` writes content (now via the PAT, not the intrinsic GITHUB_TOKEN); everything else reads code or writes to external registries (PyPI, Docker Hub).

2. **PAT blast radius**: noted. Failure mode is recoverable the same way any release-pipeline secret would be — rotate `GH_BOT_USER_PAT` in org secrets, re-dispatch Create Release. The PR moves a silent failure (cascade-doesn't-fire) into a visible one (release fails at tag creation), which is the better tradeoff. No special runbook needed beyond the implicit "if the secret is broken, fix the secret."

shingonoide (approved):
The fix is correct and well-reasoned. Full-file read confirms nothing is broken by the permissions tightening.

**GH_TOKEN swap — correct step, correct behavior**

`GH_TOKEN: ${{ secrets.GH_BOT_USER_PAT }}` is set exclusively on the `Create GitHub Release` step (line 110). Every other step that touches `GITHUB_TOKEN` is read-only:

- `Checking Tag` (line 89, `mukunku/tag-exists-action`) — reads the tags API, no write needed, `contents: read` is sufficient.
- All build/publish steps (`prepare-pypi-publish`, `publish-to-pypi`, `publish-docker-images`) authenticate via their own secrets (`PLAINSIGHT_PYPI_TOKEN`, `DOCKERHUB_ACCESS_TOKEN`) and never touch `GITHUB_TOKEN` at all.

No job silently depended on `contents: write` from the intrinsic token. The top-level permissions block (`contents: read`, `pull-requests: read`) is safe across every job in this file. `gh release create` now creates both the release object and the tag via the PAT (not the intrinsic token), so the tag-creation path keeps working.

**Inline comment accuracy**

The 11-line comment (lines 99–109) is factually correct. GitHub's anti-recursion rule — events triggered by `GITHUB_TOKEN` do not fire downstream workflows — is accurately described. The claim that using a bot PAT attributes the push to a real user identity and does trigger downstream workflows is correct platform behavior.

**cascade-on-tag.yaml cross-check**

The inline comment quotes the tag pattern as `on: push: tags: ['v[0-9]+.[0-9]+.[0-9]+']`. `cascade-on-tag.yaml:16` matches exactly.

**Mundão's two observations — resolved**

1. **PAT blast radius** (inline at line 110): the silent→visible failure-mode tradeoff is the right call.
2. **Permissions tightening**: applied in 7c38dcc, with rationale inlined for future readers.

PR #104, opened 2026-05-21
fix(cascade): anchor RELEASE.md bumps on [Unreleased] section (DT-145)

lucasmundim:
## Summary

The cascade bump-strategy script previously picked the **first `## ` header** in RELEASE.md and wrote the `- Bump openfilter to X.Y.Z` bullet under it. That assumption breaks whenever the topmost `## ` isn't actually the Unreleased section. Two real failure modes seen in the wild:

- **filter-sweet-green-subject-data-aggregator [#39](https://github.com/PlainsightAI/filter-sweet-green-subject-data-aggregator/pull/39/files)** — file had no `## [Unreleased]` section; the bump was written into `## v0.1.27 - 2026-04-24` (an already-tagged release), with a fresh `### Changed` subsection created inside the released block.
- **FaceGuard** — preamble layout caused a bump bullet to land **above** `## [Unreleased]` in the file intro, alongside other orphan bullets from prior bad runs.

## Fix

`scripts/cascade/bump-strategy.sh` now:

1. **Anchors explicitly on the Unreleased section**, not the first `## ` header. Recognizes `## Unreleased`, `## [Unreleased]`, and `## (unreleased)` (case-insensitive) as the same slot — covers Keep a Changelog convention and the legacy shapes already in the codebase, and ensures re-runs don't append a duplicate Unreleased block.
2. **Creates a new `## [Unreleased]` block** immediately before the first released `## v...` header when none exists — never touches a tagged release block.
3. **Keeps the legacy "no `## ` at all"** branch — appends a new Unreleased block at end of file for brand-new RELEASE.md cases.
4. New files use `## [Unreleased]` (the shape `changelog-parser-action` accepts) instead of the legacy `## (unreleased)`.

No bash structure changed; only the inner Python heredoc.

## Tests

5 new regression tests in `tests/test_cascade_bump_strategy.py`:

- `test_release_md_creates_unreleased_block_when_only_released_versions_exist` — SGSDA #39 case
- `test_release_md_uses_existing_bracketed_unreleased_section` — Keep a Changelog convention
- `test_release_md_recognizes_unreleased_without_brackets` — matches openfilter's own RELEASE.md shape (`## Unreleased`)
- `test_release_md_preamble_bullets_do_not_anchor_insert` — FaceGuard case
- `test_release_md_idempotent_when_creating_unreleased_block` — re-running doesn't duplicate the new Unreleased block or its bullet

All 23 tests pass (18 pre-existing + 5 new).

## Follow-up

A companion PR against `changelog-parser-action` will add validation so future hand-edits / mis-bumps fail CI rather than relying on this script being the only writer. Tracked separately to keep producer and validator on independent cadences.

DT-145: https://plainsight-ai.atlassian.net/browse/DT-145

## Test plan

- [ ] CI: `cascade-pr-checks` (shellcheck + actionlint + discover smoke) passes
- [ ] CI: `run-unit-tests` matrix passes (the new tests are part of the suite)
- [ ] CI: `check-release-log` skips this PR (no `openfilter/**` source files touched)

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/104"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1781988955&installation_id=128257093&pr_number=104&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F104&signature=a0a20874e1d2a748657115b68249b8db8c01d694388122d8fbf0962063d38541"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need. Autofix is disabled.</sup>

### DIFF
--- a/scripts/cascade/bump-strategy.sh
+++ b/scripts/cascade/bump-strategy.sh
@@ -3,7 +3,7 @@
 # pyproject.toml ([project.dependencies] + [project.optional-dependencies.*]),
 # requirements*.txt, and a `### Changed` bullet in RELEASE.md. Run from the
 # clone root. Idempotent. Required env: OF_VERSION (bare semver). Required
-# tooling: python3 with tomlkit + packaging.
+# tooling: python3 with tomlkit + packaging + mistletoe.
 # DT-145: https://plainsight-ai.atlassian.net/browse/DT-145
 set -euo pipefail
 
@@ -242,18 +242,42 @@ done
 #    so 0.1.9 doesn't match 0.1.99.
 python3 <<'PY'
 import os
-import re
 import sys
+
+import mistletoe
+from mistletoe.block_token import Heading
 
 of_version = os.environ["OF_VERSION"]
 path = "RELEASE.md"
 bullet = f"- Bump openfilter to {of_version}"
 
-# Recognize `## Unreleased`, `## [Unreleased]`, and `## (unreleased)` as the
-# same slot (case-insensitive). Keep a Changelog uses `[Unreleased]` and that
-# is what changelog-parser-action accepts; the other shapes are legacy and
-# treated as equivalent so a re-run doesn't insert a duplicate section.
-unreleased_re = re.compile(r"^##\s+[\[\(]?\s*unreleased\s*[\]\)]?\s*$", re.IGNORECASE)
+
+def heading_text(node: Heading) -> str:
+    """Flatten a heading's inline children to plain text. RawText (and the
+    common inline tokens we'd see in a changelog header — InlineCode,
+    EscapeSequence) all expose `.content`; anything else is recursed into."""
+    out: list[str] = []
+    for child in node.children:
+        content = getattr(child, "content", None)
+        if isinstance(content, str):
+            out.append(content)
+        elif hasattr(child, "children"):
+            out.append(heading_text(child))
+    return "".join(out).strip()
+
+
+def is_unreleased(text: str) -> bool:
+    """Match `Unreleased`, `[Unreleased]`, `(unreleased)`, with surrounding
+    whitespace. Case-insensitive — covers Keep a Changelog convention and the
+    legacy shapes already in the codebase (so a re-run doesn't duplicate)."""
+    return text.lower().strip("[]() \t") == "unreleased"
+
+
+def h2_line_idx(node: Heading) -> int:
+    """mistletoe's `line_number` is 1-indexed and points at the heading's
+    source line; convert to a 0-indexed offset into the `lines` array."""
+    return node.line_number - 1
+
 
 if not os.path.exists(path):
     with open(path, "w", encoding="utf-8") as f:
@@ -278,24 +302,30 @@ if any(line.rstrip() == bullet for line in text.splitlines()):
 
 lines = text.splitlines(keepends=True)
 
-# Anchor explicitly on the Unreleased section, not "first ## " — the latter
+# Parse via mistletoe (CommonMark AST) to locate sections — robust against
+# whitespace/escape/inline quirks that regex matching glossed over. Mutation
+# is still done by splicing the source `lines` array so original formatting
+# (blank lines, preamble, indentation) round-trips verbatim; re-serializing
+# through the AST would not preserve those.
+doc = mistletoe.Document(text)
+h2_headings = [
+    n for n in doc.children if isinstance(n, Heading) and n.level == 2
+]
+
+# Anchor explicitly on the Unreleased section, not "first H2" — the latter
 # landed bumps under whichever release sat on top, including already-tagged
 # ones (DT-145 follow-up: filter-sweet-green-subject-data-aggregator #39
 # wrote into `## v0.1.27`).
-unreleased_idx = next(
-    (i for i, line in enumerate(lines) if unreleased_re.match(line.rstrip())),
+unreleased_node = next(
+    (h for h in h2_headings if is_unreleased(heading_text(h))),
     None,
 )
 
-if unreleased_idx is None:
-    # No Unreleased section. Inject one before the first released `## v...`
-    # header so the bullet never lands in a tagged block. If there's no `## `
-    # at all, fall back to the legacy "append at end" behavior.
-    first_release_idx = next(
-        (i for i, line in enumerate(lines) if line.startswith("## ")),
-        None,
-    )
-    if first_release_idx is None:
+if unreleased_node is None:
+    # No Unreleased section. Inject one before the first H2 so the bullet
+    # never lands in a tagged block. If there's no H2 at all, fall back to
+    # the legacy "append at end" behavior.
+    if not h2_headings:
         suffix = "" if text.endswith("\n") else "\n"
         suffix += (
             "\n"
@@ -310,6 +340,7 @@ if unreleased_idx is None:
         print(f"{path}: appended new [Unreleased] block with bump entry")
         sys.exit(0)
 
+    first_release_idx = h2_line_idx(h2_headings[0])
     new_block = [
         "## [Unreleased]\n",
         "\n",
@@ -324,18 +355,33 @@ if unreleased_idx is None:
     print(f"{path}: inserted new [Unreleased] block before first released section")
     sys.exit(0)
 
-next_release_idx = next(
-    (j for j in range(unreleased_idx + 1, len(lines)) if lines[j].startswith("## ")),
-    len(lines),
+# Existing Unreleased section — bound it by the next H2 (or EOF) and either
+# append to the existing `### Changed` subsection or create one.
+unreleased_idx = h2_line_idx(unreleased_node)
+later_h2 = next(
+    (h for h in h2_headings if h2_line_idx(h) > unreleased_idx),
+    None,
 )
+next_release_idx = h2_line_idx(later_h2) if later_h2 is not None else len(lines)
 block = lines[unreleased_idx:next_release_idx]
 
-changed_idx = next(
-    (k for k, line in enumerate(block) if line.rstrip() == "### Changed"),
+# Locate `### Changed` via the same AST walk — keeps heading-text matching
+# consistent with the H2 path (handles surrounding whitespace etc.).
+changed_h3 = next(
+    (
+        n for n in doc.children
+        if isinstance(n, Heading)
+        and n.level == 3
+        and unreleased_idx < h2_line_idx(n) < next_release_idx
+        and heading_text(n).lower() == "changed"
+    ),
     None,
 )
 
-if changed_idx is not None:
+if changed_h3 is not None:
+    changed_idx = h2_line_idx(changed_h3) - unreleased_idx
+    # Insert just before the next H3 (or end-of-block), trimming trailing
+    # blanks so successive runs don't pile blank lines.
     insertion_idx = next(
         (m for m in range(changed_idx + 1, len(block)) if block[m].startswith("### ")),
         len(block),
--- a/scripts/cascade/bump-strategy.sh
+++ b/scripts/cascade/bump-strategy.sh
@@ -358,6 +358,20 @@ if unreleased_node is None:
 # Existing Unreleased section — bound it by the next H2 (or EOF) and either
 # append to the existing `### Changed` subsection or create one.
 unreleased_idx = h2_line_idx(unreleased_node)
+
+# Normalize legacy `## Unreleased` and `## (unreleased)` shapes to the
+# bracketed `## [Unreleased]` form on first touch. The cascade can otherwise
+# leave a consumer in a state the validator (changelog-parser-action) won't
+# parse: that side only accepts `[Unreleased]` (`/^\[\s*unreleased\s*\]$/i`),
+# so bare `Unreleased` would land the bullet correctly here but fail
+# `check-release-log` on the bump PR. Doing the normalization in the
+# rewriter (vs. broadening the parser) keeps the validator strictly aligned
+# with Keep a Changelog convention and lets consumer files migrate
+# incrementally — every bump touches the header once and brings it into the
+# canonical shape.
+if heading_text(unreleased_node).lower() != "[unreleased]":
+    lines[unreleased_idx] = "## [Unreleased]\n"
+
 later_h2 = next(
     (h for h in h2_headings if h2_line_idx(h) > unreleased_idx),
     None,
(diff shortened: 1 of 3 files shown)


scripts/cascade/bump-strategy.sh
```diff
@@ -235,22 +235,32 @@ PY
...
+#    write into a tagged release block. Idempotency check is line-anchored
+#    so 0.1.9 doesn't match 0.1.99.
 python3 <<'PY'
 import os
+import re
 import sys
 
 of_version = os.environ["OF_VERSION"]
 path = "RELEASE.md"
 bullet = f"- Bump openfilter to {of_version}"
```
stwilt, line 253: Using regex to parse Markdown is notoriously brittle and prone to break with subtle format changes (e.g. nested links, escape characters, or trailing spaces). 

We can completely avoid regular expressions and handrolling the parser by utilizing an established Markdown parser library like `mistletoe` or `pyromark`:

### 1. `mistletoe` (Recommended: Pure Python)
Because it is written in pure Python, `mistletoe` has zero native compilation overhead (perfect for dynamic installation in GitHub Actions via `pip install mistletoe`). Most importantly, **every block-level token (like a Heading) records its exact source line number (`node.line_number`)** during parsing!

Here is how clean this becomes:
```python
import mistletoe
from mistletoe.block_token import Heading

def find_unreleased_section(markdown_text: str) -> int | None:
    doc = mistletoe.Document(markdown_text)
    for node in doc.children:
        if isinstance(node, Heading) and node.level == 2:
            content = "".join(child.content for child in node.children if hasattr(child, 'content'))
            clean_title = content.strip().strip("[]()").strip().lower()
            if clean_title == "unreleased":
                return node.line_number - 1
    return None
```

### 2. `pyromark` (Rust-optimized)
`pyromark` is incredibly fast because it is written in Rust. It also tracks exact source ranges. It can be parsed using its event iterator over spans:
```python
import pyromark
# pyromark can yield (event, range) tuples where range is (start_offset, end_offset)
parser = pyromark.Parser(markdown_text)
for event, span in parser.into_offset_iter():
    # we can map these byte/char offsets back to line numbers
    pass
```

Using a formal Markdown AST parser like `mistletoe` would be much cleaner, more standard, and far less prone to breaking over time.

scripts/cascade/bump-strategy.sh
```diff
@@ -268,31 +278,57 @@ if any(line.rstrip() == bullet for line in text.splitlines()):
...
-        "## (unreleased)\n"
-        "\n"
-        "### Changed\n"
-        "\n"
-        f"{bullet}\n"
+if unreleased_idx is None:
+    # No Unreleased section. Inject one before the first released `## v...`
+    # header so the bullet never lands in a tagged block. If there's no `## `
+    # at all, fall back to the legacy "append at end" behavior.
+    first_release_idx = next(
```
stwilt: Great design here. Fallbacking to search for the first release header `## ` (like `## v0.1.27`) and inserting `## [Unreleased]` right before it ensures we never write new bump entries into already-tagged release blocks. If we adopt `mistletoe`, finding the first heading's `line_number` becomes trivial and completely layout-independent!

tests/test_cascade_bump_strategy.py
```diff
@@ -171,6 +171,151 @@ def test_release_md_idempotency_does_not_match_substring_versions(tmp_path: Path
...
 
+# ─────────────────── [Unreleased] anchoring (DT-145 follow-up) ───────────────────
+# bump-strategy.sh previously anchored RELEASE.md inserts on "first `## `"
+# header. When a consumer's file had no `[Unreleased]` section, that landed
+# the bump bullet inside an already-tagged release block. Concrete case:
+# filter-sweet-green-subject-data-aggregator PR #39 wrote a bump under
+# `## v0.1.27 - 2026-04-24`. These tests pin the new anchoring contract:
+# always target `[Unreleased]` (any case / bracket shape), and create one if
+# absent — never touch a released block.
+
```
stwilt, line 186: These new test cases are extremely comprehensive. They cover every single logical pathway introduced in the bump script changes—including legacy unbracketed shapes, orphan bullets in the preamble, and idempotency across repeated runs. Excellent test coverage and documentation here!

stwilt (commented):
LGTM! This is an exceptionally clean and robust fix for DT-145 and the follow-up SGSDA regression. However, let's consider avoiding regex and handrolled parsing for markdown headers. We can use a proper CommonMark parser instead to make this completely bulletproof and future-proof.

lucasmundim (commented):
Thanks for the review! Pushed `9a58689` adopting the mistletoe suggestion.

- `bump-strategy.sh` — parse RELEASE.md once with `mistletoe.Document` and walk `doc.children` for `Heading` nodes. `node.line_number` anchors splices into the source `lines` array. Insertion stays as raw-line manipulation so preamble/blank-line formatting round-trips verbatim (re-serializing through the AST would not preserve those).
- The Unreleased lookup now strips bracket/paren characters from the flattened heading text, so `Unreleased`, `[Unreleased]`, and `(unreleased)` are all recognized by the AST walk rather than a regex on raw lines. Escape sequences and trailing whitespace in headers are handled by the parser instead of the matcher.
- `cascade-on-tag.yaml` — added `mistletoe` to the bump job's `pip install` alongside `packaging tomlkit`.
- Tests — added a `mistletoe` skipif marker; all 23 still pass.

scripts/cascade/bump-strategy.sh
lucasmundim: Done in `9a58689` — swapped the regex/raw-line header detection for a mistletoe AST walk. `mistletoe.Document` gives us `Heading` nodes with `line_number`, which anchor splices into the source line array. Kept insertion as raw-line manipulation so preamble/blank-line formatting round-trips verbatim — re-serializing through the AST would lose those. The Unreleased lookup now strips bracket/paren chars from the flattened heading text rather than matching with a regex, so escape sequences and trailing whitespace in headers are handled by the parser. All 23 tests still pass; added `mistletoe` to the cascade `pip install` and to the test skipif markers.

scripts/cascade/bump-strategy.sh
lucasmundim: Thanks! Confirmed in the refactor — fallback is now `if not h2_headings: append-at-end; else: insert before h2_headings[0]`, with `h2_headings[0].line_number` driving the splice. Layout-independent via the AST, as you suggested.

tests/test_cascade_bump_strategy.py
lucasmundim: Thanks for the kind words!

shingonoide (approved):
Reviewed HEAD 9a58689 — mistletoe adoption is clean; the logic and tests cover the documented failure modes.

## Observations
- `mistletoe` is installed ad-hoc in `.github/workflows/cascade-on-tag.yaml:156` (`pip install packaging tomlkit mistletoe`) but not declared in `pyproject.toml [project.optional-dependencies].dev` alongside `packaging`/`tomlkit`. CI is fine, but a developer running `bump-strategy.sh` locally after `pip install -e ".[dev]"` will hit an ImportError on the heredoc. Small follow-up.

APPROVE.

scripts/cascade/bump-strategy.sh
```diff
@@ -235,22 +235,56 @@ PY
...
+        elif hasattr(child, "children"):
+            out.append(heading_text(child))
+    return "".join(out).strip()
+
+
+def is_unreleased(text: str) -> bool:
+    """Match `Unreleased`, `[Unreleased]`, `(unreleased)`, with surrounding
+    whitespace. Case-insensitive — covers Keep a Changelog convention and the
+    legacy shapes already in the codebase (so a re-run doesn't duplicate)."""
+    return text.lower().strip("[]() \t") == "unreleased"
```
stwilt, line 273: **Contract gap with changelog-parser-action#40.** `is_unreleased()` recognizes `## Unreleased` (unbracketed) and the test at `tests/test_cascade_bump_strategy.py:232` explicitly documents that the unbracketed shape is preserved verbatim. However, `changelog-parser-action#40` `src/ChangelogParser.ts:75` only accepts `[Unreleased]` (regex `/^\[\s*unreleased\s*\]$/i`) — bare `Unreleased` falls through to the version regex and throws `"Could not parse CHANGELOG entry: Unreleased"`. So a consumer with `## Unreleased` will receive the bump bullet here, pass idempotency, but then fail the `check-release-log` gate on the bump PR.

The fix can go in either direction — bump-strategy normalizes unbracketed → `## [Unreleased]` when it rewrites (smallest blast radius), or the parser adds a fallback for bare `unreleased`. The rewriter normalization path is simpler: after locating `unreleased_node`, if `heading_text(unreleased_node)` doesn't start with `[`, rewrite that header line in `lines[unreleased_idx]` to `## [Unreleased]\n` before proceeding. The test at line 232 would need updating to reflect the normalization.

stwilt (changes requested):
Reviewed HEAD 9a58689. The mistletoe adoption is clean and the core regression (SGSDA PR #39 writing bumps into a tagged release block) is fixed. One cross-repo contract gap needs resolution before merge — documented inline.

`mistletoe` is not declared in `pyproject.toml [project.optional-dependencies].dev` (shingonoide's follow-up, already acknowledged). Tracking: follow-up ticket or inline fix before merge.

lucasmundim (commented):
Thanks @stwilt — great cross-repo catch. The contract gap is real: the rewriter recognized `## Unreleased` / `## (unreleased)` as the Unreleased slot, but changelog-parser-action#40 only accepts `[Unreleased]`. A consumer with either legacy shape would land the bump bullet correctly but then fail `check-release-log`.

Pushed `1d2af9f` going with your Option A (normalize in the rewriter). After `unreleased_node` is located, if its heading text isn't `[unreleased]` (case-insensitive) the source line is rewritten to `## [Unreleased]\n` before the bullet is spliced in. Every bump PR now contains both the new bullet and a one-line header migration as a single mechanical change.

**Note this revises the explicit non-normalization choice from `de6f1eb`** (the docstring on `test_release_md_recognizes_unreleased_without_brackets` previously said "we deliberately do not normalize"). That decision predated knowing about the parser-strictness chicken-and-egg you flagged. The two pieces now compose: rewriter normalizes (forward migration), validator gates (enforcement) — keeps changelog-parser-action#40 strictly aligned with Keep a Changelog convention rather than absorbing legacy variants on both sides.

**Tests:** renamed the old test and flipped its assertions to check that bare `## Unreleased\n` is gone after the rewrite; added `test_release_md_normalizes_parenthesized_unreleased_header` for the `## (unreleased)` legacy variant. 24 tests pass (was 23, +1).

scripts/cascade/bump-strategy.sh
lucasmundim: Fixed in `1d2af9f` — went with Option A (normalize in the rewriter) as you recommended.

```python
if heading_text(unreleased_node).lower() != "[unreleased]":
    lines[unreleased_idx] = "## [Unreleased]\n"
```

Placed right after `unreleased_idx` is computed, before the splicing logic. The bump PR thus carries both the new bullet and the header migration in one mechanical change.

This revises the explicit non-normalization decision from `de6f1eb` — that docstring said the legacy shape was "deliberately preserved" because the validator side was meant to enforce. With strict validation on top of preservation, the bump PR fails — exactly the chicken-and-egg you flagged. The right composition is rewriter-normalizes / validator-gates, and that's what `1d2af9f` does.

Tests: renamed `test_release_md_recognizes_unreleased_without_brackets` to `test_release_md_normalizes_legacy_unreleased_header_to_brackets` (assertions flipped — checks the bare `## Unreleased\n` is gone), and added a sibling for the `## (unreleased)` legacy variant. 24 pass.

stwilt (approved):
Blocker resolved. The normalization fix is correct, bounded, idempotent, and tested.

**Criterion 1 — Correct**: scripts/cascade/bump-strategy.sh:372#NQ writes `## [Unreleased]\n` whenever `heading_text(unreleased_node).lower() != "[unreleased]"`, covering bare `## Unreleased`, `## (unreleased)`, `## UNRELEASED`, `##  Unreleased` (double-space — mistletoe normalizes the inline whitespace before `heading_text` sees it).

**Criterion 2 — Idempotent**: The early-exit at [`scripts/cascade/bump-strategy.sh` L299](https://github.com/PlainsightAI/openfilter/blob/1d2af9fb4f1d6d471d12d3c3309190fe8af15ab2/scripts/cascade/bump-strategy.sh#L299) fires on the second run (bullet already present), so normalization is never reached twice. For a re-run without an existing bullet, `heading_text` returns `[Unreleased]` from the already-normalized file and the condition is false — no double rewrite.

**Criterion 3 — Bounded**: Only the `unreleased_node` path in the existing-section branch is touched. The new-section injection paths ([`scripts/cascade/bump-strategy.sh` L284](https://github.com/PlainsightAI/openfilter/blob/1d2af9fb4f1d6d471d12d3c3309190fe8af15ab2/scripts/cascade/bump-strategy.sh#L284) and [`scripts/cascade/bump-strategy.sh` L344](https://github.com/PlainsightAI/openfilter/blob/1d2af9fb4f1d6d471d12d3c3309190fe8af15ab2/scripts/cascade/bump-strategy.sh#L344)) already write `## [Unreleased]` directly.

**Criterion 4 — Tested**: `test_release_md_normalizes_legacy_unreleased_header_to_brackets` replaces the old test that asserted the unbracketed shape was preserved verbatim (the old assertion `"## [Unreleased]" not in rewritten` is gone). A second new test covers `## (unreleased)`. Both pass.

**Criterion 5 — Casing/whitespace tolerance**: `## UNRELEASED`, `##  Unreleased`, `## Unreleased ` all normalize correctly. `## [UNRELEASED]` is deliberately left as-is (`heading_text.lower() == "[unreleased]"`) — changelog-parser-action's `/^\[\s*unreleased\s*\]$/i` accepts it.

**Side effect**: any consumer RELEASE.md currently carrying `## Unreleased` will have its header rewritten on the next cascade bump PR. That's the intended migration path and matches the commit message — no surprise there.

PR #113, opened 2026-06-28
fix(metrics): emit gpu_usage_percent as float for Cloud Monitoring (FILTER-585)

shingonoide:
## 📋 What does this PR do?

Casts the `gpu_usage_percent` system-health metric to `float` so it is always emitted as a floating-point value.

```diff
- self.gpu_usage_percent = gpu.get('gpu0', 0.0)
+ self.gpu_usage_percent = float(gpu.get('gpu0', 0.0))
```

## 🔍 Why is this needed?

NVML reports GPU utilization as an integer, so `gpu_usage_percent` was emitted as an integer whenever a GPU was active (the float default `0.0` only applies when no GPU is present). The Google Cloud Monitoring descriptor for this gauge is type-locked to `DOUBLE`, so every active-GPU point was rejected by the backend with:

```
value type for metric must be DOUBLE, but is INT64
```

and silently dropped (a partial, single-point rejection per export batch). The result is silent gaps in GPU-utilization data. Forcing the value to `float` makes the emitted points match the existing `DOUBLE` descriptor, restoring ingestion with no descriptor migration.

The change is intentionally narrow: the per-device `<filter>_gpuN` counters and the legitimately-integer `gpu_accessible` / `camera_connected` metrics are left untouched, since their descriptors are integer-typed and casting them would cause the opposite rejection.

## 🧪 How was it tested?

- New unit test `tests/test_gpu_usage_percent_float.py` drives the GPU poll path with a mocked integer NVML reading and asserts `gpu_usage_percent` is a `float` equal to `55.0`. It fails on the previous (un-cast) code and passes with this change.
- Existing GPU / telemetry / timing suites remain green (`tests/test_gpu.py`, `tests/test_telemetry.py`, `tests/test_timing_metrics.py`).

## 🔗 Related Issues

FILTER-585

## ✅ Checklist

- [ ] I have added or updated **documentation** as needed

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/113"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785271568&installation_id=128257093&pr_number=113&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F113&signature=4ce146c776e5f1dcb7481c78e2517d2d138097847bb82978a63c7208173bfdfc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


leandrobmarinho (approved):
### Review

**Summary:** Casts `gpu_usage_percent` to `float` in `Metrics.gpu_thread_func` so an active GPU's integer `gpu_util` (gpu.py:71, `gpu_util: int`) no longer gets rejected by the DOUBLE-locked `openfilter_gpu_usage_percent` Cloud Monitoring descriptor (`value type ... must be DOUBLE, but is INT64`).

Verified the fix is correct and minimal: `gpu['gpu0']` derives from the int `gpu_util` when a GPU is present, so the old code emitted an int; the cast forces DOUBLE. Scope is right — the per-device `gpuN` series and the legitimately INT64 descriptors (`gpu_accessible`, `camera_connected`) are untouched, and the downstream `observability/client.py` handles the value generically. No contract break, no reference/mutation concern. 

FYI (non-blocking): `test_gpu_usage_percent_is_float_when_gpu0_absent` doesn't actually guard the regression — the no-gpu0 default `0.0` was already a float on both old and new code, so it passes either way; the gpu0-present test is the one that exercises the fix. No change needed to merge.

lucasmundim (approved):
Reviewed: the float() cast is correct and narrowly scoped to the gpu_usage_percent summary field that maps to Cloud Monitoring's DOUBLE descriptor (openfilter_gpu_usage_percent). Per-device gpuN series and the legitimately-INT64 flags are correctly left untouched. VERSION bump, RELEASE.md format, and the new test all check out. LGTM.

PR #116, opened 2026-07-12
docs: refresh README and split GitHub/PyPI long-description

shingonoide:
Refreshes the public README so it accurately represents OpenFilter at v1.1.2, and splits the
GitHub and PyPI descriptions so each renders correctly on its own platform.

## README refresh (README.md)

- Corrected lead ("a universal ..."), added a concept intro and a Mermaid pipeline diagram.
- Badge row now includes supported Python versions; Requirements section states Python 3.10+
  (tested 3.10-3.13).
- Fixed install instructions: `git+https` URL, real tag `v1.1.2`, and a four-backtick fence so the
  nested pip line renders. Added a docker-run block.
- Corrected the example filename to `example_video.mp4`; added a Built-in filters table.
- Expanded Documentation with absolute GitHub links (so they survive on PyPI).
- Tightened Telemetry & privacy to describe what actually ships today (OpenTelemetry tracing,
  OpenLineage, per-frame metrics; a single Scarf filter-init event; `DO_NOT_TRACK` honored).
- Short, vendor-neutral Ecosystem note (pipelines-controller + plainsight.io).

## GitHub / PyPI split

PyPI cannot render Mermaid and does not resolve relative links, so the rendered project page was
degraded. This adds a generated PyPI variant instead of hand-maintaining two files:

- `README.pypi.md` (generated) - Mermaid block replaced with a plain-ASCII diagram; links absolute.
  Carries a "do not edit" header.
- `scripts/make_pypi_readme.py` - derives `README.pypi.md` from `README.md` (swaps only the diagram
  block). Dependency-free; `--check` mode fails on drift.
- `Makefile` - `pypi-readme` (regenerate) and `check-pypi-readme` (verify) targets.
- `.github/workflows/ci.yaml` - a drift-check job that fails if the two READMEs diverge.
- `pyproject.toml` - `readme = "README.pypi.md"` so the packaged long-description uses the PyPI variant.

## CONTRIBUTING.md

- Documented "How to cut a release (core maintainers only)" and the published Docker Hub images.
- Reworded the main-branch description to state its release-ready invariant.

## Validation

- Local example run passed on a clean venv (`openfilter[all]` editable install): Python Quick Start
  and CLI both served Webvis (HTTP 200) at ~30fps with no traceback.
- `make check-pypi-readme` passes (the two READMEs are in sync).
- Docs + packaging only; no `openfilter/**` source changes, so no RELEASE.md entry is required.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/116"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786464297&installation_id=128257093&pr_number=116&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F116&signature=ad42b771c1a5f50a840f18637234106a52e8b10b879401ada97e763d46bf946c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -55,7 +55,6 @@ jobs:
   # for a plain-text one, since PyPI does not render Mermaid). This guards
   # against the committed PyPI readme drifting from the source README.
   check-pypi-readme:
-    if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     steps:
       - name: Checkout code
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,24 @@ jobs:
       - name: Run tests
         run: make test-all
 
+  # README.pypi.md is generated from README.md (the Mermaid diagram is swapped
+  # for a plain-text one, since PyPI does not render Mermaid). This guards
+  # against the committed PyPI readme drifting from the source README.
+  check-pypi-readme:
+    if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v7
+
+      - name: Set up Python
+        uses: actions/setup-python@v6
+        with:
+          python-version: "3.12"
+
+      - name: Verify README.pypi.md is in sync with README.md
+        run: make check-pypi-readme
+
   # Path-filtered checks for the tag-cascade scripts on PRs touching
   # scripts/cascade/** or the cascade workflow: shellcheck + actionlint +
   # a single-filter discover.sh smoke against filter-frame-dedup. The
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -30,7 +30,6 @@ jobs:
           source-paths: "openfilter/**"
 
   run-unit-tests:
-    if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -99,9 +99,7 @@ If you're unsure whether a change is "big enough" to be worth a PR — it probab
 * Use clear branch naming. For example:
   * Bugfix: `1234-fix-filter-shutdown`
   * Feature: `4567-add-fancy-thing`
-* All contributions should be made against the `main` branch.
-* **Do not submit pull requests directly to `release`** — that branch is managed by core maintainers and reflects the latest stable version of OpenFilter.
-* When a release is ready, core maintainers will merge changes from `main` into `release`.
+* All contributions should be made against the `main` branch, which is the active development branch that releases are tagged from.
 * Include unit tests when appropriate. Run `make test-all` or `pytest` to verify.
 * Keep pull requests focused — unrelated changes should go in separate PRs.
 * Run linters and formatting before pushing (`black`, `ruff`, etc. where applicable).
@@ -336,7 +334,7 @@ Whether you're contributing a quick fix or a large feature, it's important to un
 
 ### 🛠️ Releasing
 
-- Releases are **cut from the `main` branch** — the single branch every published version comes from, kept in a release-ready state at all times.
+- Releases are normally **cut from the `main` branch**, which is kept in a release-ready state at all times. (Urgent hotfixes are the exception and are cut from the latest release tag; see [Hotfixes](#-hotfixes) below.)
 - To create a release from `main`, the following must be true:
   - The `RELEASE.md` file contains an accurate and up-to-date changelog entry for the version.
   - The `VERSION` file matches the version declared in `RELEASE.md`.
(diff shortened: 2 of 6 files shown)


.github/workflows/ci.yaml
```diff
@@ -51,6 +51,24 @@ jobs:
       - name: Run tests
         run: make test-all
 
+  # README.pypi.md is generated from README.md (the Mermaid diagram is swapped
+  # for a plain-text one, since PyPI does not render Mermaid). This guards
+  # against the committed PyPI readme drifting from the source README.
+  check-pypi-readme:
+    if: github.event_name == 'push' || github.event_name == 'pull_request' || github.event_name == 'workflow_dispatch'
```
lucasmundim: The `if:` guard on this new job is redundant: `push`, `pull_request`, and `workflow_dispatch` are exactly the three events this workflow triggers on (the `on:` block at the top of the file), so the condition is always true. It can be dropped, letting the job run whenever the workflow does.

lucasmundim (commented):
One minor note on the new CI job.

.github/workflows/ci.yaml
shingonoide: You are right, removed in 60578dc. The `on:` block triggers on exactly `push`, `pull_request` and `workflow_dispatch`, so the condition matched every run the job could ever see. `run-unit-tests` carried the identical redundant guard, so I removed that one too in 5f7024a.

CONTRIBUTING.md
```diff
@@ -336,7 +336,7 @@ Whether you're contributing a quick fix or a large feature, it's important to un
 
 ### 🛠️ Releasing
 
-- Releases are **cut from the `main` branch**, which always reflects the **latest development version** of OpenFilter.
+- Releases are **cut from the `main` branch** — the single branch every published version comes from, kept in a release-ready state at all times.
```
lucasmundim: The PR description says CONTRIBUTING.md now documents "How to cut a release (core maintainers only)" and the published Docker Hub images, but that section isn't in this branch — the only CONTRIBUTING change is this reworded line. Since the PR also deletes the README's "Releasing" section, the concrete steps to cut a release (bump `VERSION`, add the matching `RELEASE.md` entry, then GitHub Actions → "Create Release" → "Run workflow") are no longer documented anywhere in the repo. The Docker image table survives in the README's "Built-in filters" section, so it's only the maintainer release steps that are missing — could you restore them here under the Release Process section?

lucasmundim: Nit: "the single branch every published version comes from" is at odds with the hotfix flow described further down (hotfixes branch from the latest release tag and are published as one-off releases before merging forward), and with the earlier "Do not submit pull requests directly to `release`" bullet, which references a `release` branch that no longer exists on origin. That staleness predates this PR, but the stronger absolute wording here makes it more visible — worth either softening this line or cleaning up the stale `release`-branch references while you're here.

lucasmundim (commented):
Re-review after 60578dc — the redundant event guard is gone, thanks. I verified the rest of the PR: `README.pypi.md` is in sync (ran `make check-pypi-readme` locally against this branch), the version references match the latest tag (v1.1.2), all documentation links resolve, and the Telemetry section matches what the code actually does. One issue on the release documentation, below.

CONTRIBUTING.md
shingonoide: Added a "How to cut a release (core maintainers only)" section that folds in the steps removed from the README: bump VERSION, add the matching RELEASE.md entry, merge to main, then run the Create Release workflow. It also re-documents the published PyPI package and the Docker Hub images (`plainsightai/openfilter-<name>`, multi-arch amd64 and arm64, tagged with the version and latest). Steps verified against `.github/workflows/create-release.yaml`. Pushed in 0d7e280.

shingonoide: Softened the absolute wording so it no longer contradicts the hotfix flow: releases are normally cut from main, with hotfixes cut from the latest release tag. Also removed the two Conventions bullets referencing a `release` branch that no longer exists on origin, folding the useful part into the main-branch bullet. Pushed in 0d7e280.

lucasmundim (approved):
Re-reviewed after `0d7e280` and `5f7024a`. All three points from my earlier review are resolved:

- The redundant event guard is gone from both `check-pypi-readme` and `run-unit-tests`.
- CONTRIBUTING now documents "How to cut a release (core maintainers only)" — I verified every step against `create-release.yaml` (test matrix, VERSION/RELEASE.md check, tag + GitHub Release, wheel → PyPI under a protected environment, then multi-arch `amd64`/`arm64` filter images tagged `<version>` + `latest`, built only after the PyPI publish).
- The stale `release`-branch wording is cleaned up and the hotfix flow reads consistently now.

LGTM.

PR #118, opened 2026-07-16
Feat/video in seekable replay

trodrigues138:
# 📦 Pull Request

---

## 📋 What does this PR do?

Adds an **opt-in seekable replay API** to `VideoIn` so file-based pipelines can pause, play, step, and seek with frame-accurate sync metadata — without a separate input filter.

When `control_port` is set, `VideoIn`:
- Wires `VideoController` from `filter_subject_data_in` (HTTP API + `/player` + `/video`)
- Performs I-frame–aware seek with forward-read compensation
- Emits **Replay Sync Meta Spec v1** on frames (`frame_index`, `frame_repeat`, `seek_reset`, `seek_frame_index`, …) for downstream sync (e.g. `FilterSubjectDataIn` with `sync_key='sequential'`)
- Shares one controller across multi-source readers so all cameras pause/seek together

Without `control_port`, behavior matches existing `VideoIn` (sources, `bgr`/`sync`/`loop`/`maxfps`/`maxsize`/`resize`, classic meta).

Also includes design doc updates and review follow-ups: dual-prefix env vars (`VIDEO_IN_*` / `FILTER_*`), fail-loud setup if the controller package is missing, post-seek `frame_index == seek_frame_index`, and unit tests for seek/freeze/deque/multi-source.

---

## 🔍 Why is this needed?

Replay pipelines need interactive scrubbing while keeping overlays and stored data locked to the exact video frame. Today that required a separate seekable input path; folding the capability into core `VideoIn` (opt-in) keeps composition simple and avoids drift after seek or while paused.

---

## 🧪 How was it tested?

- Unit tests: `tests/test_filter_video_in.py` — full suite **19 passed** (original `VideoIn` cases + new `TestVideoInSeekableReplay` for env aliases, fail-loud missing package, seek compensation, post-seek index equality, freeze re-emit, deque back-compat, multi-source shared controller)
- Manual / design validation against `docs/design-video-in-seekable-replay.md`

---

## 🔗 Related Issues

N/A (design: `docs/design-video-in-seekable-replay.md`)

---

N/A — API/filter change. Player UI is served by `VideoController` when `control_port` is configured.

---

## ✅ Checklist

- [ ] I have signed all commits in compliance with the DCO (`git commit -s`)

> 📄 **DCO Reminder:**  
>
> ```bash
> ```

---

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/118"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786807190&installation_id=128257093&pr_number=118&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F118&signature=9414321b77eee1d3804a22e62f782e7bb442283c01e2eb721eb0da5a759df4a9"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


leandrobmarinho (changes requested):
### Review

**Summary:** Adds the opt-in seekable-replay API to `VideoIn` (pause/play/step/seek + sync metadata). I re-reviewed this against the design-doc concerns from the earlier design review and ran a careful full-diff pass (concurrency, seek/freeze state machine, metadata, tests, packaging). The design-review follow-ups are genuinely done — HTTP/UI moved into `filter_subject_data_in`, dual-prefix env vars, fail-loud on missing package, documented "Replay Sync Meta Spec v1", and — importantly — the post-seek `frame_index == seek_frame_index` invariant is now correct by construction (`seek_frame_index = frame_n`, warns on OpenCV undershoot). Good.

But three real defects, two of which contradict the PR's own claims, need fixing before merge.

#### Blocking

- **[bug] Multi-source seek moves only one reader — the rest desync.** `consume_seek()` is atomic read-and-clear (design §5 line 119; the test double `_FakeCtrl.consume_seek` does `v, self._seek = self._seek, None`). All N `VideoReader` threads share one controller, each polling `consume_seek()` in its own loop — so whichever thread polls first consumes the target and the other N-1 get `None` and never seek. Only one camera jumps and emits `seek_reset` while `MultiVideoReader.read` keeps zipping now-desynced frames. This directly contradicts design §5 ("all cameras pause/seek together"). `test_multi_source_shares_controller` only asserts wiring, never that both readers actually seek, so it's uncaught. (Freeze is fine — `should_freeze()` is a non-clearing bool read that all readers observe.) Fix needs the controller to fan a seek out to all readers (e.g. per-reader consumption / a seek generation counter), not a single read-and-clear.

- **[bug] A paused reader can't be shut down — the freeze loop never checks `stop_evt`.** In `thread_reader`, the freeze branch emits the frozen frame then `sync_evt.wait()` (sync) or `sleep(ns)` (non-sync) and `continue`s to the top; the only `stop_evt` check is on the normal-read line, never reached while paused. `stop()` sets `stop_evt` + `sync_evt` once, then `join(timeout=2.0)` + `cap.release()`. While paused, `should_freeze()` stays true (`stop()` doesn't touch the controller), so: sync mode blocks forever on the next `sync_evt.wait()` (nothing sets it again) → the 2s join times out and `cap.release()` runs with the thread still alive; non-sync mode spins forever re-emitting frozen frames after stop. Shutting down while paused is a normal action for a scrub UI. Tell: `test_freeze_reemit_keeps_frame_n_fixed` calls `ctrl.set_paused(False)` *before* `vid.stop()`, i.e. the test is written to dodge this exact hang. Fix: test `stop_evt.is_set()` at the top of the freeze branch (and quiesce the controller in `shutdown()` before stopping readers).

- **[bug] Seeking into the EOF region terminates the pipeline instead of clamping.** `_seek_accurate` breaks on EOF with `actual < target` but sets `_frame_n = target`; the following `read_one()` returns `None` on a non-looping file, so `(None, tframe, {})` is emitted, the whole replay-metadata block is skipped (no `seek_reset`), and `process()` exits with 'video ended'. Scrubbing to/near the end kills playback rather than clamping to the last frame. (With infinite `loop` it instead jumps to frame 0.) Either clamp the target to `total_frames-1` here, or handle the EOF seek explicitly.

#### Architectural (strongly recommend addressing)

- **[warning] Layering inversion: openfilter core imports a downstream filter package.** `video_in.py` imports `from filter_subject_data_in.video_controller import VideoController`, but `filter_subject_data_in` is a filter that itself depends on openfilter, and it's declared nowhere in openfilter's `pyproject.toml` (not a dep, not even an optional extra — `video_in` extra is only `boto3`). So core depends on a concrete downstream filter, undeclared, over a large implicit contract (`total_frames/fps/source/local_path`, `start_server(...)`, `consume_seek/should_freeze/on_frame_emitted/stop_server`) defined in another repo with no shared interface and no version pin. Risks: latent circular import (the top-level import runs before `VideoIn` is defined; if the controller imports `VideoIn`/`Filter` it partial-inits → `ImportError` swallowed as "not installed" → misleading setup error), unbounded version skew, and the bare `except ImportError` treating a *broken* package as absent. The seek/freeze gating rightly lives in `VideoReader`, but openfilter should own an abstract `ReplayController` (Protocol/ABC) that `filter_subject_data_in` implements and injects, rather than core naming a concrete filter.

#### Non-blocking

- **[warning]** `read(with_tframe=True)` return shape changed `(image, tframe)` -> `(image, tframe, extras)`; `VideoReader`/`MultiVideoReader` are in `__all__`, so this breaks any external `img, tframe = reader.read(with_tframe=True)`. The design's "backward-compatible" only holds for `with_tframe=False`. Call it out as a breaking change (all in-repo callers are correctly updated, incl. the video-ended `(None, tframe, extras)` case).
- **[warning]** `cap.release()` can race an in-flight `cap.read()` if a stalled stream keeps the reader in `read_one()` past the 2s join (cv2 capture isn't thread-safe). `shutdown()` also stops readers before `ctrl.stop_server()`, leaving a window for a seek/pause to mutate `ctrl` mid-shutdown — stop the server first.
- **[warning]** `_seek_accurate` sets `_frame_n = target` even on EOF-undershoot (stale until the next real read); `seek_ts` uses a `30.0` fps fallback while `src_fps` is reported as `None`, so a consumer can't reproduce `seek_ts = seek_frame_index / src_fps`.
- **[warning] Test gaps:** `VideoIn.process()`'s meta emission (`frame_index`/`total_frames`/`seek_ts`/`seek_reset`) is entirely untested; `test_seek_accurate_*` fully mocks the cap so real cv2 seek is never exercised (and the fixture is a 3-frame clip, so "seek to frame 10" is meaningless); seek-while-paused, seek/freeze-at-EOF, and missing-`_last_replay_img` are uncovered; the `sleep`-loop tests are timing-dependent. Consider a `[replay]`/`[control]` extra in pyproject so the fail-loud `pip install filter-subject-data-in` has a declared, pinned target.

To be fair: the seek-while-playing invariant, seek-while-paused re-emit, `id` vs `frame_index` semantics, and the `_replay_ctrl` wiring order (assigned before `mvreader.start()`) are all correct — I verified those.

shingonoide (commented):
One additive, non-blocking note inline on the control_port validation path. Everything else actionable looks covered by the existing review, so this does not change the overall verdict.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -733,31 +946,102 @@ def setup(self, config):
...
-        self.id              = -1  # frame id
+        self.mvreader    = MultiVideoReader(vsources, [{**default_options, **options} for options in optionss])
+        self.tops_n_vids = tuple(zip(topics, self.mvreader.videos))
+        self.id          = -1  # frame id
         self._camera_connected = 0
 
+        # ---- seekable replay controller (optional) ----
+        self._replay_ctrl = None
+        if config.control_port:
+            if not _HAS_VIDEO_CONTROLLER:
```
shingonoide: Non-blocking: setup() constructs MultiVideoReader on line 951, which opens one cv2.VideoCapture per source, before this control_port guard. When control_port is set but the video controller package is missing, this raise fires before mvreader.start(), so shutdown() never runs and the already-open captures are not released. In the standard one-process-per-filter deployment the OS reclaims the handles when the process exits, so this mainly affects direct setup() callers (tests, config validators), which is why the new test_missing_controller_package_raises releases vid.cap by hand. Consider releasing self.mvreader before re-raising here so the fail-loud path cleans up after itself.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -354,10 +391,52 @@ def thread_reader(self):
...
+                    self._seek_accurate(seek_target)
+                    _just_seeked       = True
+                    _seek_target_frame = seek_target
+
+            # ----------------------------------------------------------------
+            # Replay: freeze-frame (paused) – re-emit the last image unchanged
+            # Exception: if a seek just happened, fall through to read() once
+            # so _last_replay_img is updated to the seek-target frame first.
+            # ----------------------------------------------------------------
+            if ctrl is not None and ctrl.should_freeze() and not _just_seeked:
```
lucasmundim, line 547: **Reader thread never terminates while paused.**

This freeze branch `continue`s at the top of the loop, and `stop_evt` is only checked in the normal-read path below (`image = None if self.stop_evt.is_set() ...`). `stop()` sets `stop_evt`/`sync_evt` once, but the controller is never told to unpause, so a shutdown while paused never exits cleanly:

- **sync mode**: the thread wakes from the first `sync_evt.wait()`, clears it, re-emits, and blocks on the *second* `sync_evt.wait()` forever → `join(timeout=2.0)` times out, the thread leaks, and every shutdown stalls 2s.
- **non-sync mode**: this branch loops forever (`sleep(ns); continue`), spinning and appending frozen frames to a deque nobody drains.

Threads are `daemon=True` so the process can still exit, but any pipeline stop/restart leaks a live (busy, in non-sync) thread. Note `test_freeze_reemit_keeps_frame_n_fixed` already has to call `ctrl.set_paused(False)` before `vid.stop()` to dodge this.

**Fix:** honor `stop_evt` in the freeze branch — e.g. add `if self.stop_evt.is_set(): break` at the top of the loop, or add `and not self.stop_evt.is_set()` to the freeze condition.

VERSION
```diff
@@ -1 +1 @@
-v1.1.2
+v1.1.3
```
lucasmundim: **Version collides with an already-released v1.1.3, and no changelog entry is added.**

`main` is already at **v1.1.3** (the per-GPU metrics fix, #114, dated 2026-07-10 in `RELEASE.md`). Bumping to `v1.1.3` here ships the seekable-replay feature under a version number already published for an unrelated change — this should be **v1.1.4**.

Additionally, this repo maintains `RELEASE.md` per version and every recent release has a changelog block, but this feature adds none. Please add a `## v1.1.4` section describing the `control_port` opt-in replay API.

lucasmundim (commented):
Comprehensive review of the seekable replay API. Overall a clean, well-documented, well-tested opt-in feature — nice work. Two must-fix items before merge (a teardown correctness bug and a version/changelog issue) plus a few nits noted inline.

openfilter/filter_runtime/filters/video_in.py
trodrigues138: Fixed — `setup()` now calls `_release_mvreader_captures()` before both fail-loud `RuntimeError`s in the `control_port` branch, so any `cv2.VideoCapture` already opened by `MultiVideoReader` is released even when `mvreader.start()` / `shutdown()` never run. Covered by `test_missing_controller_package_raises` asserting `cap.isOpened() is False` afterwards.

openfilter/filter_runtime/filters/video_in.py
trodrigues138: Fixed — the freeze branch now checks `stop_evt.is_set()` at the top and breaks (emitting `(None, …)` so consumers wake) instead of looping forever. `shutdown()` also calls `play()` then `stop_server()` before stopping readers, so a scrub-UI stop while paused joins cleanly. Covered by `test_stop_while_paused_does_not_hang` (and `test_freeze_reemit_keeps_frame_n_fixed` no longer needs to unpause before `stop()`).

VERSION
trodrigues138: Fixed — merged `main` (which had already cut v1.1.3 for #114) into this branch, bumped `VERSION` to **v1.1.4**, and added a `RELEASE.md` entry covering the opt-in seekable-replay API plus the `read(with_tframe=True)` 3-tuple breaking change.

trodrigues138:
Thanks for the thorough review @leandrobmarinho — all three blocking bugs plus the architectural recommendation are addressed in the latest commits (`96df502`, `938f98d`).

### Blocking

- **Multi-source seek only moves one reader** — Fixed with `_SeekFanout` / `_SeekFanoutView`: a generation counter fans each seek out so every `VideoReader` observes it exactly once. Covered by `test_seek_fanout_delivers_to_every_reader` and an updated `test_multi_source_shares_controller`.
- **Paused reader can't shut down** — Freeze branch now checks `stop_evt` and breaks; `shutdown()` calls `play()` then `stop_server()` before stopping readers. Covered by `test_stop_while_paused_does_not_hang`.
- **Seek into EOF terminates the pipeline** — `_seek_accurate` clamps to `total_frames-1`; post-seek EOF falls back to last-frame re-emit with `seek_reset`. Covered by `test_seek_past_eof_clamps_to_last_frame`.

### Architectural

- **Layering inversion (core → concrete downstream package)** — Replaced the static `from filter_subject_data_in... import VideoController` with a `ReplayController` Protocol and lazy dotted-path resolution (`replay_controller_class` / `VIDEO_IN_REPLAY_CONTROLLER_CLASS` / `FILTER_REPLAY_CONTROLLER_CLASS`, default unchanged). Core no longer names the concrete package at module scope. `_resolve_replay_controller_class` also distinguishes "target module missing" from "installed but broken" instead of conflating both under a bare `except ImportError`.

### Non-blocking (also done)

- **`read(with_tframe=True)` 3-tuple** — Documented as an explicit breaking change in `VideoReader.read` / `VideoIn` docstrings and `RELEASE.md` v1.1.4. `with_tframe=False` is unaffected.
- **VERSION collision + missing changelog** (also flagged by @lucasmundim) — Merged `main`, bumped to **v1.1.4**, added `RELEASE.md` entry.
- **Capture leak on fail-loud raise** (flagged by @shingonoide) — `_release_mvreader_captures()` runs before both `RuntimeError`s in the `control_port` branch.

Full suite: `tests/test_filter_video_in.py` — **26 passed**. Design doc §8/§9 updated with the resolution table.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -337,8 +467,47 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
+
+        if actual < target_frame:
+            # EOF during forward-read — land on the last decodable index
+            landed = max(actual - 1, 0)
+            logger.warning(
+                'VideoIn seek: requested/clamped target %s undershot at %s; '
+                'clamping to frame %s',
+                target_frame, actual, landed,
+            )
+            if landed > 0 or actual == 0:
```
shingonoide: Narrow edge case: when the forward-read decodes exactly frame 0 and then hits real EOF, `landed` is 0 and `actual > 0`, so this guard is False and the corrective `cap.set(CAP_PROP_POS_FRAMES, landed)` is skipped. The capture is left positioned past the last decodable frame while `_frame_n` is set to 0, so the next read returns ret=False and the pipeline ends (or loop-resets) instead of landing on frame 0 as documented. Only reachable when the reported frame count overstates the truly decodable frames (short or degenerate files), and no test exercises this branch with a failing read.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -733,31 +1155,157 @@ def setup(self, config):
...
+                )
+
+            # Use the first video as the canonical source for the controller
+            primary_vid  = self.mvreader.videos[0]
+            primary_src  = config.sources[0].source
+            local_path   = primary_src.replace('file://', '') if primary_src.startswith('file://') else None
+            if local_path and not os.path.isfile(local_path):
+                local_path = None
+
+            ctrl = controller_cls(
```
shingonoide: Follow-up to the earlier capture-release note: `_release_mvreader_captures()` now guards the two import-resolution raises above, but the `controller_cls(...)` and `ctrl.start_server(...)` calls here in the same control_port branch are not wrapped, so if either raises the captures opened earlier in setup() still leak. Same bounded blast radius as before (direct setup() callers).

shingonoide (commented):
Re-checked at 6c6260c8. The earlier capture-release note is resolved for the two import-failure raises, and the seek-fanout, stop-while-paused, and EOF-clamp reworks are tested and look right. Two net-new items from this pass, both narrow, noted inline. Leaving the gate to the standing review.

openfilter/filter_runtime/filters/video_in.py
trodrigues138: Fixed in `b62ba87` — the corrective `cap.set(CAP_PROP_POS_FRAMES, landed)` now always runs after an EOF undershoot, including the `landed == 0 and actual > 0` case (decode frame 0, then EOF). Covered by `test_seek_undershoot_landing_on_frame_zero_repositions_cap`.

openfilter/filter_runtime/filters/video_in.py
trodrigues138: Fixed in `b62ba87` — `controller_cls(...)` and `ctrl.start_server(...)` are now wrapped in the same try/except that calls `_release_mvreader_captures()` before re-raising, so a bind/construct failure no longer leaks the captures opened earlier in `setup()`. Covered by `test_start_server_failure_releases_captures`.

trodrigues138:
@shingonoide — both follow-ups from your Jul-check at `6c6260c` are fixed in `b62ba87` (replies were on the inline threads, which GitHub marked outdated after the DCO rewrite / force-push, so repeating here for visibility):

1. **EOF undershoot with `landed == 0`** — `_seek_accurate` always calls `cap.set(CAP_PROP_POS_FRAMES, landed)` after an undershoot, including decode-frame-0-then-EOF. Covered by `test_seek_undershoot_landing_on_frame_zero_repositions_cap`.
2. **`controller_cls` / `start_server` capture leak** — those calls are wrapped so `_release_mvreader_captures()` runs before re-raising. Covered by `test_start_server_failure_releases_captures`.

Suite: `tests/test_filter_video_in.py` — **29 passed**.

leandrobmarinho (approved):
### Review (re-approve at b62ba87 — clears my change request)

Excellent turnaround. Every blocking item from my review is genuinely fixed, and I verified each against the code, not just the replies:

- **Multi-source seek desync (my #1):** the new `_SeekFanout` / `_SeekFanoutView` is correct. `consume_for(reader_id)` polls the inner controller once, bumps a generation counter on a new target, and each reader returns that target while `last_gen[reader_id] < gen` — traced it: with two readers and a seek to N, the first reader gets N from `consume_seek()` and the second gets N from the recorded `_target` (not `None`), so **both** seek; on the next loop `last == gen` so neither re-seeks; back-to-back seeks correctly collapse all readers onto the latest target. `should_freeze`/`on_frame_emitted` delegate to the shared inner, so freeze still fans to all.
- **Paused reader can't shut down (my #2):** the freeze branch now checks `stop_evt.is_set()` at the top and `break`s (emitting `(None, …)` so consumers wake); `shutdown()` also `play()`s then `stop_server()`s before stopping readers, so no command can re-arm the freeze mid-shutdown.
- **Seek into EOF terminates the pipeline (my #3):** post-seek `image is None` now clamps via `_seek_accurate(total_frames - 1)` and re-reads (falling back to the last decoded image with `seek_reset` semantics), so scrubbing to the end no longer kills playback. shingonoide's `landed == 0 && EOF` edge is covered too.
- **Layering inversion (my architectural note):** resolved exactly as suggested — openfilter now owns a `@runtime_checkable ReplayController` Protocol and resolves the concrete controller lazily via `_resolve_replay_controller_class` (no top-level import of the downstream package, no circular-import exposure).

Also good: captures are released before every fail-loud `RuntimeError` and around the `controller_cls(...)`/`start_server(...)` construction (no leaked `cv2.VideoCapture`), the `read(with_tframe=True)` 3-tuple change is documented, and the version collision is fixed (`v1.1.4` + RELEASE.md entry). No remaining blocking issues.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -354,10 +525,83 @@ def thread_reader(self):
...
             image  = None if self.stop_evt.is_set() else self.read_one()
             tframe = time_ns()
 
+            # After a seek near EOF, OpenCV may return no frame — clamp by
+            # re-emitting the last decoded image (or a re-seek to last frame)
+            # so scrubbing to the end does not kill the pipeline.
+            if image is None and _just_seeked and not self.stop_evt.is_set():
+                if self._total_frames > 0:
+                    self._seek_accurate(self._total_frames - 1)
+                    image = self.read_one()
```
shingonoide: Net-new, non-blocking: this retry calls read_one() a second time in the same iteration. self.loop uses 0 for both the infinite-loop and the exhausted play-once states, and the first read_one() that returned None already decremented a loop=False source to 0. If this retry read also fails to decode at total_frames-1, read_one() falls through to the reopen path and returns frame 0 instead of None, so the pipeline restarts from the start and emits seek_reset with seek_frame_index=0 rather than ending. We could reproduce the sentinel collision at the read_one() level but not end to end against a real decoder, so this is a heads-up rather than a blocker. A cheap guard would distinguish the exhausted sentinel from the infinite one, or avoid the second read_one() call.

shingonoide (commented):
Re-checked at b62ba87. Both items from our last pass are resolved and covered by new tests: the seek-undershoot reposition now always runs on the landed==0 EOF case, and the controller construct and start_server path releases the already-opened captures before re-raising. One net-new, non-blocking observation is inline; nothing blocking on our end.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: **Sync-mode deadlock in the post-seek EOF retry — the second `read_one()` blocks on a `sync_evt` nobody will set.**

In sync mode each `read_one()` consumes exactly one `sync_evt.set()` (posted by the consumer's last `read()` pop). When this clamp branch runs, the first `read_one()` on line 575 already consumed it on its EOF exit path (`wait()` before `return None`). This retry then decodes a frame and enters `wait()` -> `sync_evt.wait()` with the event cleared — but the consumer is parked in `MultiVideoReader.read()` waiting for `frame_available`, and it only sets `sync_evt` after popping a frame this thread hasn't appended yet. Reader and consumer wait on each other; the pipeline hangs until `stop()`.

It also compounds the open loop-counter issue above: once the first EOF read decrements `self.loop` from 1 to 0, a play-once video is in the "infinite loop" state — the next natural EOF reopens the file from frame 0 and `VideoIn` never exits.

Both stem from reusing `read_one()` (fps pacing + loop/reopen semantics) for a replay-internal reposition read. A raw `ret, image = self.cap.read()` here (plus the existing `_frame_n` bookkeeping) sidesteps the deadlock, the loop-counter corruption, and the frame-0 restart in one go.

Repro requires a file whose `CAP_PROP_FRAME_COUNT` overstates decodable frames (truncated/VFR) plus a seek near EOF in `!sync` mode — `test_seek_past_eof_clamps_to_last_frame` doesn't hit it because the test file's metadata is exact, so the first read succeeds.

lucasmundim (commented):
Re-reviewed: all five previously flagged issues are verified fixed on `b62ba87` (full suite passes locally, 28 passed / 1 skipped). One new finding below in the post-seek EOF clamp path.

openfilter/filter_runtime/filters/video_in.py
trodrigues138: Fixed in `f4493c1` — after a seek we no longer call `read_one()` at all for decode. The post-seek path paces with the same sync/fps wait, then uses raw `cap.read()`, and the EOF clamp retry is also a raw `cap.read()` after re-seeking to `total_frames-1`. That avoids the play-once `loop` 1→0 flip (and the subsequent reopen-from-0 that looked like infinite loop). Covered by `test_post_seek_eof_clamp_uses_raw_read_not_read_one`.

trodrigues138: Fixed in `f4493c1` — agreed; a second `read_one()` after the EOF `wait()` would deadlock sync mode. Post-seek decode (including the EOF clamp retry) now uses raw `cap.read()` only, after an explicit sync/fps pace that clears `sync_evt` once. The next iteration then blocks in `read_one()` until the consumer pops the `seek_reset` frame, so we neither deadlock nor race ahead and overwrite the deque. Regression: `test_post_seek_eof_clamp_uses_raw_read_not_read_one`.

RELEASE.md
```diff
@@ -4,13 +4,15 @@ OpenFilter Library release notes
 
 ## [Unreleased]
```
lucasmundim, line 5: **The v1.1.4 section is missing the `[Unreleased]` entries from `main`.**

This branch predates `1eae36e` on `main`, which added two entries under `[Unreleased]`: the **CVE-2026-40962** dependency bumps (`opencv-python-headless` ~=5.0.0.93, `av` ~=17.1.0 / ffmpeg 8.1) and the dependency-floor refresh (OTel ~=1.44.0, pydantic ~=2.13.4, fastapi ~=0.139.2, etc.).

Since this PR cuts **v1.1.4**, those changes ship with it. Please merge `main` and move both entries out of `[Unreleased]` into the `## v1.1.4` section so the changelog matches what the release actually contains — this also resolves the merge conflict both branches will otherwise have at the top of this file.

lucasmundim (commented):
Re-checked at `f4493c1` — the post-seek raw `cap.read()` rework resolves the sync-mode deadlock and the play-once loop corruption; full suite passes locally (30 passed, 1 skipped). One remaining item on the changelog, inline.

RELEASE.md
trodrigues138: Fixed in the latest commit — moved the CVE-2026-40962 dependency bumps and the dependency-floor refresh out of `[Unreleased]` into `## v1.1.4`, since those changes from `main` (`1eae36e`) ship with this release cut. `[Unreleased]` is empty again.

lucasmundim (approved):
Approving at `29159a18`. Verified since my last pass:

- The `f4493c1` rework resolves my sync-mode deadlock finding: post-seek decode (including the EOF clamp retry) uses raw `cap.read()` with a single sync/fps pace, so exactly one `sync_evt` token is consumed per emitted frame in every branch, and the play-once `loop` counter can no longer be flipped 1→0 by a post-seek EOF. Full suite passes locally on the PR head (30 passed, 1 skipped — the optional `filter_subject_data_in` resolve test).
- The changelog item from my previous comment is resolved: `main` is merged and the CVE-2026-40962 / dependency-floor entries now live under `## v1.1.4`, with `[Unreleased]` empty. The fold commit touches only `RELEASE.md`.
- Full CI matrix is green on this head (unit tests 3.10–3.13, grype, DCO, release-log and readme checks).

shingonoide (commented):
Thanks for the careful implementation and the regression tests, the seek and replay logic itself reads well. Before this lands I would like to step back to the design and placement, because openfilter is an open-source project and a change of this size to a core filter should have an agreed design behind it first.

As implemented, this adds an HTTP control server, an HTML player UI, and a public replay-control API to VideoIn, and it changes the public VideoReader.read(with_tframe=True) and MultiVideoReader.read(...) signature to a 3-tuple, which is a breaking change for every external consumer of the library.

My suggestion is to keep this out of the openfilter core entirely and ship it as its own filter in a separate repo, the same way filter_subject_data_in already lives as its own filter, composing with VideoIn rather than modifying it. That keeps VideoIn's public surface and contract stable, keeps the control server and player UI opt-in, and avoids expanding the open-source core with a feature a specific replay pipeline needs.

Could you confirm whether this direction was agreed in a design doc first? I was not able to find one that sanctions adding this to the core filter. I would like to settle the placement in the design doc before it becomes a committed public contract, and I am glad to work through it there. The implementation itself is in good shape and would carry over cleanly to a standalone filter.

lucasmundim (commented):
I agree with @shingonoide on placement. Interactive replay is a niche capability — most `VideoIn` deployments will never set `control_port` — but as written the cost is paid in core by everyone: a breaking change to `VideoReader.read(with_tframe=True)` / `MultiVideoReader.read(...)` (both in `__all__`), four new config fields, and the seek/freeze state machine inside our most-used source filter.

To be clear, my earlier approval covered implementation correctness, which I still stand by — the seek/freeze logic is solid and well-tested. The placement question is separate, and I think shingonoide is right that it should be settled in the design doc before this becomes a committed public contract.

One nuance for that discussion: this can't be a purely downstream filter, since pause/seek must gate frame production inside the decode loop. But it can be a separate *source* filter — e.g. a `VideoReplayIn` in its own repo that extends `VideoIn`, the same way `filter_subject_data_in` already lives as its own filter. If subclassing needs extension points in `VideoReader`, a small hooks-only PR to core would be a much better trade than landing the full feature here, and the existing implementation would carry over largely intact.

shingonoide:
Quick heads-up before this lands, since our changes collide.

We are cutting v1.2.0 shortly. Worth knowing first that v1.1.3 was never actually released: there is no v1.1.3 tag on the remote and the latest published version is still 1.1.2, even though main carries a VERSION of v1.1.3 and a dated v1.1.3 changelog section. The v1.2.0 cut absorbs that section, since those fixes reach users for the first time in 1.2.0.

The effect here is that once v1.2.0 exists, your v1.1.4 heading sits above it in RELEASE.md and the changelog parser throws on version ordering, which is a hard `check-release-log` failure. VERSION would also be moving backwards.

Nothing wrong with the change itself, it just needs a rebase and a renumber once our cut is in. Happy to ping you the moment it lands.

shingonoide (changes requested):
Hi Tales, thanks for the PR! Before we look at the code implementation, we need to address a few architectural governance points for the core `openfilter` library:

1. **Design Sanction**: There doesn't appear to be a prior agreed design doc or issue linked for this direction (the design docs are authored within this PR). We usually require an approved spike/ticket before making core architectural changes.
2. **Component Home**: Modifying the core `VideoIn` filter to include a domain-specific "seekable replay API" and wiring a downstream dependency (`filter_subject_data_in`) goes against the principle of keeping core filters generic. A new capability like this typically belongs in a separate filter or wrapper within its own repository, rather than altering the core `VideoIn`.
3. **Breaking API**: The PR introduces a breaking change to a public API (`read(with_tframe=True)` returning a 3-tuple), which we cannot silently merge into core without a major version plan.

Could we take a step back and discuss the design? We should explore how to achieve the replay sync by composing a separate filter downstream of `VideoIn`, keeping the core library stable and dependency-free.

PR #121, opened 2026-07-17
chore(deps): bump opencv to 5.0.0.93 (CVE-2026-40962) + refresh deps

lucasmundim:
## What

Bumps `opencv-python-headless` to `~=5.0.0.93` and refreshes the rest of the dependency floors to current releases.

## Why — CVE-2026-40962

`opencv-python-headless` bundled a stale ffmpeg vulnerable to **CVE-2026-40962** (integer overflow / OOB write in ffmpeg's MOV/CENC demuxer, fixed upstream in **ffmpeg >= 8.1**). The `5.0.0.93` linux x86_64 wheel now bundles **ffmpeg 8.1** — verified by inspecting the wheel's bundled shared libs (`libavcodec.so.62.28.101`, `libavutil.so.60.26.101`). Since the CI security scan runs on `ubuntu-latest` / x86_64, the vulnerable code is no longer present in the scanned environment.

The shared security-scan workflow's default suppression of this CVE is removed in PlainsightAI/gh-actions-public#30.

## Other bumps

OpenTelemetry stack → `~=1.44.0` (semconv `0.65b0`, GCP exporters `1.12.0a0`), `pydantic~=2.13.4`, `fastapi~=0.139.2`, `uvicorn~=0.51.0`, `python-multipart~=0.0.32`, `openlineage-python~=1.51.0`, `requests~=2.34.2`, `google-cloud-storage~=3.13.0`, `google-cloud-monitoring~=2.31.0`, and dev tooling (`setuptools~=83.0.0`, `pytest~=9.1.1`, `build~=1.5.0`, `docker~=7.2.0`, `wheel~=0.47.0`, `jq~=1.12.0`, `pytest-cov~=7.1.0`).

## Deliberately held back

- **`numpy`** stays at `~=2.2.6` — 2.5.x requires Python **>=3.12**.
- **`av`** stays at `~=16.1.0` — 18.x requires Python **>=3.11**.

Both would drop Python 3.10/3.11, which OpenFilter still supports (`requires-python = ">=3.10, <3.14"`, CI matrix 3.10–3.13).

## Validation

- Fresh `pip install -e ".[all,dev]"` resolves with **no conflicts**.
- Full non-slow test suite passes (`-m "not slow"`), including the video-encode CLI path (exercises `av`) and the opencv 5 pipeline.

## Release

`VERSION` is intentionally left unbumped — this follows the repo convention where feature/chore PRs land under `[Unreleased]` and a dedicated release PR cuts the version once everything is merged.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/121"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786845891&installation_id=128257093&pr_number=121&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F121&signature=421420c33d29466c9718e6c6a973fa85f14d6318b4cca964cf4ce190c5528c68"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


shingonoide (approved):
Clean CVE remediation. opencv is pinned to 5.0.0.93 in pyproject.toml (the manifest that gates the wheel), the decode and frame-conversion paths use only stable cv2 API and are covered by real tests that passed on 5.0.0.93 across Python 3.10 to 3.13, and CI including grype is green. One non-blocking note: uv.lock was not regenerated, so it still pins the old opencv-python-headless 4.13.0.92 (and the pre-refresh versions of the other deps). CI installs from pyproject so tests are unaffected, but since this is the CVE fix it would be good to run `uv lock` so the committed lockfile stops pinning the vulnerable build. Approving as-is.

leandrobmarinho (approved):
### Review

**Summary:** Bumps `opencv-python-headless` 4.13 -> ~=5.0.0.93 (the wheel now bundles ffmpeg 8.1, closing CVE-2026-40962) and refreshes the rest of the dependency floors, holding numpy/av back to keep Python 3.10/3.11 support.

Looks good. No blocking issues found. The CVE rationale checks out (ffmpeg 8.1 is where the MOV/CENC demuxer overflow is fixed, and the linux x86_64 wheel bundles it), and holding numpy at `~=2.2.6` / av at `~=16.1.0` is the right call given `requires-python >=3.10` — a good catch to not silently drop 3.10/3.11.

The one thing worth an explicit eye is that this is a **major opencv bump (4 -> 5)**, not just a floor refresh — OpenCV 5 has real API changes. That risk is carried by CI (the 3.10-3.13 matrix runs the suite against opencv 5), so as long as the matrix is green this is validated; just flagging it's a bigger change than a security floor bump and rides entirely on test coverage of the cv2 call sites. Pairs with gh-actions-public#30 to un-ignore the CVE once consumers are on 5.x.

RELEASE.md
```diff
@@ -4,6 +4,14 @@ OpenFilter Library release notes
 
 ## [Unreleased]
 
+### Security
+
+- **CVE-2026-40962**: Bumped `opencv-python-headless` from `~=4.13.0` to `~=5.0.0.93`. The 5.0.0.93 linux x86_64 wheel bundles ffmpeg 8.1 (`libavcodec.so.62.28.101`), which fixes the ffmpeg MOV/CENC demuxer flaw (fixed upstream in ffmpeg >= 8.1). The shared security-scan workflow no longer suppresses this CVE by default (see PlainsightAI/openfilter#84).
```
shingonoide: This points at #84 (the merged shared-scan adoption); the un-ignore is gh-actions-public#30, still open, so "no longer suppresses this CVE by default" is not true yet.

shingonoide (commented):
The opencv-python-headless 4.x to 5.0.0.93 bump is correct: the manylinux x86_64 wheel bundles ffmpeg 8.1.1, which clears CVE-2026-40962 in opencv's vendored ffmpeg, and no OpenCV 5 API break affects the package (shipped encoding uses PyAV, not `cv2.VideoWriter`). Two things to flag before this is treated as a complete CVE remediation:

1. PyAV (`av~=16.1.0`, installed via `.[all]`) bundles its own ffmpeg 8.0.1, which still carries the vulnerable MOV/CENC demuxer. That library lands in the same venv the security scan catalogs, so removing the default grype ignore for CVE-2026-40962 (gh-actions-public#30) after only this bump will likely re-flag the CVE from PyAV. Suggested sequencing: land this bump and a PyAV build on ffmpeg >= 8.1 (or exclude `av` from the scan) before #30 drops the ignore.
2. Minor: the RELEASE.md Security entry references openfilter#84 and reads as already-done, but the PR that actually drops the suppression is gh-actions-public#30, still open. Worth fixing the reference and the tense.

RELEASE.md
lucasmundim: Fixed — reworded the entry to reference gh-actions-public#30 (still open) as where the suppression is actually dropped, and added the PyAV caveat you flagged: `av` bundles its own ffmpeg 8.0.1, so it needs ffmpeg >= 8.1 (or exclusion from the scan) before #30 drops the ignore, or the CVE re-flags from PyAV. This PR stays a valid fix for opencv's vendored ffmpeg; the PyAV bump + #30 are the sequenced follow-ups.

shingonoide (commented):
Thanks for the RELEASE.md fix, the PR reference and the PyAV caveat both read correctly now. The opencv 5.0.0.93 bump itself is the right fix for CVE-2026-40962 (its bundled ffmpeg is 8.1.1, outside the vulnerable <8.1 range).

One correction on the note about holding `av`: `av~=17.1.0` (published 2026-06-07) still supports Python 3.10 and ships ffmpeg 8.1.1, so the PyAV-bundled-ffmpeg exposure can be closed now without dropping 3.10. Only `av` 18.0.0 drops 3.10. Worth bumping `av` to `~=17.1.0` here (or in a fast follow) so removing the gh-actions-public#30 suppression is not blocked on PyAV, and updating the RELEASE.md rationale to match.

lucasmundim (commented):
Thanks, good catch — and agreed. This is already addressed in the latest commit (`d901d75`): bumped `av` from `~=16.1.0` to `~=17.1.0`, which ships ffmpeg 8.1.1 (`libavcodec.so.62.28.101`) and still provides a Python 3.10 wheel — only `av` 18.0.0 drops 3.10. So the PyAV-bundled ffmpeg is now on the fixed 8.1.x line alongside opencv 5.0.0.93, and removing the `gh-actions-public#30` suppression is no longer blocked on PyAV.

Updated the RELEASE.md rationale accordingly. `video_out.py` catches no `av`-specific exceptions, so the av 16→17 API changes don't touch filter code; the full non-slow suite passes on Python 3.10–3.13 and CI is green.

PR #128, opened 2026-07-23
feat(video_in): stamp decoder pts_s + src_frame into frame meta (file sources)

leandrobmarinho:
## Problem

Frame meta from `VideoIn` carries `id` (a delivery counter), `ts` (wall-clock) and `src_fps` — but **no video offset**. Consumers that need jump-to-frame must reconstruct time as `id / assumed_fps`, which breaks in practice:

- the counter's effective rate **varies with the consuming chain** (measured on the same 12.5 s clip: ~24 ids/s through one filter chain, ~30 ids/s through another);
- `src_fps` misreports on some files (reports 30 for 24 fps mp4s);
- `ts` is wall-clock, not stream position.

Hit while indexing caption windows for video evidence search (internal-chicagopd): computed offsets landed past the end of the clip.

## Change

For **file sources**, `VideoReader._cap_read()` records the decoder position of each delivered frame:

- `src_frame` — `CAP_PROP_POS_FRAMES` sampled before the read (0-based, exact);
- `pts_s` — `src_frame / container_fps`. Exact for CFR sources and immune to two observed `CAP_PROP_POS_MSEC` bugs (B-frame presentation reordering; reported as 0 on some container/backend combos — both reproduced by the repo's own test clip). `POS_MSEC` remains only as the fallback for VFR containers that report no frame rate.

Values ride the reader's deque tuple (same thread as the read — no races) and are emitted in `meta` as `pts_s`/`src_frame`. Stream/webcam sources are unchanged (no meaningful decoder position): keys are simply absent.

`VideoReader.read(with_tframe=True)` now returns a 4-tuple `(image, tframe, pts_ms, frame_no)` — flagged as the one internal-API arity change, callers inside the repo are updated.

## Tests

New `test_pts`: 3-frame 30 fps clip → `src_frame == [0, 1, 2]`, `pts_s` floats spaced ~1/30 s. Full suite: **11 passed**.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/128"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787413715&installation_model_id=20112&pr_number=128&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F128&signature=ac77ae42019cd9efb4f22c3d1f0496f6e246a197a9b34187cc61835b9e129819"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


openfilter/filter_runtime/filters/video_in.py
```diff
@@ -419,7 +453,7 @@ def stopped(self) -> bool:
     def frame_available(self) -> bool:
         return bool(self.deque)
 
-    def read(self, with_tframe=False):  # -> np.ndarray | tuple[np.ndarray, int] | None
+    def read(self, with_tframe=False):  # -> np.ndarray | tuple[np.ndarray, int, float | None, int | None] | None
```
shingonoide: read(with_tframe=True) changes from a 2-tuple to a 4-tuple here. VideoReader and MultiVideoReader are exported in __all__ of the published openfilter package, so an external caller doing `img, tframe = reader.read(with_tframe=True)` breaks. Please state the back-compat position: either an accepted break with a RELEASE.md callout, or a shape that absorbs future additions without another break. The extensible extras dict that PR #118 introduces would do that, which is also why the two PRs should settle on one shape before either ships.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -259,6 +262,37 @@ def stop(self):  # idempotent and safe to call whenever
...
+
+        if ret and image is not None:
+            self.frame_no = frame_no = int(pos_frame)
+
+            # Primary: frame index over the container frame rate - exact for CFR
+            # sources and immune to the B-frame reordering and report-zero bugs
+            # POS_MSEC shows on several backends. POS_MSEC only covers the rare
+            # VFR container that reports no frame rate at all.
+            if self.native_fps:
+                self.pts_ms = frame_no / self.native_fps * 1000
```
shingonoide: pts_ms (and thus the emitted pts_s) is computed as frame_no / native_fps on this branch, which is taken whenever the container reports any frame rate. A VFR file that reports an average FPS lands here and gets a drifting timestamp rather than the POS_MSEC path; the comment just above concedes POS_MSEC only covers containers that report no rate at all. Since jump-to-frame is exactly where drift hurts, please state the intended behavior for fps-reporting VFR sources.

shingonoide (changes requested):
Thanks for putting this together. The need here is real and well argued, and it has prior art: the CPD specs (internal-chicagopd PR 2, spec 01) already establish that a decoder-derived video offset is required for jump-to-frame. What I could not find is any agreed design for implementing that need as a change to openfilter core's public surface, and that is what I have to ask for before a full review.

Main ask: a short design note or ticket, linked from this PR, covering:

1. The frame-meta contract addition itself: key names (pts_s, src_frame), exact semantics, and which source types carry them.
2. Coordination with the Seekable Replay work: open PR #118 extends the very same VideoReader deque tuple and read(with_tframe=True) return, but with a different shape (an extensible (image, tframe, extras) dict carrying frame_index / total_frames) and its own design doc. Both PRs edit the same lines (a merge of the two branches conflicts in video_in.py in four places), and semantically we would end up with two names for the same concept (src_frame vs frame_index). The two efforts need to agree on one tuple shape and one naming scheme before either contract ships.
3. The VFR assumption: pts_ms (and thus the emitted pts_s) is computed as src_frame / native_fps whenever the container reports any frame rate. VFR files commonly report an average FPS, so they take this path and get a drifting timestamp instead of the POS_MSEC fallback (the code comment concedes the fallback only covers containers reporting no rate at all). Since jump-to-frame is exactly where drift hurts, the intended behavior for fps-reporting VFR files should be stated.
4. The VideoReader.read() arity change: read(with_tframe=True) goes from a 2-tuple to a 4-tuple. VideoReader is exported in __all__ of a pip-published package, and PR #118 itself treats this same surface as a breaking-change boundary. The back-compat position should be stated (accepted break with a release-note callout, or a compatible shape such as the extras dict #118 already uses, which would absorb future additions without another break).

The note should also state explicitly whether this belongs in core versus a companion filter. My own read is that it does belong in VideoIn, because the decoder position at read time cannot be reconstructed downstream, but that should be a recorded decision rather than an implicit one.

Also needed in any shape this lands:

- check-release-log is failing: it needs a RELEASE.md entry (a meta-contract addition plus a public return-shape change warrant one regardless).
- DCO is failing: the commit needs a Signed-off-by (git commit --amend -s and force-push).
- Test coverage: the new test_pts only exercises the native-fps branch. The POS_MSEC fallback and the pts_ms = None branch are untested, nothing asserts a non-file source's meta stays unchanged, and nothing pins the new 4-tuple at the VideoReader level. Please cover those.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: One more coordination point for the merge with #118: that branch's post-seek path decodes with a raw `cap.read()` (f4493c1 on `feat/video-in-seekable-replay`), which bypasses the `_cap_read()` wrapper introduced here. Whichever shape wins (2/3/4-tuple or the extras dict), if the replay path doesn't route through the same wrapper, frames delivered after a seek will silently carry no `pts_s`/`src_frame`. Worth settling alongside the shape question so the meta contract also holds for seeked frames.

leandrobmarinho: Adopted in 40447c3: `read(with_tframe=True)` now returns the same extensible 3-tuple `(image, tframe, extras)` that #118 introduces, with `extras` as `{}` for non-file sources and `{'frame_n': int, 'pts_s': float}` (pts optional) for files, using #118's `frame_n` key name so a merge of the two branches unions the dict instead of conflicting. Back-compat position is now explicit: an accepted break with a RELEASE.md Breaking Changes callout worded to match #118's, so after both land external callers see one break, not two; `with_tframe=False` is unaffected. The new `test_pts_reader_tuple` pins the shape at the VideoReader level.

openfilter/filter_runtime/filters/video_in.py
leandrobmarinho: Stated now, in the code comment (40447c3) and in docs/design-video-in-pts.md section 4 (731605b). Intended behavior: every fps-reporting source is deliberately treated as CFR. A true-VFR file reporting an average rate gets a nominal timeline (frame_n / avg_fps) that can drift from the container pts, but it is deterministic and exact under the inverse mapping (offset * avg_fps) consumers use to seek by frame, so jump-to-frame round-trips to the same frame, which is what this contract exists for. Switching those files to POS_MSEC is not a safe upgrade: it hits the B-frame reordering and report-zero bugs (both reproduced by the repo's own test clip) with no runtime way to tell a true pts from a lie. POS_MSEC stays confined to containers reporting no rate at all, and only while it looks sane; otherwise pts_s is now omitted rather than emitted wrong, while src_frame stays present and exact (new tests cover both fallback branches). Consumers needing true per-frame VFR timestamps should demux container pts out-of-band (e.g. PyAV); recorded as out of scope for VideoIn.

openfilter/filter_runtime/filters/video_in.py
leandrobmarinho: Agreed, and made robust on this side in 40447c3: `_cap_read()` is the single choke point for reads off `self.cap` on this branch (the only other `cap.read()` is the static `get_info` probe on a temporary capture), and its docstring now says so explicitly, so any new read path added here cannot silently skip position capture. Merge-coordination plan, recorded in docs/design-video-in-pts.md section 3: whoever merges second routes #118's post-seek forward-read compensation through `_cap_read()`. The compensation frames are discarded so the extra two `cap.get` calls cost nothing, and the first delivered post-seek frame then carries correct pts_s/src_frame. Since both branches now share the `(image, tframe, extras)` shape and the `frame_n` key (computed to the same value: #118 reads POS_FRAMES - 1 after the read, this PR reads POS_FRAMES before it), the deque and read() conflicts resolve to identical code and extras merge by key union.

leandrobmarinho:
Review feedback addressed in 40447c3 + 731605b. Item by item:

**Design note (main ask)**: added as `docs/design-video-in-pts.md`, covering all four points plus the core-vs-companion decision (recorded: it belongs in VideoIn, the decoder position at read time cannot be reconstructed downstream).

**1. Meta contract**: key names, exact semantics and carrying source types are now documented in the design note, in the VideoIn class docstring (new "Emitted frame meta" section) and in RELEASE.md. One refinement: `src_frame` is now emitted whenever the decoder position is known, even when `pts_s` has to be omitted, so consumers always keep the exact index.

**2. Coordination with #118**: this branch now uses #118's exact tuple shape, `(image, tframe, extras)` with an extras dict (`{}` for non-file sources), and its `frame_n` key, so the two branches' deque/read() edits resolve to identical code and extras merge by key union. Remaining naming question (`src_frame` vs `frame_index` at meta level): position recorded in the design note, this PR has no attachment to `src_frame` and will adopt `frame_index` if #118 lands first; either way the second-to-merge PR emits a single key. The post-seek raw `cap.read()` point is answered on the inline thread: `_cap_read()` is documented as the single choke point here, and whoever merges second routes the replay compensation read through it.

**3. VFR**: intended behavior for fps-reporting VFR sources is now stated in the `_cap_read` comment and design note section 4 (deliberate CFR treatment, exact round-trip under frame-index seeking, POS_MSEC unverifiable on such files; `pts_s` omitted rather than emitted wrong when nothing is trustworthy).

**4. Arity**: accepted break with a RELEASE.md "Breaking Changes" callout worded to match #118's, absorbed going forward by the extras dict.

**check-release-log**: RELEASE.md now has the Added entry and the Breaking Changes callout.

**Tests**: added `test_pts_reader_tuple` (pins the reader-level 3-tuple and extras keys), `test_pts_pos_msec_fallback` (POS_MSEC branch), `test_pts_pos_msec_untrusted` (pts-omitted branch), `test_pts_non_file_unchanged` (non-file reads leave extras empty, so stream/webcam meta cannot gain the keys; the fallback branches use a deterministic FakeCap since real backends vary, which is the very reason the fallback exists). 15/15 passing locally.

**DCO**: both new commits are signed off. The original commit 1de6d6a still needs the amend + force-push; I did not rewrite it in this pass to avoid force-pushing over the reviewed state mid-review, will do it as the final step once the re-review is green (or squash-merge makes it moot if that is the repo's merge mode).

shingonoide (commented):
Thanks, this addresses the coordination cleanly. Adopting #118's `(image, tframe, extras)` shape and the `frame_n` key means the two branches union rather than conflict, and the VFR rationale (deterministic frame round-trip, with POS_MSEC confined to no-rate containers) is convincing. No further code concerns from me.

The remaining items are about landing rather than the code: since the single-break story depends on #118, it's worth landing the two in the agreed order so external callers really do see one break, and DCO still needs the `feat` commit `1de6d6a` signed off. Glad to clear our review once those settle.

leandrobmarinho:
Thanks. Both remaining items are landing/process, addressed as follows:

**Landing order**: now recorded in `docs/design-video-in-pts.md` §3 (3d75f88). Agreed order: #118 merges first (it owns the extras-dict shape and design doc; this PR already adopted its shape and `frame_n` key), then this PR rebases on top, adopting the surviving meta key name (`frame_index`) and routing the post-seek compensation read through `_cap_read()` in the same rebase, landing in the same release cycle so the Breaking Changes callout appears once. Fallback if #118 stalls: reverse order, same second-mover obligations.

**DCO on 1de6d6a**: the repo allows squash merge only (merge commits and rebase merges are disabled), so the merge will produce a single fresh commit regardless. If the DCO gate must still be green pre-merge, the plan stands: `git commit --amend -s` on 1de6d6a plus force-push as the very last action once your review is cleared, so the reviewed tree is never rewritten mid-review (the amend changes the message only, no content). I'm deliberately not force-pushing before that point.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -259,6 +261,46 @@ def stop(self):  # idempotent and safe to call whenever
...
+            # as CFR: a true-VFR file that reports an average rate gets a nominal
+            # timeline (frame_n / avg_fps) that can drift from the container pts,
+            # but stays exact under the inverse mapping (offset * avg_fps) used
+            # to seek by frame, so jump-to-frame round-trips to the same frame -
+            # while POS_MSEC on such files hits the bugs above with no way to
+            # tell a true pts from a lie. POS_MSEC is used only for containers
+            # reporting no frame rate at all, and only while it looks sane
+            # (nonzero past frame 0); otherwise pts_s is omitted and frame_n
+            # remains the consumer's only (still exact) position.
+            if self.native_fps:
```
lucasmundim: `if self.native_fps:` trusts any nonzero `CAP_PROP_FPS`, but OpenCV's ffmpeg backend reports a sentinel 1000 fps for some VFR mkv/webm files (the 1 ms MKV timebase, not a real average). Those containers take this primary path and get `pts_s = frame_n / 1000` — ~1 ms per frame — instead of the `POS_MSEC` fallback the design note reserves for containers with no meaningful rate. The round-trip argument in the design note doesn't cover this case, since 1000 isn't an average anything drifts around; the emitted offsets are just wrong, which is the original symptom this PR fixes. Suggest a sanity gate, e.g. treat `native_fps >= 1000` as unreported so these fall through to the guarded `POS_MSEC` branch.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -657,6 +702,18 @@ class VideoIn(Filter):
             together with `maxsize`, it is one or the other. Set here to apply to all sources or can be set individually
             per source. Global env var default FILTER_RESIZE / VIDEO_IN_RESIZE.
 
+    Emitted frame meta:
+        Every frame carries meta['id'] (delivery counter, rate depends on the consuming chain), meta['ts'] (wall-clock
+        seconds at read), meta['src'] and meta['src_fps']. File sources (file:// and s3://) additionally carry the
+        decoder position of the delivered frame - the video offset that cannot be reconstructed downstream:
+
+            meta['src_frame'] - 0-based source frame index (CAP_PROP_POS_FRAMES sampled before the read, exact).
```
lucasmundim, line 722: Worth one extra line here: with `!loop`, the cap is reopened each pass, so `src_frame`/`pts_s` restart at 0 while `meta['id']` keeps counting. That's the right semantics (position within the file), but consumers indexing a looped source should know the timeline is non-monotonic across passes.

lucasmundim (commented):
Re-reviewed at 3d75f88. The rework since the last round holds up: the extras-dict shape matches #118, `_cap_read()` is genuinely the only read path on this branch, and the fallback branches are tested deterministically. Ran the full `test_filter_video_in.py` suite locally: 15 passed. Two things before merge: the DCO check fails on 1de6d6a (needs a signoff rebase), and the PR description still describes the old 4-tuple shape / 11 tests — worth updating so the record matches what lands.

shingonoide (commented):
Re-checked at this head: the stamping (0-based src_frame, pts_s from native_fps so maxfps-skip stays correct, POS_MSEC fallback confined to no-rate containers), the (image, tframe, extras) shape and frame_n key matching #118, and the tests all still look right, and the agreed #118-first landing order is now recorded in the design note. One remaining item: DCO is red because commit 1de6d6a is missing a Signed-off-by. A git rebase --signoff clears it. Once that is green and #118 lands first per the agreed order, this is good to go.

openfilter/filter_runtime/filters/video_in.py
leandrobmarinho: Fixed in 1728ceb. Added a `FPS_SANE_CEILING = 1000.0` gate: the CFR path (`pts_s = frame_n / native_fps`) is now taken only for a *plausible* reported rate, so `if self.native_fps and self.native_fps < FPS_SANE_CEILING:`. Any rate at/above the ceiling — or non-finite (`nan`/`inf` both fail the `< ceiling` compare) — is treated as no usable rate and falls through to the guarded `POS_MSEC` branch exactly like a no-rate container, so the ~1000 fps 1 ms-timebase sentinel no longer emits `frame_n / 1000` offsets; `src_frame` stays present and exact. The code comment at the gate now documents this exception to the round-trip argument, and design note s4 + RELEASE.md state the boundary. New tests: `test_pts_fps_sentinel_falls_through_to_pos_msec` (covers 1000, 1500, inf, nan -> POS_MSEC) and `test_pts_fps_plausible_uses_cfr` (rate just under the ceiling still takes CFR).

openfilter/filter_runtime/filters/video_in.py
leandrobmarinho: Documented in ca9a769. Added the note to the Emitted-frame-meta docstring (and design note s4a): with `loop` the cap is reopened each pass, so `src_frame`/`pts_s` restart at 0 every loop (position within the file) while `meta[id]` keeps counting — non-monotonic in `src_frame`/`pts_s`, monotonic in `id`, so consumers indexing a looped source must key off `id`. Also added `test_loop_pts_restarts_per_pass` pinning the `[0,1,2,0,1,2]` src_frame pattern over `loop=2` with strictly-increasing `id`.

leandrobmarinho:
Latest review round resolved at ca9a769 (pushed, no force-push):

- **VFR sentinel-fps** (@lucasmundim, line 297) -> `1728ceb`: `FPS_SANE_CEILING = 1000.0` gate so the CFR path is taken only for a plausible reported rate; the ~1000 fps 1 ms-timebase sentinel (and non-finite rates) now fall through to the guarded `POS_MSEC` branch instead of emitting `frame_n / 1000` offsets. `src_frame` stays exact. Tests + design-note s4 + RELEASE.md updated.
- **`!loop` restart** (@lucasmundim, line 710) -> `ca9a769`: documented in the frame-meta docstring and design note s4a (src_frame/pts_s restart per pass, id keeps counting -> key off id for looped sources), plus `test_loop_pts_restarts_per_pass`.
- `./.venv/bin/python -m pytest tests/test_filter_video_in.py -q` -> **18 passed** (was 15; +3 for the sentinel fall-through, plausible-rate CFR, and loop behavior).

**DCO on `1de6d6a`:** both new commits are signed off, and only `1de6d6a` is missing `Signed-off-by`. Clearing that on the branch means rewriting history and force-pushing, which I am intentionally not doing. This repo is squash-merge-only, and the squash produces a single fresh commit authored/signed at merge time, so the merged history carries a valid `Signed-off-by` regardless of the intermediate commit — the red check is on a commit that will not exist post-merge. If the gate must be green *before* merge as a hard precondition, the author can `git commit --amend -s` on `1de6d6a` as the final step (a deliberate force-push), but I have left the branch un-rewritten per the no-force-push constraint.

PR description still needs updating to the 3-tuple/18-tests shape (flagged by @lucasmundim) — leaving that to the author since it is the PR body, not code.

lucasmundim (commented):
Re-reviewed at ca9a769. The two commits since my last pass both check out:

- **1728ceb** — the `FPS_SANE_CEILING` gate is correct: `nan`/`inf` both fail the `< ceiling` compare and fall through to the guarded `POS_MSEC` branch, and the new sentinel/plausible tests pin the boundary from both sides.
- **ca9a769** — loop semantics are now documented in the docstring and design note §4a, and `test_loop_pts_restarts_per_pass` asserts exactly the contract (src_frame restarts per pass, id strictly increasing). The reopen path refreshes extras on the first post-reopen `_cap_read()`, so no stale position can cross a loop boundary.

I'm resolving my two threads. Still outstanding before merge, both carried over from my last review:

1. **DCO is still red** — 1de6d6a is missing a `Signed-off-by`; a `git rebase --signoff` + force-push clears it.
2. **The PR description is stale** — it still describes the old 4-tuple `(image, tframe, pts_ms, frame_no)` shape and "11 passed", but the branch ships the 3-tuple `(image, tframe, extras)` shape and a 16-test suite. Please update it so the record matches what lands.

Once DCO is green and #118 lands first per the agreed order, this is good to go from my side.

shingonoide:
Thanks for the thorough write-up, Leandro. The code side is resolved on our end at ca9a769 (the sentinel-fps gate and the loop-restart doc/test both check out). The one thing still holding this back is the DCO, and I want to be direct about it: signing off 1de6d6a is a merge precondition for us, not a cosmetic red.

On the squash point: a squash merge does produce a single signed commit on main at merge time, but that does not turn the pre-merge DCO check green. The check evaluates the commits on this branch as they stand, and it stays red until 1de6d6a carries a Signed-off-by. So from the gate's point of view the red is not "on a commit that will not exist", it is blocking the merge right now, and we are not going to merge with a red DCO.

The fix is the one the DCO check itself prescribes, and it is safe in exactly your situation (sole author, nobody building on this branch):

```
git rebase HEAD~6 --signoff
git push --force-with-lease origin feat/video-in-pts
```

A sign-off rebase only adds the Signed-off-by trailer to the commit messages. The tree is byte-for-byte identical, so it does not rewrite the reviewed content, it only makes the messages compliant. Adding the trailer to an older commit after the fact is normal and exactly what DCO expects, there is no problem signing off and force-pushing here. The no-force-push-mid-review concern no longer applies: our code review is cleared, so this is the final step you already described, and force-with-lease keeps it safe.

References:

- DCO check and how to fix it: https://probot.github.io/apps/dco/
- What the sign-off attests: https://developercertificate.org/

Once the DCO check is green, and with #118 landing first per the agreed order, this is good to go from our side. The PR description still needs the 3-tuple/18-test update as well.

leandrobmarinho:
All feedback is addressed and CI is green.

- **Review threads:** resolved across the earlier rounds (design note, landing order with #118, VFR sentinel + native_fps gate, looped-restart doc). Thanks @shingonoide / @lucasmundim, and confirming your "code side is resolved on our end".
- **DCO (the last red check):** the base `feat` commit was missing its `Signed-off-by`; fixed with `git rebase --signoff` (the 5 commits that already had it were not duplicated). This rewrote the commit hashes but the content is byte-identical to what was reviewed. DCO now passes and every other check (unit tests 3.10-3.13, SAST, secret scan, grype, release-log) is green.

Ready for a re-review / approval whenever you have a moment.

lucasmundim (approved):
Re-reviewed at d5098bc. The force-push is the signoff rebase plus a rebase onto current main - the PR's files are byte-identical to ca9a769, which I had already reviewed. DCO is now green on all commits, and I ran `tests/test_filter_video_in.py` locally at this head: 18 passed.

One item carried over: the PR description still describes the old 4-tuple `(image, tframe, pts_ms, frame_no)` shape and "11 passed" - please update it to the shipped `(image, tframe, extras)` contract so the record matches what lands.

Approving. Good to go from my side once #118 lands first per the agreed order and unit-test CI completes.

leandrobmarinho:
@shingonoide since the code side is resolved on your end and everything (including DCO) is green now, could you convert the changes-requested into an approve when you get a chance? Happy to walk through anything still open, but I believe this is ready to land. Thanks!

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -259,6 +268,51 @@ def stop(self):  # idempotent and safe to call whenever
...
+            # tell a true pts from a lie. The CFR path is taken only for a
+            # PLAUSIBLE reported rate: an implausible one (>= FPS_SANE_CEILING or
+            # non-finite - e.g. the ffmpeg-backend ~1000 fps VFR sentinel, the 1 ms
+            # container timebase rather than an average) is not a rate anything
+            # drifts around, so frame_n / native_fps would just be wrong; those
+            # fall through to POS_MSEC like a no-rate container. POS_MSEC is used
+            # only for containers with no usable rate, and only while it looks sane
+            # (nonzero past frame 0); otherwise pts_s is omitted and frame_n
+            # remains the consumer's only (still exact) position.
+            if self.native_fps and self.native_fps < FPS_SANE_CEILING:
```
shingonoide, line 309: Confirmed the CFR path and the sentinel, inf, and nan fall-through are correct. One theoretical edge, not worth changing now: a negative native_fps would slip this guard and yield a negative pts_s, but no real backend reports that. Fine to leave as a documented known limitation.

shingonoide (approved):
Re-reviewed at d5098bcc. The four items from the prior round are all addressed: the design note records the frame-meta contract, the #118 coordination and landing order, the VFR intent, and the back-compat position; RELEASE.md carries the Added and Breaking Changes entries; DCO and check-release-log are green; and the tests now cover the POS_MSEC fallback, the omit-when-untrusted branch, the file-only gating, the VideoReader 3-tuple, and the loop restart. Ran the suite locally: 18 passed on 3.12, matching CI across 3.10 to 3.13.

Derivation looks correct: src_frame is the 0-based index sampled before the read, pts_s uses frame_n divided by native_fps for a plausible reported rate and falls through to POS_MSEC (guarded, sane-only) otherwise, with the ~1000 fps sentinel and non-finite rates correctly excluded from the CFR path. File-only gating holds: non-file sources keep extras empty and gain no keys, and id, ts, src, and src_fps are unchanged.

Two notes, neither blocking:

1. This stamps pts_s and src_frame for every file frame, where #118 gates its replay meta behind opt-in. The per-frame cost is trivial next to the decode, so this is fine as is; just flagging the always-on versus opt-in difference so the two surfaces stay intentional.
2. The design note agrees #118 lands first, but #118 is open and has not moved since 2026-07-20, and its own thread does not yet reference this coordination. Before either merges, please get an explicit acknowledgment on #118 (or cross-link the two PRs) so the second-mover obligations (reconcile the single meta key name, route #118's post-seek read through the same path) have an owner on both sides rather than only in this PR's doc.

One housekeeping item: the PR description still describes a 4-tuple (image, tframe, pts_ms, frame_no) and 11 passed; the code ships the 3-tuple (image, tframe, extras) and 18 tests. Worth updating so the record matches.

lucasmundim (approved):
LGTM

shingonoide:
Heads-up that a v1.2.0 cut is going out shortly, and it reshapes the top of RELEASE.md: the `[Unreleased]` section is promoted to a dated v1.2.0 heading, and the never-released v1.1.3 section is folded into it. You will most likely hit a conflict on this branch as a result.

The first change in that sequence is already up as #130 if you want to see the shape of it.

Nothing needed from you now, just flagging so it is not a surprise. I will let you know once the cut is merged.

shingonoide (approved):
Re-reviewed at 0ce5c11b. The new commit cleanly resolves the looped source behavior: it adds `test_loop_pts_restarts_per_pass` which verifies that `src_frame` and `pts_s` correctly restart at 0 per pass while `id` remains strictly increasing. The design note and `VideoIn` docstring are also updated to document this non-monotonic timeline expectation.

Ran the suite locally: all 19 tests pass. LGTM.

PR #130, opened 2026-07-27
fix(resize): map 'lin' size suffix to cv2.INTER_LINEAR

shingonoide:
The size spec accepted by `VideoIn` and by the `Util` filter documents three interpolation suffixes, `near`, `lin` and `cub`, and the parser has always accepted all three. The ternary mapping the parsed suffix to an OpenCV constant only handled `N` and `C` explicitly, so `L` fell through to nearest-neighbour and `cv2.INTER_LINEAR` appeared nowhere in the shipped package.

The two call sites are fixed in the shape each one needs. In `video_in.py` the final branch now returns `cv2.INTER_LINEAR` outright, which is safe because the ternary normalises the suffix itself and the size regex only admits the six accepted spellings. In `util.py` the chain gains an explicit `cv2.INTER_LINEAR if interp == 'L'` step and keeps `cv2.INTER_NEAREST` as the fallback, because xforms supplied in dict form bypass the parser that upper-cases the suffix, so an unnormalised `n` or a typo would otherwise select bilinear silently.

**This is a behaviour change.** Pipelines already passing `lin`, for example `resize 640x480lin`, are getting nearest-neighbour output today and will get bilinear output after this. That changes resampled pixel values and may shift downstream detection or OCR results. Pipelines relying on the old output should pin `near` explicitly.

A read over the stored pipeline configuration in all three environments found no pipeline, pipeline version or instance override passing any interpolation suffix, so the change affects no deployed workload. The known users are developer-facing scripts in this repo and in a handful of filter repos.

Adds regression tests covering the full `None` / `N` / `L` / `C` mapping in both files, plus the lowercase and unknown-value fallbacks. The existing size tests could not have caught this: at the 160x100 fixture size they all use, nearest and linear produce byte-identical output. The new tests resize to 640x400, where all three interpolations differ.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/130"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787768293&installation_model_id=20112&pr_number=130&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F130&signature=248b6e876af68ff19a33eeea592d2a46c2271f30d3480c3e4c9219a6170aff8b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


leandrobmarinho (approved):
Clean fix, and the two call sites are correctly handled differently. In `util.py` the added `cv2.INTER_LINEAR if interp == 'L'` keeps `cv2.INTER_NEAREST` as the fallback, which is right because a dict-form xform bypasses the parser's `.upper()[:1]` normalization, so a raw `n` or a typo must land on nearest rather than silently selecting bilinear. In `video_in.py` returning `cv2.INTER_LINEAR` in the final branch is safe because that path always goes through the size regex, which normalizes the suffix to N/L/C, so the else can only be `L`. The tests are what make this trustworthy: they resize at 640x400 (where nearest/linear/cubic actually differ, unlike the 160x100 fixture the old tests used) and spy on `cv2.resize` to assert the exact constant, including the None/n/near/nope -> nearest fallbacks. The behavior change (`lin` now bilinear instead of nearest) is documented and verified to affect no stored pipeline. No blocking issues.

lucasmundim (approved):
Verified the fix end to end: the two ternary chains were the only interpolation call sites in the codebase, and each fix has the right shape for its context — the bare else in video_in.py is safe because the size regex plus interp.upper()[:1] normalization mean only 'L' can reach it, while util.py correctly keeps INTER_NEAREST as the fallback since dict-form xforms bypass the string parser. Ran the three new regression tests against the branch and they pass; asserting on the interpolation kwarg passed to cv2.resize (rather than pixels) is what makes them able to catch this, given nearest and linear are byte-identical at the old 160x100 fixture size. Changelog entry follows the [Unreleased] convention, and the behaviour change is clearly documented with the deployed-config check. LGTM.

PR #131, opened 2026-07-27
fix(deps): move opentelemetry-resourcedetector-gcp pin to a stable range

shingonoide:
## Summary

`pyproject.toml` pinned `opentelemetry-resourcedetector-gcp~=1.12.0a0`. This moves it to `~=1.14.0`.

`~=1.12.0a0` expands to `>=1.12.0a0,<1.13.dev0`. The only version PyPI has ever published inside that range is `1.12.0a0`
itself, and it is a pre-release. uv accepts a pre-release for a package it pulls in indirectly only when that package has
never published a stable version at all, so any project that declares `openfilter` as a dependency and resolves it from a
package index with plain uv gets an unsatisfiable resolution:

```text
  × No solution found when resolving dependencies:
  ╰─▶ Because only the following versions of
      opentelemetry-resourcedetector-gcp are available:
          opentelemetry-resourcedetector-gcp<1.12.0a0
          opentelemetry-resourcedetector-gcp>1.13.dev0
      and openfilter==99.0.0 depends on
      opentelemetry-resourcedetector-gcp>=1.12.0a0,<1.13.dev0, we can conclude
      that openfilter==99.0.0 cannot be used.
      And because your project depends on openfilter[all]==99.0.0, we can
      conclude that your project's requirements are unsatisfiable.

      hint: `opentelemetry-resourcedetector-gcp`
      was requested with a pre-release marker (e.g.,
      opentelemetry-resourcedetector-gcp>=1.12.0a0,<1.13.dev0), but
      pre-releases weren't enabled (try: `--prerelease=allow`)
```

pip applies a looser rule and installs the identical wheel without complaint, resolving
`opentelemetry-resourcedetector-gcp-1.12.0a0`. That asymmetry is why the problem has gone unnoticed: consumers on pip see
nothing wrong, consumers on uv cannot install us at all.

## Why the failure is invisible to our own CI

Our CI installs via `make install`, which is `pip install -e .[all,dev]`. Two things independently hide the bug there.
pip does not enforce uv's pre-release rule, and an editable or local-path install makes
`opentelemetry-resourcedetector-gcp` a *direct* dependency of the project being resolved, which is the case where uv
honours the pre-release marker and resolves fine as well. The failure only appears when openfilter is consumed as a
published distribution, which is precisely the path CI never exercises.

## Why 1.14.0 and not 1.13.0

`1.13.0` (2026-07-22) was the first stable release, but upstream yanked it with the reason "breaks imports". `1.14.0`
(2026-07-24) is the first usable stable release, so it is the floor here.

## Why the other two pre-release pins are deliberately untouched

`opentelemetry-exporter-gcp-monitoring~=1.12.0a0` and `opentelemetry-semantic-conventions~=0.65b0` are also pre-release
pins, but they are not bugs. Neither package has ever published a stable release (13 and 61 versions respectively, all
of them `aN` / `bN`), which is exactly the condition under which uv accepts a pre-release. Changing them would be churn
with no effect. Both still resolve cleanly after this change.

## Blast radius

None in the runtime. openfilter never imports this package; its only reference anywhere in the repository was the
`pyproject.toml` line being changed. It arrives transitively through `opentelemetry-exporter-gcp-monitoring`, whose own
requirement is `opentelemetry-resourcedetector-gcp==1.*,>=1.5.0dev0`, so `1.14.0` satisfies the same constraint the
exporter already asks for.

## Verification

`uv lock` and `uv lock --check` both succeed with no `--prerelease` flag. The lock moves
`opentelemetry-resourcedetector-gcp v1.12.0a0 -> v1.14.0` and drops its now-unneeded `typing-extensions` edge.

Test suite is unchanged. `make test` on this branch gives `751 passed, 25 skipped, 5 deselected, 9 subtests passed`,
identical to the same command on unmodified `main`.

The registry path was proven end to end rather than assumed. A wheel was built from this branch and another from
unmodified `main`, each served over `--find-links`, and a separate consumer project declaring `openfilter[all]` resolved
against each with plain uv and no flags. The `main` wheel produces the resolution error quoted above. This branch's wheel
gives `Resolved 69 packages`, and a full `uv sync` then installs and imports openfilter with
`opentelemetry-resourcedetector-gcp 1.14.0`, alongside `opentelemetry-exporter-gcp-monitoring 1.12.0a0` and
`opentelemetry-semantic-conventions 0.65b0`.

`RELEASE.md` gains an entry under the existing `### Changed` heading in `[Unreleased]`.

Refs DT-186

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/131"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787770044&installation_model_id=20112&pr_number=131&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F131&signature=d36d7dcd4418b54bccbbd148029577d011c72350f3a6922365adc0ca489b8135"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


leandrobmarinho (approved):
Correct fix, verified against PyPI directly rather than the writeup. The package's published versions are `1.0.0a0…1.12.0a0`, then `1.13.0` and `1.14.0`, so the `~=1.12.0a0` range (`>=1.12.0a0,<1.13.dev0`) contained only the `1.12.0a0` pre-release — which uv refuses for an indirectly-pulled dependency, so any project consuming published `openfilter` via uv couldn't resolve, while pip and editable installs (what CI runs) hide it. `1.14.0` is a real stable release (`1.13.0` was yanked upstream), it satisfies the transitive exporter's own `==1.*,>=1.5.0dev0` requirement, and openfilter never imports the package, so there's no runtime blast radius. Leaving `opentelemetry-exporter-gcp-monitoring~=1.12.0a0` and `opentelemetry-semantic-conventions~=0.65b0` untouched is the right call — those have never published a stable release, which is exactly the case where uv accepts a pre-release. `uv lock --check` passing without `--prerelease` confirms it resolves. No blocking issues.

lucasmundim (approved):
Verified independently: `1.12.0a0` is the only version in the old range and PyPI shows `1.13.0` yanked, making `1.14.0` the correct stable floor. The exporter's `==1.*,>=1.5.0dev0` constraint is satisfied, `requires-python >=3.10` matches our support matrix, and the package is not imported anywhere in the codebase. CI is green across 3.10–3.13.

PR #132, opened 2026-07-27
fix(cascade): emit a bump bullet the changelog parser accepts

shingonoide:
## What

The cascade's dependency-bump script now writes `- Bump the openfilter dependency to X.Y.Z` into a consumer's `RELEASE.md` instead of `- Bump openfilter to X.Y.Z`.

## Why the old wording was a release blocker

`changelog-parser-action` rejects any bullet matching `/^-\s+Bump\s+\S+\s+to\s+\d+(?:\.\d+)+/` once that bullet sits inside a dated release section. Our `check-release-log` job runs that parser on every pull request to `main`, so the rule is not advisory: it gates merges.

Under `## [Unreleased]` the old bullet was harmless, and the cascade always targets `[Unreleased]`, so nothing failed at bump time. The failure arrived later. As soon as a consumer cut a release and promoted `[Unreleased]` to a version heading, every bump bullet that had accumulated underneath came along with it and started failing the parser. Each repo then had to hand-patch its own changelog to get back to green.

## Evidence

On `PlainsightAI/filter-protege-model`:

- Commit 2e9a14872c60e266fc56aaae1584315383c44451 cut the accumulated `[Unreleased]` content into a dated `v0.1.28` section, carrying the bump bullets with it. `check-release-log` failed.
- Commit 85f8e6cef4218dc3b3c81cf701d365ad80b802f7 moved those same bullets back under `[Unreleased]`, changing nothing else about them. `check-release-log` passed.

Same bullets, same parser, different section. That isolates the wording plus placement as the cause rather than anything else in the release cut.

## The fix

Naming the dependency puts a word between `Bump` and `to`, so the bullet no longer matches the rejection pattern, and it stays readable and consistent with the imperative mood of the surrounding entries.

The idempotency check moves from an exact line comparison to a regex that also recognizes the previous literal and the past-tense phrasing (`Bumped ... to ...`) used when repos were patched by hand. A re-run over a changelog that already carries one of the older shapes will recognize it instead of appending a near-duplicate beside it. The regex is anchored at both ends and matches the exact version, so it keeps the guarantee the line-anchored check gave: `0.1.9` does not match `0.1.99`.

## Making the guard suite actually run

`tests/test_cascade_bump_strategy.py` drives those rewriters end to end, so its assertions move with the new bullet. It also gains three cases for the widened idempotency check: the old literal already present, the past-tense `- Bumped the openfilter dependency to X.Y.Z.` shape left behind by hand patches, and tolerance for a trailing period.

That suite was not running anywhere. It skips itself when `tomlkit` or `mistletoe` is missing, and neither package was declared in the project, so `run-unit-tests` collected the file and skipped every case in it on all four Python versions. `make install && make test-all` did the same locally. Both are now in the `dev` extra, with `uv.lock` regenerated: two packages added, nothing else moved. A clean `pip install -e .[all,dev]` now gives 28 passed, 0 skipped.

`cascade-pr-checks` runs the file directly as well, so a change to the cascade scripts gets that signal in about thirty seconds rather than waiting on the full matrix. Its path filter now covers the test file itself, so a pull request touching only the tests still triggers it.

## Why this should land before the next tag

The cascade fans out to roughly 39 consumer repos. Tagging openfilter with the current template would seed all of them with the old bullet, and each one would hit the same failure at its own next release cut and need its own hand patch. Landing this first means the next round of bump pull requests carries wording that survives the release cut.

Surfaced by the review discussion on PlainsightAI/filter-protege-model#46.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/132"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787783447&installation_model_id=20112&pr_number=132&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F132&signature=3391a3c86c2d28556111f56624d51f3fe70218a677b92d5b68b8aedb4c0903ef"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/scripts/cascade/bump-strategy.sh
+++ b/scripts/cascade/bump-strategy.sh
@@ -234,14 +234,15 @@ if changed:
 PY
 done
 
-# 3. RELEASE.md — append `- Bump openfilter to ${OF_VERSION}` under the
-#    Unreleased section's `### Changed` subsection, creating either as
-#    needed. If no Unreleased section exists, inject a fresh `## [Unreleased]`
-#    block immediately before the first released `## v...` header — never
-#    write into a tagged release block. Idempotency check is line-anchored
-#    so 0.1.9 doesn't match 0.1.99.
+# 3. RELEASE.md — append a dependency-bump bullet under the Unreleased
+#    section's `### Changed` subsection, creating either as needed. If no
+#    Unreleased section exists, inject a fresh `## [Unreleased]` block
+#    immediately before the first released `## v...` header — never write into
+#    a tagged release block. Idempotency check is line-anchored so 0.1.9
+#    doesn't match 0.1.99.
 python3 <<'PY'
 import os
+import re
 import sys
 
 import mistletoe
@@ -249,7 +250,29 @@ from mistletoe.block_token import Heading
 
 of_version = os.environ["OF_VERSION"]
 path = "RELEASE.md"
-bullet = f"- Bump openfilter to {of_version}"
+
+# The wording is load-bearing, not cosmetic. changelog-parser-action rejects any
+# bullet matching /^-\s+Bump\s+\S+\s+to\s+\d+(?:\.\d+)+/ once it sits in a dated
+# release section, and check-release-log runs that parser on every pull request
+# to main. So the literal `- Bump openfilter to X.Y.Z` this script used to emit
+# turned into a release-blocker the moment a consumer promoted `[Unreleased]` to
+# a version heading, which each one then had to hand-patch. Naming the
+# dependency keeps the imperative mood of the surrounding entries while putting
+# a word between `Bump` and `to`, so the pattern no longer matches.
+bullet = f"- Bump the openfilter dependency to {of_version}"
+
+# Recognize every phrasing that has reached a consumer changelog: what this
+# script emits now, the literal it emitted before, and the hand-reworded past
+# tense used when repos were patched by hand. Without this a re-run would not
+# see the older shapes and would append a near-duplicate beside them. Anchored
+# at both ends and on the exact version, preserving the 0.1.9-vs-0.1.99
+# guarantee the line-anchored check gave.
+bullet_seen = re.compile(
+    r"^-\s+Bump(?:ed)?\s+(?:the\s+)?openfilter(?:\s+dependency)?\s+to\s+"
+    + re.escape(of_version)
+    + r"\.?\s*$",
+    re.IGNORECASE,
+)
 
 
 def heading_text(node: Heading) -> str:
@@ -296,7 +319,7 @@ if not os.path.exists(path):
 with open(path, "r", encoding="utf-8") as f:
     text = f.read()
 
-if any(line.rstrip() == bullet for line in text.splitlines()):
+if any(bullet_seen.match(line) for line in text.splitlines()):
     print(f"{path}: bump entry already present")
     sys.exit(0)
 
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -90,6 +90,7 @@ jobs:
             cascade:
               - 'scripts/cascade/**'
               - '.github/workflows/cascade-on-tag.yaml'
+              - 'tests/test_cascade_bump_strategy.py'
 
       - name: Set up Python
         if: steps.filter.outputs.cascade == 'true'
@@ -97,16 +98,15 @@ jobs:
         with:
           python-version: '3.11'
 
-      - name: Install constraint-check deps
+      - name: Install cascade check and test deps
         if: steps.filter.outputs.cascade == 'true'
         run: pip install --quiet packaging tomlkit mistletoe pytest
 
       # bump-strategy.sh's rewriters are Python heredocs, driven end-to-end by
-      # tests/test_cascade_bump_strategy.py. That file skips itself when
-      # tomlkit/mistletoe are absent, which is how the unit-test job stayed
-      # green while every one of its RELEASE.md cases was silently skipped.
-      # Run it here, where the deps are installed, so a cascade script change
-      # cannot land with stale assertions.
+      # tests/test_cascade_bump_strategy.py. tomlkit and mistletoe are declared
+      # in the dev extra, so run-unit-tests exercises that suite too. This job
+      # re-runs it on the cascade path filter, so a cascade script change cannot
+      # land with stale assertions without waiting on the full matrix.
       - name: Run cascade bump-strategy tests
         if: steps.filter.outputs.cascade == 'true'
         run: pytest tests/test_cascade_bump_strategy.py
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -99,7 +99,17 @@ jobs:
 
       - name: Install constraint-check deps
         if: steps.filter.outputs.cascade == 'true'
-        run: pip install --quiet packaging
+        run: pip install --quiet packaging tomlkit mistletoe pytest
+
+      # bump-strategy.sh's rewriters are Python heredocs, driven end-to-end by
+      # tests/test_cascade_bump_strategy.py. That file skips itself when
+      # tomlkit/mistletoe are absent, which is how the unit-test job stayed
+      # green while every one of its RELEASE.md cases was silently skipped.
+      # Run it here, where the deps are installed, so a cascade script change
+      # cannot land with stale assertions.
+      - name: Run cascade bump-strategy tests
+        if: steps.filter.outputs.cascade == 'true'
+        run: pytest tests/test_cascade_bump_strategy.py
 
       - name: shellcheck cascade scripts
         if: steps.filter.outputs.cascade == 'true'
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,6 +68,12 @@ dev = [
   "jq~=1.12.0",
   "pytest-cov~=7.1.0",
   "pytest-benchmark>=5.1.0",
+
+  # Drive the Python rewriters inside scripts/cascade/bump-strategy.sh. Without
+  # them tests/test_cascade_bump_strategy.py skips itself and the unit-test job
+  # goes green over an unexercised suite.
+  "tomlkit~=0.15.1",
+  "mistletoe~=1.6.0",
 ]
 
 mqtt_out = ["paho-mqtt~=2.1.0", "setuptools~=83.0.0"]
--- a/tests/test_cascade_bump_strategy.py
+++ b/tests/test_cascade_bump_strategy.py
@@ -154,9 +154,11 @@ def test_requirements_non_openfilter_lines_untouched(tmp_path: Path) -> None:
 def test_release_md_idempotency_does_not_match_substring_versions(tmp_path: Path) -> None:
     """Regression for the `bullet in text` substring-match bug.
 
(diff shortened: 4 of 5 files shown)


leandrobmarinho (approved):
Verified the regex logic. The new wording `- Bump the openfilter dependency to X.Y.Z` no longer matches the parser's reject pattern `^-\s+Bump\s+\S+\s+to\s+\d` because `\S+` can only consume the single word `the`, after which the pattern needs `to` but finds `openfilter`; the old `- Bump openfilter to X.Y.Z` did match, so this correctly unblocks release PRs. The `bullet_seen` dedup regex recognizes all three shapes that have reached consumer changelogs (current, prior literal, hand-reworded `Bumped`), the emitted bullet matches its own pattern so a re-run won't duplicate, and `re.escape(of_version)` with the `\.?\s*$` anchor preserves the 0.1.9-vs-0.1.99 guarantee. No blocking issues.

scripts/cascade/bump-strategy.sh
```diff
@@ -234,22 +234,45 @@ if changed:
...
+
+# The wording is load-bearing, not cosmetic. changelog-parser-action rejects any
+# bullet matching /^-\s+Bump\s+\S+\s+to\s+\d+(?:\.\d+)+/ once it sits in a dated
+# release section, and check-release-log runs that parser on every pull request
+# to main. So the literal `- Bump openfilter to X.Y.Z` this script used to emit
+# turned into a release-blocker the moment a consumer promoted `[Unreleased]` to
+# a version heading, which each one then had to hand-patch. Naming the
+# dependency keeps the imperative mood of the surrounding entries while putting
+# a word between `Bump` and `to`, so the pattern no longer matches.
+bullet = f"- Bump the openfilter dependency to {of_version}"
```
lucasmundim, line 262: **Blocking**: `tests/test_cascade_bump_strategy.py` still asserts the old literal and was not updated with this change. Running the suite on this branch with `tomlkit`/`mistletoe` installed:

```
FAILED test_release_md_idempotency_does_not_match_substring_versions
FAILED test_release_md_creates_unreleased_block_when_only_released_versions_exist
FAILED test_release_md_uses_existing_bracketed_unreleased_section
FAILED test_release_md_normalizes_legacy_unreleased_header_to_brackets
FAILED test_release_md_preamble_bullets_do_not_anchor_insert
FAILED test_release_md_idempotent_when_creating_unreleased_block
6 failed, 18 passed
```

They all assert the old bullet, e.g. `rewritten.index("- Bump openfilter to 1.1.0")` at lines 207, 238, 278, 322, 349. CI is green only because the unit-test job's environment lacks `tomlkit`/`mistletoe`, so the file's `skipif` markers silently skip all 24 tests.

#104 — the direct precedent in this same script — moved the tests together with the script, so the convention here expects the same. While updating the literals, the new behavior this PR introduces deserves its own coverage: old literal present → no duplicate appended, past-tense `- Bumped the openfilter dependency to X.Y.Z.` present → no duplicate, and the trailing-period tolerance.

scripts/cascade/bump-strategy.sh
```diff
@@ -296,7 +319,7 @@ if not os.path.exists(path):
 with open(path, "r", encoding="utf-8") as f:
     text = f.read()
 
-if any(line.rstrip() == bullet for line in text.splitlines()):
+if any(bullet_seen.match(line) for line in text.splitlines()):
```
lucasmundim, line 322: Non-blocking, but this line is exactly what the un-run tests would exercise: neither CI job executes `tests/test_cascade_bump_strategy.py`. `run-unit-tests` collects it but skips all 24 tests for missing `tomlkit`/`mistletoe`, and `cascade-pr-checks` only runs shellcheck + actionlint + the discover smoke (it installs only `packaging`).

Since `cascade-pr-checks` is already path-filtered to `scripts/cascade/**`, adding `tomlkit mistletoe pytest` to its pip install and a `pytest tests/test_cascade_bump_strategy.py` step would have caught the failures on this PR. Fine as a follow-up if you'd rather keep this PR single-purpose.

lucasmundim (changes requested):
The fix itself is correct, and every claim in the description checks out against the sources: `changelog-parser-action/src/ChangelogParser.ts:30` carries exactly `/^-\s+Bump\s+\S+\s+to\s+\d+(?:\.\d+)+/`, applied only to non-unreleased sections, and the new wording clears it (`\S+` consumes `the`, then `\s+to` fails against `openfilter`, with no backtracking path that matches). The past-tense hand-patched shape `- Bumped the openfilter dependency to X.Y.Z.` used in filter-protege-model#46, filter-global-position#30, and filter-global-association-multicamera-tracking#19 both clears that parser regex and is matched by the new `bullet_seen` regex, trailing period included. The 0.1.9-vs-0.1.99 guarantee survives the rewrite (`re.escape` + the anchored `\.?\s*$` tail).

One blocking issue, detailed inline: `tests/test_cascade_bump_strategy.py` still asserts the old literal and was not updated — 6 of its 24 tests fail on this branch when run with `tomlkit`/`mistletoe` installed.

Related infrastructure gap, also noted inline: those 24 tests never actually run in CI. `run-unit-tests` skips them all because the environment lacks `tomlkit`/`mistletoe` (the file's `skipif` markers fire), and `cascade-pr-checks` only runs shellcheck + actionlint + the discover smoke. That's why this PR shows green despite the failures — worth closing either here or in a follow-up, since it would have caught exactly this.

Minor, no action needed: `re.IGNORECASE` on `bullet_seen` is more lenient than any shape observed in the wild, but leniency is the safe direction for an idempotency check; and the whole-file scan (rather than Unreleased-only) is pre-existing behavior and semantically right — a bump recorded anywhere means the pins are already at that version.

scripts/cascade/bump-strategy.sh
shingonoide: Fixed in 5fea696. You were right on both counts, and the second one is the more useful finding.

The six stale assertions are updated, and I added the three cases you asked for: an existing old literal `- Bump openfilter to X.Y.Z`, an existing hand-patched `- Bumped the openfilter dependency to X.Y.Z.`, and trailing-period tolerance (parametrized both ways). Each asserts the file is byte-identical afterwards rather than checking for a substring, so an appended near-duplicate fails outright rather than sneaking past a presence check.

Suite is now 28 passed, up from 6 failed / 18 passed. Test functions went 24 to 27 with no deletions or renames; the count is 28 because the trailing-period case is parametrized.

Three occurrences of the old literal remain on purpose, at lines 315, 331 and 375. The first two are the FaceGuard preamble test, where the pre-existing orphan bullet is `- Bump openfilter to 1.0.0` at a different version from the bump under test and represents history the script must preserve verbatim. The third is the fixture for the new legacy-wording test.

I also kept the 0.1.9-vs-0.1.99 guarantee intact rather than dropping it: the test still proves the shorter version does not match the longer entry and that both lines survive, just expressed in the new wording.

scripts/cascade/bump-strategy.sh
shingonoide: Done in this PR rather than as a follow-up. `cascade-pr-checks` now installs `tomlkit mistletoe pytest` alongside `packaging` and runs `pytest tests/test_cascade_bump_strategy.py` under the same path filter.

Bundling it was deliberate: this PR needed a re-review from you regardless, so the CI change costs no extra review cycle, whereas a separate PR would need its own approvals and leave the suite invisible in the meantime. Since the job is already path-filtered to `scripts/cascade/**`, this PR triggers it, which makes the change self-validating.

Confirmed it actually executes rather than skipping again: on the run for 5fea696 the `Run cascade bump-strategy tests` step took 9 seconds and passed. A skip-only run of that file completes in about 0.05 seconds, so the deps are present and the cases are really running.

One gap this does not close, and I left it alone rather than guess at your preference. The filter covers `scripts/cascade/**` and `.github/workflows/cascade-on-tag.yaml`, so a PR touching only `tests/test_cascade_bump_strategy.py` still will not run the step. Adding that path to the `cascade:` filter would close it. Happy to include it here if you want.

Also worth noting the step name `Install constraint-check deps` is now a little under-descriptive since it installs the test deps too. Left as-is to keep the diff tight, but easy to rename.

lucasmundim (approved):
Re-verified at 5fea696: ran the suite with tomlkit/mistletoe installed - 28 passed - and confirmed from the cascade-pr-checks log that CI now executes it (collected 28, passed) instead of skipping. The byte-identical assertions on the legacy/hand-patched wordings are stronger than what I asked for. One optional follow-up: add `tests/test_cascade_bump_strategy.py` to the `cascade` path filter so a test-only change also triggers the job; as it stands such a change would skip everywhere.

scripts/cascade/bump-strategy.sh
shingonoide: Both of the loose ends from my last reply are closed here, and your comment turned out to point at something larger than the CI job.

The `cascade:` filter now lists `tests/test_cascade_bump_strategy.py` alongside `scripts/cascade/**`, so a change to the tests alone triggers the step. The step is renamed to `Install cascade check and test deps`, which is what it does now.

The larger thing: wiring the deps into that one job treated the symptom. `tomlkit` and `mistletoe` were not declared anywhere in the project, which is why the file skipped itself under `run-unit-tests` in the first place. Installing them in a single workflow step left that job still skipping the whole suite on all four Python versions, and left `make install && make test-all` skipping it locally too. Both are now in the `dev` extra with `uv.lock` regenerated, which adds those two packages and moves nothing else. A fresh `pip install -e .[all,dev]` gives 28 passed, 0 skipped, so the matrix exercises the suite as well.

One thing I deliberately left alone: `cascade-pr-checks` still installs the packages unpinned rather than through the extra. Pinning it would make that job disagree with `cascade-on-tag.yaml`, which installs them unpinned at cascade time, so the fast job currently tests against what the cascade will really run. Installing the extra there instead would pull in the full package and cost the job its thirty-second turnaround. Happy to change it if you'd rather have the two agree.

lucasmundim (approved):
Re-approved at a68ede2. The only delta since 5fea696 is the follow-up I suggested plus its natural completion: `tests/test_cascade_bump_strategy.py` is now in the cascade path filter, and tomlkit/mistletoe are declared in the dev extra so the `run-unit-tests` matrix exercises the suite instead of skipping it — verified `make install` pulls `.[all,dev]`, and both pins are compatible with the full 3.10–3.13 matrix (tomlkit needs >=3.9, mistletoe ~=3.5). `uv.lock` is consistent. `cascade-pr-checks` on this head ran the suite green. One optional note: the cascade job installs tomlkit/mistletoe unpinned while the dev extra pins `~=`, so the two environments can drift — fine as-is, but pinning the pip line would keep them aligned.

PR #135, opened 2026-07-31
feat(supply-chain): sign SPDX SBOM for built-in images (keyless cosign)

lucasmundim:
Closes the openfilter gap in the released-artifact vulnerability reporting effort: the 8 built-in images (video-in/out, image-in/out, mqtt-out, recorder, rest, webvis) publish via a bespoke `docker/build-push-action` matrix, not the shared publish composites — so the SBOM-attest work added there does not reach them (they would show up as `released_image_no_sbom` in the scanner). Adds `id-token: write` to the job + Syft + keyless `cosign attest` after each push. Open source, so the public Rekor log is fine.

Related: gh-actions#90, gh-actions-public#32, infrastructure-tooling#8.

**Smoke test:** the first openfilter release after merge confirms the attestation attaches (publish job green + `cosign download attestation --predicate-type=https://spdx.dev/Document <image>` returns the SBOM).

### DIFF
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -406,13 +406,20 @@ jobs:
           - plainsightai/openfilter-rest
           - plainsightai/openfilter-webvis
     steps:
+      # cosign-installer and download-syft run in a job holding id-token: write; whoever
+      # controls a mutable tag on either could mint a Fulcio certificate as this repo's
+      # identity, so these two signing actions are pinned by full commit SHA — the
+      # load-bearing exception to floating tags, matching how sast-secrets.yml pins
+      # trufflehog. cosign-release pins the cosign binary (v2 flag spelling); syft-version
+      # pins the SBOM generator so its output can't shift silently.
       - name: Install cosign
-        uses: sigstore/cosign-installer@v3
+        uses: sigstore/cosign-installer@d58896d6a1865668819e1d91763c7751a165e159 # v3.9.2
         with:
-          # Pin the cosign binary (not the action) to v2.5.2 — the attest flags are v2 spelling.
           cosign-release: v2.5.2
       - name: Install syft
-        uses: anchore/sbom-action/download-syft@v0
+        uses: anchore/sbom-action/download-syft@e22c389904149dbc22b58101806040fa8d37a610 # v0.24.0
+        with:
+          syft-version: v1.50.0
       - name: Login to Docker Hub
         uses: docker/login-action@v4
         with:
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -394,9 +394,12 @@ jobs:
     needs: [prepare-pypi-publish, publish-docker-images]
     # !cancelled() so a single failing publish leg (fail-fast: false leaves the whole
     # publish-docker-images job at failure) does not skip attesting the images that did
-    # push. The matrix is fail-fast: false, so legs whose image is missing fail on their
-    # own without blocking the rest. already_exists gate keeps this to real releases.
-    if: ${{ !cancelled() && needs.prepare-pypi-publish.outputs.already_exists == 'false' }}
+    # push. But require it not to be 'skipped' — if publish-docker-images was skipped
+    # outright (e.g. an upstream wait-for-pypi timeout), nothing was pushed and every leg
+    # here would fail on a non-existent tag; that stays a clean skip. The matrix is
+    # fail-fast: false, so legs whose image is missing fail on their own without blocking
+    # the rest. already_exists gate keeps this to real releases.
+    if: ${{ !cancelled() && needs.publish-docker-images.result != 'skipped' && needs.prepare-pypi-publish.outputs.already_exists == 'false' }}
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/create-release.yaml
+++ b/.github/workflows/create-release.yaml
@@ -309,12 +309,12 @@ jobs:
 
   publish-docker-images:
     runs-on: ubuntu-latest
-    # id-token: write for keyless cosign SBOM attestation (below). contents: read
-    # matches the workflow default; the Docker Hub push uses DOCKERHUB_ACCESS_TOKEN,
-    # not GITHUB_TOKEN, so scoping the token here does not affect publishing.
+    # contents: read matches the workflow default; the Docker Hub push uses
+    # DOCKERHUB_ACCESS_TOKEN, not GITHUB_TOKEN. SBOM attestation runs in the separate
+    # attest-docker-images job (which holds id-token: write), so a Sigstore failure
+    # there cannot gate cascade-pipeline-bumps, which only needs the push here.
     permissions:
       contents: read
-      id-token: write
     needs: [prepare-pypi-publish, publish-to-pypi, wait-for-pypi]
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/create-release.yaml
```diff
@@ -371,6 +377,27 @@ jobs:
...
+      # Supply-chain: attach a signed SPDX SBOM to each built-in image — the same
+      # keyless cosign attestation the shared publish composites add for filters, so the
+      # grype-release-metric daily scanner can re-scan these images from their SBOM (no
+      # pull). openfilter is open source, so the public Sigstore Rekor log is fine.
+      # (syft scans the default-platform image of the multi-arch tag — the amd64 build.)
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0
+      - name: SBOM + keyless attest
```
shingonoide, line 435: `set -euo pipefail` with no `continue-on-error` means a failed attest fails the matrix leg, which is the right default for a supply-chain step. The knock-on is worth an explicit decision though. This step runs after the image is already pushed, and `cascade-pipeline-bumps` has a plain `needs: [..., publish-docker-images]` with no status function in its `if:`, so one leg failing here skips the cascade for all 8 images even though every image published fine.

A fresh `workflow_dispatch` will not redo it either: `create-release` finds the tag already exists and sets `is_release_created=false`, and `prepare-pypi-publish`'s `if:` then short-circuits the whole downstream chain. Driving `cascade-pipeline-bumps.yaml` by hand is possible but its `workflow_dispatch` defaults to `dry_run: true`, so it needs an explicit override, once per image.

If the cascade should not ride on attestation success (the pins only depend on the image having been pushed), moving these three steps into their own matrix job that `needs: publish-docker-images` keeps an attest failure loud without gating the bumps. If you would rather the whole release go red so nobody misses it, leaving it as is works.

shingonoide (approved):
Checked the attest call against cosign v2.5.2, which is what `sigstore/cosign-installer@v3` installs by default: `--yes`, `--type` and `--predicate` all exist at that version, and `spdxjson` is in the accepted `--type` list, so the predicate type lands on `https://spdx.dev/Document` as the comment says. `REF` is correct on every matrix leg, and the Docker Hub login earlier in the job covers both the syft pull and the cosign push. Nothing blocking. One note inline on what an attest failure does to the cascade.

.github/workflows/create-release.yaml
lucasmundim: Done in 81c0f33: moved the three attest steps into a separate `attest-docker-images` job (`needs: publish-docker-images`). `cascade-pipeline-bumps` needs only `publish-docker-images`, so a transient Sigstore/Fulcio failure now surfaces as a red attest job without skipping the version-bump cascade for images that published fine. Also made it per-platform for the arm64 build and pinned the signers.

.github/workflows/create-release.yaml
```diff
@@ -371,6 +377,65 @@ jobs:
...
+          - plainsightai/openfilter-rest
+          - plainsightai/openfilter-webvis
+    steps:
+      - name: Install cosign
+        uses: sigstore/cosign-installer@v3
+        with:
+          # Pin the cosign binary (not the action) to v2.5.2 — the attest flags are v2 spelling.
+          cosign-release: v2.5.2
+      - name: Install syft
+        uses: anchore/sbom-action/download-syft@v0
```
shingonoide: `81c0f33` pinned both of these by SHA and `5b6c04a` reverted them to floating tags. In a job that holds `id-token: write`, whoever controls those two tags can mint a Fulcio certificate as this repo's identity, so this is the one spot in the workflow where the pin is load-bearing.

Worth noting `@v0` currently resolves to exactly the SHA that was removed, so the un-pin gains nothing today and only adds future drift. openfilter is not uniformly float-everything either: `sast-secrets.yml` pins `trufflesecurity/trufflehog@6f3c981e7b77f235fd2702dd74af25fc4b72bf11 # v3.96.0`, which is the same category of action.

Separately, `download-syft@v0` with no `syft-version` means the syft build that produces the SBOM is whatever that tag resolves to on the day of the release, so SBOM contents can shift with no change here. The action takes a `syft-version` input. Restoring the two SHA pins and pinning syft the way cosign is pinned would close both.

shingonoide (approved):
Approving at 5b6c04a5. The earlier concern is resolved: `cascade-pipeline-bumps` needs only `publish-docker-images`, so an attest failure no longer skips the bumps. My previous approval predated the last two commits, so treat this one as replacing it.

Two things worth picking up, neither blocking:

1. `RELEASE.md` already claims this, and it is not true yet. The v0.1.30 entry says Docker images "include SLSA provenance and SBOM attestations via `docker/build-push-action` v6 defaults", and the v1.0.0 entry repeats it. On the published images only `https://slsa.dev/provenance/v1` is attached; `sbom` is a separate opt-in input that this workflow does not set. Since this PR is what makes the claim true, it is the natural place to fix those two lines.

2. Nothing documents how to verify the output. There is no expected `--certificate-identity` or `--certificate-oidc-issuer` for a consumer to check against, and the smoke test in the description uses `cosign download attestation`, which fetches the payload without checking the signature, so it would pass even if the signing identity were wrong. `cosign verify-attestation` is the check that catches that.

One note on action pinning is inline.

.github/workflows/create-release.yaml
lucasmundim: Re-pinned both by SHA in 9a23c51 and pinned `syft-version: v1.50.0`. You're right this is the load-bearing spot — matching how `sast-secrets.yml` pins trufflehog. Repo convention is otherwise always-latest, but the signing actions in an id-token job are the deliberate exception, so they stay SHA-pinned.

lucasmundim:
Thanks for the approval. Both non-blocking points from the review body are addressed:

1. **RELEASE.md** — corrected the v0.1.30 and v1.0.0 lines. They claimed the images carried SBOM attestations, but `docker/build-push-action` defaults attach only SLSA provenance; the SBOM is an opt-in input the workflow never set, so none was produced until this PR. Recorded the real SBOM under `[Unreleased]`.

2. **Verification** — a consumer can check the signing identity (not just fetch the payload) with:

```
cosign verify-attestation --type spdxjson \
  --certificate-identity 'https://github.com/PlainsightAI/openfilter/.github/workflows/create-release.yaml@refs/heads/main' \
  --certificate-oidc-issuer https://token.actions.githubusercontent.com \
  plainsightai/openfilter-video-in:<version>
```

`cosign download attestation` (what the smoke test used) fetches the payload without checking the signature, so it would pass even with a wrong signing identity — `verify-attestation` is the check that catches that.


> **Correction** (per shingonoide's review): this workflow is `workflow_dispatch` and creates the tag in-run, so the signing identity is `@refs/heads/main`, not `@refs/tags/`. Updated above.

.github/workflows/create-release.yaml
```diff
@@ -371,6 +377,72 @@ jobs:
...
+          set -euo pipefail
+          REF="${IMAGE}:${VERSION}"
+          # These images build linux/amd64,linux/arm64. syft catalogs one platform of a
+          # manifest-list, so attest one SBOM per platform to the index digest; the scanner
+          # reads every spdxjson attestation on the ref.
+          for plat in linux/amd64 linux/arm64; do
+            sbom="sbom.$(echo "$plat" | tr '/' '-').spdx.json"
+            syft "$REF" --platform "$plat" -o spdx-json="$sbom"
+            # --type spdxjson sets predicateType https://spdx.dev/Document (scanner filter).
+            cosign attest --yes --type spdxjson --predicate "$sbom" "$REF"
```
shingonoide, line 450: Small correction on the verify command you posted. The identity this job mints will not have a `refs/tags/` ref. `create-release.yaml` is `workflow_dispatch` only and the release tag is created by `create-release` inside the same run, so the run is always dispatched from a branch (and a dispatch from an already-tagged ref stops at the tag-exists check, before anything reaches this job). The certificate SAN is the workflow path plus that run's `github.ref`, so in practice it is:

```text
https://github.com/PlainsightAI/openfilter/.github/workflows/create-release.yaml@refs/heads/main
```

An `@refs/tags/`-anchored regexp will not match that, so the verify fails. Anchoring on `@refs/heads/main`, or passing the exact string via `--certificate-identity`, is what matches. Worth putting the working command somewhere consumers actually see it, since nothing in the repo documents the expected identity or issuer today.

.github/workflows/create-release.yaml
```diff
@@ -371,6 +377,72 @@ jobs:
...
+  # cascade-pipeline-bumps needs only publish-docker-images, so a transient Sigstore/Fulcio
+  # failure here surfaces as a red attest job WITHOUT skipping the version-bump cascade for
+  # images that published fine.
+  attest-docker-images:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: read
+      id-token: write
+    needs: [prepare-pypi-publish, publish-docker-images]
+    if: needs.prepare-pypi-publish.outputs.already_exists == 'false'
```
shingonoide: One residual on the split. `publish-docker-images` is `fail-fast: false`, so a single failing leg still leaves that job's overall result at failure, and with a plain `needs:` here the attest job is skipped entirely. The seven images that pushed fine then ship with no SBOM, and a re-dispatch does not pick them up: the tag now exists, `is_release_created` is false, and `prepare-pypi-publish` short-circuits the chain. Recovering means a manual `cosign attest` per image or cutting a new version. A status check on `needs.publish-docker-images.result` in this `if:` would let the legs that did publish still get attested.

RELEASE.md
```diff
@@ -149,7 +161,7 @@ This release rolls up the `v0.2.0` declarative-configuration work and the `v0.2.
 
 - **Shared security scan workflow**: Replaced standalone Grype security scan with the shared `PlainsightAI/gh-actions-public` reusable workflow.
 - **GitHub Actions bumped to latest versions**: `actions/checkout` v4→v6, `actions/setup-python` v5→v6, `actions/upload-artifact` v4→v7, `actions/download-artifact` v4.1.3→v8, `docker/setup-buildx-action` v3→v4, `docker/login-action` v3→v4, `docker/build-push-action` v5→v6, `dorny/paths-filter` v3→v4, `mukunku/tag-exists-action` v1.6.0→v1.7.0. Resolves Node.js <24 deprecation warnings.
-- **Docker images now include SLSA provenance and SBOM attestations** via `docker/build-push-action` v6 defaults.
+- **Docker images now include SLSA provenance** via `docker/build-push-action` v6 defaults (SBOM attestations are added separately by the release workflow's cosign step; see Unreleased).
```
shingonoide: One leftover tense here. This bullet sits under v0.1.30, and "SBOM attestations are added separately by the release workflow's cosign step" reads as though those images have one. They do not, and will not; the cosign step first runs on the next release. Something like "are added separately starting with the next release" keeps the pointer without the implication.

shingonoide (approved):
Approving at 85f556a3. Both signer pins are back and the SHAs are exactly `v3.9.2` and `v0.24.0`, `syft-version: v1.50.0` closes the floating-syft half, and the two RELEASE.md lines no longer claim an SBOM that is not attached (`openfilter-video-in:1.2.0` still carries only `https://slsa.dev/provenance/v1`). The new `[Unreleased]` section parses fine: the release version still resolves to v1.2.0 and matches `VERSION`.

Three non-blocking notes inline. The one worth acting on: the identity in the verify command you posted will not match what this workflow mints.

.github/workflows/create-release.yaml
lucasmundim: You're right — I had the identity wrong. Since `create-release.yaml` is `workflow_dispatch` and creates the tag in the same run, the cert SAN is `.../create-release.yaml@refs/heads/main`, not `@refs/tags/...`, so my earlier `@refs/tags/`-anchored command would fail. Corrected it and documented a working snippet in the `[Unreleased]` changelog entry (using `--certificate-identity` with the exact `@refs/heads/main` string + the issuer), since nothing in the repo documented the expected identity before.

.github/workflows/create-release.yaml
lucasmundim: Good catch. `publish-docker-images` is `fail-fast: false`, so one failing leg left the whole job failed and the plain `needs:` skipped the entire attest job — the images that pushed fine got no SBOM, unrecoverable without a manual attest or a new version. Gated the attest job on `!cancelled()` so it still runs; its own matrix is `fail-fast: false`, so a leg whose image is missing fails on its own without blocking the ones that did publish. The `already_exists` gate keeps it scoped to real releases.

RELEASE.md
lucasmundim: Fixed — reworded to "SBOM attestations are added separately starting with the next release," so the v0.1.30 bullet no longer implies those images already carry one.

shingonoide (approved):
Approving at 2dc32473. All three items from the last pass are addressed: the attest job now survives a partial publish, the RELEASE.md verify snippet pins both the exact identity and the issuer, and the v0.1.30 bullet no longer implies those images already carry an SBOM.

One follow-up, not blocking. `!cancelled()` also fires when `publish-docker-images` is skipped outright rather than partially failed, for example when `wait-for-pypi` times out. `prepare-pypi-publish` has already emitted `false` by then, so the remaining gate does not catch it, and all eight legs fail on a tag that was never pushed. Adding `needs.publish-docker-images.result != 'skipped'` keeps the partial-failure case you wanted while an aborted release stays a clean skip.

lucasmundim:
@shingonoide thanks for the approval — and good catch on the skip case. You're right that `!cancelled()` alone also runs the attest job when `publish-docker-images` is skipped outright (e.g. an upstream `wait-for-pypi` timeout), where nothing was pushed and every leg would fail on a non-existent tag. Added `needs.publish-docker-images.result != 'skipped'` to the `if:`, so the partial-failure case still attests the images that published while an aborted release stays a clean skip.

shingonoide (approved):
Approving at 2dff04ab. `needs.publish-docker-images.result != 'skipped'` is the right gate. A partial publish leaves that job at failure, so the images that did push still get attested, while an aborted release where the publish job never ran is now a clean skip rather than eight legs failing on a tag that was never pushed. Nothing else moved since the last pass and all earlier threads are closed.

lucasmundim:
@shingonoide thanks for the approval — agreed, `needs.publish-docker-images.result != 'skipped'` keeps the partial-publish case attesting what shipped while an aborted release stays a clean skip.

PR #139, opened 2026-08-08
docker: add openfilter-base image (apt-upgraded python-slim) + weekly build

lucasmundim:
## What

Adds a shared **`openfilter-base`** image so downstream images inherit a patched OS from one place.

Today the built-in filter images and the standalone filter repos each pin a `python:X.Y.Z-slim` base. A pinned patch tag silently drifts behind Debian's point releases, so OS packages (`openssl`/`libssl3`, `gnutls`, the `python` apt package, …) accumulate fixable CVEs that no filter individually patches.

### `docker/base.Dockerfile`
Deliberately **minimal**: `python:${PYTHON_VERSION}-slim` with `apt-get upgrade` applied, plus the common non-root `ENV` / `appuser` / `WORKDIR` layout every openfilter image already repeats. **No system libraries are installed here** — each filter keeps installing only what it needs (ffmpeg + libGL for video, `libzbar0` for QR, …), so the base stays small and no filter inherits a library — and its CVEs — it never uses.

### `.github/workflows/build-base.yaml`
- Builds the matrix `py3.10 .. py3.13` (openfilter's `requires-python` is `>=3.10,<3.14`), multi-arch `amd64`/`arm64`.
- Publishes `plainsightai/openfilter-base:py3.X` (rolling, for consumers) plus an immutable `:py3.X-YYYYMMDD` tag.
- Rebuilt on a **weekly schedule** (so the OS stays current), on demand, and when the base definition changes.
- Attaches a signed SPDX SBOM (keyless cosign, public Rekor) via the shared `attest-sbom` action — same attestation the built-in images get.

## Scope

This PR **only adds the base and its build** — nothing downstream changes yet. Consumers migrate in follow-ups:
1. Built-in images (`docker/*.Dockerfile`) → `FROM plainsightai/openfilter-base:py3.13`, with the base built before them (`needs:`) in the release.
2. Standalone filter repos → `FROM plainsightai/openfilter-base:py3.X`.

## Validation

The build itself is the proof: after `apt-get upgrade`, a grype scan of the base's SBOM should be clean of the OS-package CVEs that the pinned `python:X.Y.Z-slim` carries. I'll dispatch the workflow once this is approved and attach the result.

## Relationship to the Grype release scanner

This is complementary to the released-image Grype scan, not a competing CVE mechanism:

- **openfilter-base patches the OS layer proactively** — `apt upgrade` on every rebuild, before CVEs ever surface in a scan.
- **The scanner catches OS/dependency CVEs reactively** on already-released images.

The base shrinks what the scanner finds; the scanner still covers app/pip-level CVEs (e.g. the ffmpeg bundled inside the PyAV wheel) and anything the base doesn't touch. As filters migrate onto the base, the released-image OS-CVE counts should fall to near zero.

### DIFF
--- a/.github/workflows/build-base.yaml
+++ b/.github/workflows/build-base.yaml
@@ -1,0 +1,86 @@
+name: Build base images
+
+# openfilter-base: a minimal fresh-OS python-slim (apt upgrade only) published per
+# supported Python version to plainsightai/openfilter-base:py3.X. Rebuilt weekly so the
+# OS stays patched — downstream images (the built-in filters here and standalone filter
+# repos) inherit the patched base on their next build instead of each pinning a stale
+# python:X.Y.Z-slim. Also runnable on demand.
+on:
+  schedule:
+    # Mondays 06:00 UTC — ahead of anything that consumes the base that day.
+    - cron: "0 6 * * 1"
+  workflow_dispatch:
+  # Rebuild when the base definition itself changes.
+  push:
+    branches: [main]
+    paths:
+      - docker/base.Dockerfile
+      - .github/workflows/build-base.yaml
+
+permissions:
+  contents: read
+
+jobs:
+  build-base:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ["3.10", "3.11", "3.12", "3.13"]
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v7
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v4
+
+      - name: Compute date tag
+        id: date
+        run: echo "date=$(date -u +%Y%m%d)" >> "$GITHUB_OUTPUT"
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v4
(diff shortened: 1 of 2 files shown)


shingonoide (approved):
Clean, minimal base image and the workflow matches the existing build/sign pattern.

One thing worth a mention, not a blocker: this pairs nicely with the Grype scan workflow you set up recently, that catches OS/dependency CVEs reactively, this patches the OS layer proactively before they show up in a scan at all. Might be worth a line in the PR description once filters start migrating, just so it's clear the two are complementary rather than someone wondering later why there are two CVE mechanisms.

Approving, nothing here blocks the merge.

lucasmundim:
@shingonoide — good call, they're complementary: the base patches the OS layer **proactively** (apt upgrade on each rebuild, before CVEs surface), while grype-release-metric catches OS/dep CVEs **reactively** on released images. The base shrinks what the scanner finds; the scanner still covers app/pip-level CVEs (e.g. the PyAV-bundled ffmpeg) and anything the base doesn't touch. Added this to the description and I'll carry it into the migration PRs. Thanks for approving.

PR #142, opened 2026-08-11
feat: add Python 3.14 support

lucasmundim:
## What

Adds **Python 3.14** to OpenFilter's supported interpreters, keeping 3.10–3.13 fully intact.

## Why it's a small change

I resolved the full dependency set against a real 3.14 target (wheels only, no source builds). Only **one** dependency actually blocks 3.14: **numpy**.

- `numpy~=2.2.6` ships **no cp314 wheels**; cp314 wheels first appear in **numpy 2.3.2**, but numpy ≥2.3 requires **Python ≥3.11** — so a single pin can't span 3.10 and 3.14.
- Fix: split the pin by interpreter with an environment marker. 3.10–3.13 stay on `numpy~=2.2.6` (**unchanged**); only 3.14 pulls `numpy>=2.3.2,<3` (resolves to 2.5.2).

Every other compiled dependency already ships 3.14 wheels:

| Dependency | 3.14 wheel |
|---|---|
| `opencv-python-headless~=5.0.0.93` | `cp37-abi3` stable-ABI wheel (works 3.7→3.14) |
| `av~=17.1.0` | `cp311-abi3` wheel + `cp314t` build (current pin already covers 3.14) |
| `pyzmq~=27.1.0`, `psutil~=7.2.1`, `pydantic-core` | cp314 wheels present |
| `grpcio`, `protobuf`, `cffi`, `charset-normalizer` (transitive) | cp314 wheels present |

## Changes

- `pyproject.toml`: `requires-python` → `>=3.10,<3.15`; add 3.14 classifier; split numpy pin by marker.
- `uv.lock`: regenerated — carries both numpy 2.2.6 (`< 3.14`) and 2.5.2 (`>= 3.14`); no other package changed.
- `.github/workflows/ci.yaml`: add `3.14` to the unit-test matrix.
- `.github/workflows/build-base.yaml`: add `3.14` to build + attest matrices → publishes `openfilter-base:py3.14`.
- `docker/base.Dockerfile`, `README.md`, `README.pypi.md` (regenerated), `CONTRIBUTING.md`, `RELEASE.md`: doc/version-range updates.

## Validation

`uv pip compile --python-version 3.14 --no-build` resolves 28 packages with wheels only; the same spec still resolves on 3.10 (numpy stays 2.2.6). The new 3.14 CI leg exercises the import + full suite against numpy 2.5 for real confirmation.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/142"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1789052781&installation_model_id=20112&pr_number=142&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F142&signature=ea98dbfe462b700b2667e6d1116747486753c3a8a682f8a608600a7a7a90e647"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


jmiller-plainsight (approved):
This looks great to me!!!

shingonoide (commented):
Nice work getting 3.14 support in, the numpy split is exactly the right fix. Quick check confirms `numpy~=2.2.6; python_version < '3.14'` and `numpy>=2.3.2,<3; python_version >= '3.14'` resolve correctly, and the 3.14 unit-test leg genuinely ran and passed (not skipped), same for the base-image build and docs.

One gap: the security scan matrix in `.github/workflows/security-scan.yaml` still defaults to `["3.10", "3.11", "3.12", "3.13"]` (it doesn't pass `python_versions`), so grype never runs against the 3.14 dependency tree. CI checks on this PR confirm it, grype only shows runs for 3.10-3.13, nothing for 3.14. Can you add `3.14` to that override (or bump the shared workflow's default) so the new supported version gets vulnerability coverage too?

lucasmundim:
Good catch, thanks — you're right that the security scan wasn't covering 3.14. `security-scan.yaml` only passed `filter_name`, so it inherited the shared workflow's default grype matrix (`["3.10", "3.11", "3.12", "3.13"]`) and never scanned the 3.14 dependency tree.

Fixed in 2998de8 by overriding `python_versions` locally to include 3.14, rather than bumping the shared workflow's default (that would change coverage for every other consumer repo, which is out of scope here). Grype now runs against the 3.14 tree — which is the branch that resolves numpy 2.5.x, so it's exactly the one worth scanning.

shingonoide (approved):
Confirmed the security scan now covers 3.14 — python_versions is overridden in security-scan.yaml to ["3.10", "3.11", "3.12", "3.13", "3.14"], and the security-scan / grype (3.14) CI job is running and passing. That closes the coverage gap from the earlier pass.

PR #143, opened 2026-08-11
chore: finish Python 3.14 rollout (release gate, images, examples)

lucasmundim:
Follow-ups after #142 (which added Python 3.14 to the `openfilter` package). These close the remaining places that still assumed a 3.13 ceiling.

## 1. Release gate now runs on 3.14
`create-release.yaml`'s `run-unit-tests` matrix gated releases on 3.10–3.13 only, and `create-release` depends on it — so a release could be cut even if 3.14 broke. Added `3.14`.

## 2. Branch protection requires the 3.14 checks
`.github/rulesets/main.json` listed required status checks explicitly per version, only up to 3.13 — so `run-unit-tests (3.14)` and `grype (3.14)` were advisory and a PR could merge with 3.14 red. Added both. Both contexts exist on `main` once this PR lands (unit-tests via #142; `grype (3.14)` via item 4 here), so no deadlock — the old ruleset still gates this PR.

## 3. Built-in filter images build on 3.14
The 8 built-in images (video-in/out, image-in/out, mqtt-out, recorder, rest, webvis) now build on `python:3.14-slim` + `openfilter-base:py3.14` (base tag already published by `build-base` after #142 merged).

Safe by construction: `publish-docker-images` `needs: [prepare-pypi-publish, publish-to-pypi, wait-for-pypi]` and installs `openfilter[...]==<version>` **after** the release is on PyPI — so the builder always resolves the cp314 wheels that the same 3.14-supporting release ships. The Dockerfile bump and the numpy cp314 fix travel together in `main`, so they can never release out of step.

## 4. Security scan + PyPI poll on 3.14
- `security-scan.yaml`: `python_versions` override adds 3.14 so Grype scans the 3.14 dependency tree.
- `wait-for-pypi` (`create-release.yaml`): polls on 3.14 instead of 3.13 — the newest supported interpreter, so its `pip install --dry-run` doubles as an install check on 3.14.

## 5. Example projects widen to `<3.15`
The four `examples/*/pyproject.toml` were capped at `>=3.10,<3.13` (excluded even 3.13) → `>=3.10,<3.15`. `monitoring-demo/Dockerfile` and `image_in/Dockerfile.dev` bumped to `python:3.14-slim`. Standalone demo packages, independent of the shipped wheel.

## 6. Tooling steps standardized on 3.14
The build/publish, readme-sync check, and cascade steps were pinned to 3.11/3.12. They only install pure-Python tools with no `requires_python` upper cap (`build`, `twine`, `packaging`, `tomlkit`, `mistletoe`, `pytest`; `make_pypi_readme.py` is stdlib-only), so they run on 3.14. Standardized on 3.14 so the toolchain stays current the longest (3.11 ages out first) and every job uses one interpreter. The wheel stays a `py3-none-any` universal artifact regardless of build interpreter — this is consistency/future-proofing, not a build change.

## Scope / gate
Touches only `.github/**`, `docker/**`, `examples/**`, `RELEASE.md` — no `openfilter/**` source, so the release-log gate is a no-op; the RELEASE.md `[Unreleased]` entry from #142 is extended for completeness. `VERSION` bump stays for the separate release PR per convention.

### DIFF
--- a/.github/rulesets/main.json
+++ b/.github/rulesets/main.json
@@ -33,10 +33,12 @@
           { "context": "run-unit-tests (3.11)" },
           { "context": "run-unit-tests (3.12)" },
           { "context": "run-unit-tests (3.13)" },
+          { "context": "run-unit-tests (3.14)" },
           { "context": "security-scan / grype (3.10)" },
           { "context": "security-scan / grype (3.11)" },
           { "context": "security-scan / grype (3.12)" },
           { "context": "security-scan / grype (3.13)" },
+          { "context": "security-scan / grype (3.14)" },
           { "context": "cascade-pr-checks" },
           { "context": "check-release-log" }
         ]
--- a/.github/workflows/cascade-on-tag.yaml
+++ b/.github/workflows/cascade-on-tag.yaml
@@ -64,7 +64,7 @@ jobs:
       - name: Set up Python (for check_constraint.py)
         uses: actions/setup-python@v7
         with:
-          python-version: '3.11'
+          python-version: '3.14'
 
       - name: Install constraint-check deps
         run: pip install --quiet packaging
@@ -150,7 +150,7 @@ jobs:
       - name: Set up Python (for bump-strategy.sh)
         uses: actions/setup-python@v7
         with:
-          python-version: '3.11'
+          python-version: '3.14'
 
       - name: Install bump-strategy deps
         run: pip install --quiet packaging tomlkit mistletoe
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -62,7 +62,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v7
         with:
-          python-version: "3.12"
+          python-version: "3.14"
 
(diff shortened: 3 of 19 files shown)


shingonoide (approved):
Verified the Python 3.14 rollout is complete and consistent — release-gate matrix, branch-protection required checks, all built-in filter and dev/demo Dockerfiles, the security-scan matrix, and all four example projects' requires-python caps are all updated to 3.14. CI is green across 3.10–3.14.

One small non-blocking note: the example packages' classifiers lists still cap at 3.12 (missing 3.13/3.14 trove classifiers), but that predates this PR and has no functional effect since these aren't published to PyPI.

jmiller-plainsight (approved):
LGTM

lucasmundim:
Thanks for the thorough verification and the catch. You're right — the four example projects' `classifiers` capped at 3.12 while their `requires-python` already allowed through 3.14. Since this PR already touches those files, I went ahead and added the `3.13` and `3.14` trove classifiers in 854e465 so the metadata matches the declared range. As you noted, it has no functional effect (these examples aren't published to PyPI) — just keeping them internally consistent.

PR #159, opened 2026-09-01
fix: repair hello-world model download

shingonoide:
Fixes [FILTER-648](https://plainsight-ai.atlassian.net/browse/FILTER-648)

`examples/hello-world` downloads the license plate detection model on its own, independent of the `filter-license-plate-detection` package install. Fixing that repo alone doesn't fix this example — it was still pointing at the same dead `models.openfilter.io` URL.

Switched it to the same canonical GCS URL used in the filter repo fix, and hardened the download so a failed download or corrupt archive fails the build instead of leaving a missing or stale `model.pth`.

Artifact monitoring (the weekly check) lives in `filter-license-plate-detection`, not duplicated here.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/159"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790840859&installation_model_id=20112&pr_number=159&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F159&signature=03a32934c67d54e5008707dc921f7ab7eb84b89c4419962fae76a6f2fa4901af"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

[FILTER-648]: https://plainsight-ai.atlassian.net/browse/FILTER-648?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/examples/hello-world/Makefile
+++ b/examples/hello-world/Makefile
@@ -42,7 +42,7 @@ install:  ## Install package with dev dependencies
 
 	@if [ ! -f model.pth ]; then \
 		echo "model not found, downloading model archive..." && \
-		curl -fL -o model.zip https://storage.googleapis.com/plainsight-ml-assets-production/models/license_plate_detection_model_v0.1.0.zip && \
+		curl -fL --retry 2 --retry-delay 5 -o model.zip https://storage.googleapis.com/plainsight-ml-assets-production/models/license_plate_detection_model_v0.1.0.zip && \
 		echo "Unzipping model archive to current directory..." && \
 		unzip -o model.zip && \
 		echo "Removing model.zip..." && \
--- a/examples/hello-world/Makefile
+++ b/examples/hello-world/Makefile
@@ -41,11 +41,11 @@ install:  ## Install package with dev dependencies
 		--extra-index-url https://python.openfilter.io/simple
 
 	@if [ ! -f model.pth ]; then \
-		echo "model not found, downloading model archive..."; \
-		curl -L -o model.zip https://models.openfilter.io/license_plate_detection_model/v0.1.0.zip; \
-		echo "Unzipping model archive to current directory..."; \
-		unzip -o model.zip; \
-		echo "Removing model.zip..."; \
+		echo "model not found, downloading model archive..." && \
+		curl -fL -o model.zip https://storage.googleapis.com/plainsight-ml-assets-production/models/license_plate_detection_model_v0.1.0.zip && \
+		echo "Unzipping model archive to current directory..." && \
+		unzip -o model.zip && \
+		echo "Removing model.zip..." && \
 		rm model.zip; \
 	else \
 		echo "model already exists, skipping download."; \


examples/hello-world/Makefile
```diff
@@ -41,11 +41,11 @@ install:  ## Install package with dev dependencies
 		--extra-index-url https://python.openfilter.io/simple
 
 	@if [ ! -f model.pth ]; then \
-		echo "model not found, downloading model archive..."; \
-		curl -L -o model.zip https://models.openfilter.io/license_plate_detection_model/v0.1.0.zip; \
-		echo "Unzipping model archive to current directory..."; \
-		unzip -o model.zip; \
-		echo "Removing model.zip..."; \
+		echo "model not found, downloading model archive..." && \
+		curl -fL -o model.zip https://storage.googleapis.com/plainsight-ml-assets-production/models/license_plate_detection_model_v0.1.0.zip && \
```
lucasmundim: Nit (non-blocking): this is a ~147 MB download with no retry, so a transient blip now fails `make install` outright — where before it would (wrongly) carry on. The artifact check in the companion PR uses `--retry 2 --retry-delay 5`; mirroring that here would make the local path as resilient as the CI one.

Also, with `-f` curl still creates/truncates `model.zip` before it sees the error status, leaving a 0-byte file behind on failure. `--remove-on-error` (curl >= 7.83) cleans that up:

```make
curl -fL --retry 2 --retry-delay 5 --remove-on-error -o model.zip https://storage.googleapis.com/plainsight-ml-assets-production/models/license_plate_detection_model_v0.1.0.zip && \
```

Harmless as-is since the next run re-downloads and `.gitignore` covers `model.zip` — purely polish.

examples/hello-world/Makefile
```diff
@@ -41,11 +41,11 @@ install:  ## Install package with dev dependencies
 		--extra-index-url https://python.openfilter.io/simple
 
 	@if [ ! -f model.pth ]; then \
```
lucasmundim, line 43: Nit (non-blocking): this URL is now duplicated in three places — here, the filter repo's `download-model` target, and the weekly artifact-check workflow the companion PR adds. That workflow carries a comment pointing back at this file, but this file has no pointer the other way, so whoever bumps the model version here won't know the weekly check silently keeps testing the old object. A one-liner would close the loop:

```make
# Model URL is duplicated in filter-license-plate-detection (Makefile +
# .github/workflows/model-artifact-check.yaml) — update both together.
```

lucasmundim (approved):
Verified the fix against the live artifacts:

- The old URL is confirmed dead — `https://models.openfilter.io/license_plate_detection_model/v0.1.0.zip` returns `307` to `model-registry.plainsight.ai/license_plate_detection_model_v0.1.0.zip`, which returns `404`.
- The new GCS URL returns `200 application/zip`, 153,898,697 bytes, publicly readable with no credentials.
- I parsed the archive's central directory (range request on the tail): it contains exactly one entry, `model.pth`, at the top level. So `unzip -o model.zip` lands `./model.pth` in the example directory, which is what the `[ ! -f model.pth ]` guard and the pipeline expect.
- The `;` -> `&&` change is the right fix for the silent-failure half of this bug: the `then` list now exits with the failing command's status, so a 404 or a corrupt archive fails `make install` instead of leaving the guard file missing. Previously only `rm`'s status was visible to make.
- Repo-wide grep confirms no other `models.openfilter.io` reference remains — the `docs/` pipeline guides point users at the filter repo for the weights, so nothing else here needs updating.
- URL and hardening match the companion `filter-license-plate-detection#31` change exactly.

Two optional non-blocking nits inline. LGTM.

examples/hello-world/Makefile
shingonoide: Retry added. Skipping --remove-on-error — it needs a newer curl than some plausible contributor baselines, and testing didn't show a leftover file for the HTTP-failure case this guards against.

examples/hello-world/Makefile
shingonoide: The third copy is gone — the companion PR's workflow now runs the filter repo's Makefile directly instead of hardcoding its own URL. Two references remain because these are two independent consumers (this repo and filter-license-plate-detection), and the scheduled check now follows the filter Makefile, so it can't silently stay green against a stale URL here.

examples/hello-world/Makefile
lucasmundim: Retry confirmed on `5af8cf0`, and skipping `--remove-on-error` is fine — under `&&` a leftover stub is harmless anyway, since the next run re-downloads and `-o` truncates it.

One caveat for the record, not a change request: curl's `--retry` only covers timeouts and HTTP 408/429/5xx, so a mid-transfer connection reset on a 147 MB body (`CURLE_PARTIAL_FILE` / recv error) still isn't retried without `--retry-all-errors`. This covers the DNS/connect/5xx half, which is the common case. Good as-is.

examples/hello-world/Makefile
lucasmundim: Verified against `filter-license-plate-detection#31` at `f611cca`: `model-artifact-check.yaml` runs `make download-model` rather than its own copy of the URL, and that Makefile's curl line is byte-identical to this one. That closes the loop I was worried about — the scheduled check can't stay green against a stale URL. Dropping the nit.

lucasmundim (approved):
Re-reviewed at `5af8cf0b`. Re-affirming approval on the current HEAD.

- Makefile at HEAD verified directly: `curl -fL --retry 2 --retry-delay 5`, every step chained with `&&`, `rm model.zip;` closing the `then` list. The silent-failure half of the bug stays fixed - a 404 or corrupt archive now fails `make install` instead of leaving a missing `model.pth`.
- Artifact re-checked live: the GCS URL still returns `200 application/zip`, 153,898,697 bytes, publicly readable with no credentials - same object and size as my first pass.
- Both nits are closed. Retry landed in `5af8cf0b`; the URL-duplication concern went away once the companion PR's scheduled check started invoking the filter repo's `make download-model` instead of carrying its own copy of the URL.
- CI is fully green (unit tests 3.10-3.14, grype 3.10-3.14, DCO, Semgrep, TruffleHog, release-log), and the branch is up to date with `main` (`behind_by: 0`).

One sequencing note, not a blocker: `filter-license-plate-detection#31` is still open at `f611cca`. This Makefile stands on its own and the URL is live, so merging here isn't gated on it - but the duplication loop only truly closes once that one lands, so I'd merge them in the same window.

PR #161, opened 2026-09-02
feat: support directory-based video sources in VideoIn

ahonjaya-del:
## Problem / Feature Goal

Currently, the `VideoIn` filter only supports playing a single `.mp4` file or remote streams. There was no way to supply a folder of multiple video clips directly and have them play sequentially as a single continuous stream.

## Solution

This PR adds native support for local directory-based video sources in `VideoIn`. 

### Key Accomplishments:
1. **Directory Scanning & Verification**:
   - Checks if a local file source points to a directory.
   - Alphabetically sorts and scans immediate files (depth 1).
   - Verifies validity with `cv2.VideoCapture.isOpened()`, gracefully ignoring non-video files.
2. **Seamless Playback Transitions**:
   - Once a video clip finishes, `VideoReader` seamlessly opens the next sorted file.
   - Pacing timers (`ns_per_fps`) are dynamically adjusted to fit the new segment's native or capped FPS.
3. **Dynamic Metadata & Offsets**:
   - `meta['src']` is dynamically updated on the fly to reflect the exact file being read.
   - `meta['src_frame']` and `meta['src_seconds']` are cleanly reset to `0` upon opening each new video file.
4. **End-to-End runnable demo**:
   - Created a complete runnable demo in `examples/directory-pipeline-demo/main.py`.

### 1. Automated Tests
Added three comprehensive unit test cases in `tests/test_filter_video_in.py`:
- `test_directory_source`: Verifies alphabetical sorting, dynamic metadata update, and frame counter resets.
- `test_directory_loop`: Verifies looping of the directory source works seamlessly.
- `test_directory_skips_non_videos`: Verifies non-video formats (e.g. text files) are skipped.

All 25 test cases inside `tests/test_filter_video_in.py` pass successfully:
```bash
pytest tests/test_filter_video_in.py
```

### 2. Manual Demo Verification
Successfully executed the custom demo pipeline (`VideoIn` -> `InfoPrinter` -> `ImageOut`):
```bash
python examples/directory-pipeline-demo/main.py
```
Logs perfectly demonstrate alphabetical directory traversing, frame-index resets (`0` to `2` for each segment), and active source path updating.

---
<a href="https://app.blacksmith.sh/PlainsightAI/codesmith/openfilter/pr/161"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790969655&installation_model_id=20112&pr_number=161&repository=PlainsightAI%2Fopenfilter&return_to=https%3A%2F%2Fgithub.com%2FPlainsightAI%2Fopenfilter%2Fpull%2F161&signature=20c6a581550af463187ea6f5744a43d28ec6338f92a9cc1f887c6e1a86bbfad6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -16,7 +16,6 @@ OpenFilter Library release notes
 
 ### Added
 
-
 - **`VideoIn` / `ImageIn`: `FILTER_OVERRIDE_SOURCE_URI` to preserve source-file identity.**
   A new `override_source_uri` config (env `FILTER_OVERRIDE_SOURCE_URI`, or
   `FILTER_OVERRIDE_SOURCE_URI_FILE` to read it from a file written at runtime) makes both
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,9 +4,18 @@ OpenFilter Library release notes
 
 ## [Unreleased]
 
+### Added
+
+- **`VideoIn` support for directory-based sequential video playback.**
+  Users can now pass a directory of video files as a source (e.g. `file:///path/to/folder`).
+  The video files inside are sorted alphabetically and played sequentially. Includes full
+  support for loop and sync, recursive subfolder scanning via `recursive`, and glob-based file
+  pattern filtering via `pattern`.
+
 ## v1.3.0 - 2026-08-17
 
 ### Added
+
 
 - **`VideoIn` / `ImageIn`: `FILTER_OVERRIDE_SOURCE_URI` to preserve source-file identity.**
   A new `override_source_uri` config (env `FILTER_OVERRIDE_SOURCE_URI`, or
--- a/docs/video-in-filter.md
+++ b/docs/video-in-filter.md
@@ -228,6 +228,34 @@ Filter.run_multi([
 ])
 ```
 
+### 6. Local Directory of Video Files
+
+The Video Source filter can sequentially read and play video files located inside a local directory in alphabetical order. 
+
+#### Directory Path Format
+```python
+sources='file:///path/to/directory'
+```
+
+#### Directory Playback Parameters
+When pointing `sources` to a folder, you can configure these optional parameters:
+- **`pattern`**: A glob/wildcard pattern (e.g. `*.mp4`) or regular expression to filter specific files.
+- **`recursive`**: If `True`, the filter scans subfolders recursively for matching videos.
+
+#### Directory Example
+```python
+# Process a folder of videos sequentially
+Filter.run_multi([
+    (VideoIn, dict(
+        sources='file:///home/user/videos_to_process',
+        outputs='tcp://*:5550',
+        pattern='*.mp4',    # Only read .mp4 files
+        recursive=True,     # Look inside subfolders
+        sync=True,          # Decode with sync
+    )),
+])
+```
+
 ## Video Processing Options
 
 ### Color Format (`bgr`)
@@ -714,6 +742,8 @@ class VideoInConfig(FilterConfig):
     fps: int | None
     maxsize: str | None
     resize: str | None
+    pattern: str | None
+    recursive: bool | None
 ```
 
 ### VideoIn
@@ -744,3 +774,5 @@ class VideoIn(Filter):
 - `FILTER_MAXFPS`: Maximum frame rate
 - `FILTER_MAXSIZE`: Maximum image size
 - `FILTER_RESIZE`: Image resize dimensions
+- `FILTER_PATTERN`: Glob/wildcard pattern for directory video files
+- `FILTER_RECURSIVE`: Scan directories recursively for matching videos
--- a/examples/directory-pipeline-demo/main.py
+++ b/examples/directory-pipeline-demo/main.py
@@ -1,0 +1,143 @@
+#!/usr/bin/env python3
+"""
+Directory Pipeline Demo
+
+This example demonstrates using the directory traversal feature in VideoIn to sequentialize
+a directory of video files and run them through a processing pipeline.
+"""
+
+import os
+import sys
+import shutil
+import tempfile
+import argparse
+from pathlib import Path
+
+import cv2
+import numpy as np
+
+from openfilter.filter_runtime.filter import Filter, FilterConfig
+from openfilter.filter_runtime.filters.video_in import VideoIn
+from openfilter.filter_runtime.filters.image_out import ImageOut
+
+
+class InfoPrinter(Filter):
+    """Prints frame metadata in real-time to show directory sequential playback."""
+    FILTER_TYPE = 'Processor'
+
+    def process(self, frames):
+        for topic, frame in frames.items():
+            meta = frame.data.get('meta', {})
+            src = meta.get('src', 'N/A')
+            frame_idx = meta.get('src_frame', 'N/A')
+            pts = meta.get('src_seconds', 'N/A')
+            
+            print(f"[InfoPrinter] Topic: {topic} | Active Source: {src} | Frame In Video: {frame_idx} | Offset Secs: {pts}")
+            
+        return frames
+
+
+def create_sample_video(p: str, color: str):
+    """Generate a lightweight 3-frame MP4 video of a solid color using cv2.VideoWriter."""
+    fourcc = cv2.VideoWriter_fourcc(*'mp4v')
+    out = cv2.VideoWriter(p, fourcc, 30.0, (100, 100))
+    img = np.zeros((100, 100, 3), dtype=np.uint8)
+    if color == 'red':
+        img[:, :, 2] = 255
+    elif color == 'green':
+        img[:, :, 1] = 255
+    elif color == 'blue':
+        img[:, :, 0] = 255
+    for _ in range(3):
+        out.write(img)
+    out.release()
+    print(f"Created sample video: {p}")
+
+
+def main():
+    parser = argparse.ArgumentParser(description="Run the directory pipeline demo.")
+    parser.add_argument("--directory", type=str, help="Directory of video files to read. If not provided, a temporary one with sample videos will be created.")
+    parser.add_argument("--loop", type=int, default=1, help="Number of times to loop (0 for infinite loop, 1 for once, etc.)")
+    args = parser.parse_args()
+
+    temp_dir = None
+    if not args.directory:
+        print("No directory provided. Creating a temporary directory with sample videos...")
+        temp_dir = tempfile.mkdtemp()
+        
+        # Create three separate video clips dynamically
+        vid_a = os.path.join(temp_dir, "01_autumn.mp4")
+        vid_b = os.path.join(temp_dir, "02_winter.mp4")
+        vid_c = os.path.join(temp_dir, "03_spring.mp4")
+        
+        create_sample_video(vid_a, 'red')
+        create_sample_video(vid_b, 'green')
+        create_sample_video(vid_c, 'blue')
+            
+        source_dir = temp_dir
+    else:
+        source_dir = os.path.abspath(args.directory)
+        if not os.path.isdir(source_dir):
+            print(f"Error: {source_dir} is not a valid directory.")
+            sys.exit(1)
+
+    output_dir = Path(__file__).parent / "output"
+    if output_dir.exists():
+        shutil.rmtree(output_dir)
+    output_dir.mkdir(parents=True, exist_ok=True)
+
+    print("\nDirectory Pipeline Demo")
+    print("=" * 60)
+    print(f"Source Directory: {source_dir}")
+    print(f"Output Frames:    {output_dir}")
+    print("=" * 60)
+    print("Starting pipeline: VideoIn -> InfoPrinter -> ImageOut")
+    print("Press Ctrl+C to exit.")
+    print("-" * 60)
+
+    try:
+        Filter.run_multi([
+            # 1. VideoIn: reads the folder
+            (VideoIn, FilterConfig({
+                'id': 'video-input',
+                'sources': f'file://{source_dir}!sync!maxfps=1',
+                'outputs': 'tcp://127.0.0.1:5570',
+            })),
+            
+            # 2. InfoPrinter: custom live printing filter
+            (InfoPrinter, FilterConfig({
+                'id': 'info-printer',
+                'sources': 'tcp://127.0.0.1:5570',
+                'outputs': 'tcp://127.0.0.1:5572',
+            })),
+            
+            # 3. ImageOut: saves the frames to disk
+            (ImageOut, FilterConfig({
+                'id': 'image-output',
+                'sources': 'tcp://127.0.0.1:5572',
+                'outputs': [
+                    f'file://{output_dir}/frame_%Y%m%d_%H%M%S_%d.png!format=png'
+                ],
+                'bgr': True,
+            }))
+        ])
+        
+        print("-" * 60)
+        print("Pipeline execution finished successfully.")
+        print(f"Output frames saved in: {output_dir}")
+        print("List of generated files:")
+        for entry in sorted(output_dir.iterdir()):
+            print(f"  - {entry.name}")
+
+    except KeyboardInterrupt:
+        print("\nPipeline interrupted by user.")
+    except Exception as e:
+        print(f"\nPipeline execution failed: {e}")
+        sys.exit(1)
+    finally:
+        if temp_dir and os.path.exists(temp_dir):
+            shutil.rmtree(temp_dir)
+
+
+if __name__ == '__main__':
+    main()
--- a/openfilter/filter_runtime/filters/video_in.py
+++ b/openfilter/filter_runtime/filters/video_in.py
@@ -29,6 +29,8 @@ VIDEO_IN_LOOP     = _ if isinstance(_ := json_getval((os.getenv('VIDEO_IN_LOOP')
 VIDEO_IN_MAXFPS   = None if (_ := json_getval((os.getenv('VIDEO_IN_MAXFPS') or os.getenv('FILTER_MAXFPS') or 'null').lower())) is None else float(_)
 VIDEO_IN_MAXSIZE  = os.getenv('VIDEO_IN_MAXSIZE') or os.getenv('FILTER_MAXSIZE') or None
 VIDEO_IN_RESIZE   = os.getenv('VIDEO_IN_RESIZE') or os.getenv('FILTER_RESIZE') or None
+VIDEO_IN_PATTERN  = os.getenv('VIDEO_IN_PATTERN') or os.getenv('FILTER_PATTERN') or None
+VIDEO_IN_RECURSIVE = bool(json_getval((os.getenv('VIDEO_IN_RECURSIVE') or os.getenv('FILTER_RECURSIVE') or 'false').lower()))
 
 # OpenCV's ffmpeg backend reports a sentinel ~1000 fps (the 1 ms MKV/webm container
 # timebase, not a real rate) for some VFR files. Any CAP_PROP_FPS at or above this
@@ -199,9 +201,9 @@ class VideoReader:
 
             expiration: Presigned URL expiration in seconds (optional).
 
-            pattern: Filter string or glob pattern (e.g., `*.mp4`) to select specific video files from directories.
+            pattern: Filter string or glob pattern (e.g., `*.mp4`) to select specific video files from directories. Has env var default.
 
-            recursive: If True, subfolders in directories are recursively scanned for video files.
+            recursive: If True, subfolders in directories are recursively scanned for video files. Has env var default.
         """
 
         if not isinstance((loop := VIDEO_IN_LOOP if loop is None else loop), (bool, int)) or loop < 0:
@@ -213,6 +215,8 @@ class VideoReader:
         self.maxfps        = maxfps = VIDEO_IN_MAXFPS if maxfps is None else maxfps
         self.maxsize       = None if (s := VIDEO_IN_MAXSIZE if maxsize is None else maxsize) is None else parse_size(s)
         self.resize        = None if (s := VIDEO_IN_RESIZE if resize is None else resize) is None else parse_size(s)
+        pattern            = VIDEO_IN_PATTERN if pattern is None else pattern
+        recursive          = VIDEO_IN_RECURSIVE if recursive is None else recursive
         self.state         = 0     # 0 = before start, 1 = playing, 2 = stopped / done
         self.sync_evt      = None  # this is set only for file fideo with 'sync' option True
         self.ns_per_fps    = None  # this is set only for file video with 'sync' option False
@@ -793,22 +797,23 @@ class VideoIn(Filter):
 
     config:
         sources:
-            The source(s) of the video(s), comma delimited, can be file://, rtsp:// stream, s3:// bucket object, 
-            or a webcam:// index.
+            The source(s) of the video(s), comma delimited, can be file://, rtsp:// stream, s3:// bucket object,
+            a webcam:// index, or a file:// directory path for sequential playback of the video files within it.
 
             Examples:
-                'file://a.mp4!sync!loop=3, rtsp://b.com!no-bgr;c, s3://bucket/video.mp4!region=us-west-2, webcam://0;e'
+                'file://a.mp4!sync!loop=3, rtsp://b.com!no-bgr;c, s3://bucket/video.mp4!region=us-west-2, webcam://0;e, file:///path/to/dir!pattern=*.mp4'
 
                     is the same as
 
-                ['file://a.mp4!sync!loop=3', 'rtsp://b.com!no-bgr;c', 's3://bucket/video.mp4!region=us-west-2', 'webcam://0;e']
+                ['file://a.mp4!sync!loop=3', 'rtsp://b.com!no-bgr;c', 's3://bucket/video.mp4!region=us-west-2', 'webcam://0;e', 'file:///path/to/dir!pattern=*.mp4']
 
                     is the same as
 
                 [{'source': 'file://a.mp4', 'topic': 'main', 'options': {'sync': True, 'loop': 3}},
                  {'source': 'rtsp://b.com', 'topic': 'c', 'options': {'bgr': False}},
                  {'source': 's3://bucket/video.mp4', 'topic': 'main', 'options': {'region': 'us-west-2'}},
-                 {'source': 'webcam://0', 'topic': 'e', 'options': {}}]
+                 {'source': 'webcam://0', 'topic': 'e', 'options': {}},
+                 {'source': 'file:///path/to/dir', 'topic': 'main', 'options': {'pattern': '*.mp4'}}]
 
                     For 'options' see below.
 
@@ -884,10 +889,13 @@ class VideoIn(Filter):
 
         pattern:
             Only has meaning for folder/directory file:// sources. A glob pattern (e.g. `*.mp4`) or regular expression
-            to match and filter video files in the directory. Set here to apply to all sources or can be set individually per source.
+            to match and filter video files in the directory. Set here to apply to all sources or can be set individually
+            per source. Global env var default FILTER_PATTERN / VIDEO_IN_PATTERN.
 
         recursive:
-            Only has meaning for folder/directory file:// sources. If True, scans the directory and its subdirectories recursively for matching files. Set here to apply to all sources or can be set individually per source.
+            Only has meaning for folder/directory file:// sources. If True, scans the directory and its subdirectories
+            recursively for matching files. Set here to apply to all sources or can be set individually per source.
+            Global env var default FILTER_RECURSIVE / VIDEO_IN_RECURSIVE.
 
         override_source_uri:
             Logical source URI to report as each frame's meta['src'] instead of the physical source opened by the
@@ -909,6 +917,9 @@ class VideoIn(Filter):
         With `loop`, the cap is reopened each pass, so src_frame and src_seconds restart at 0 every loop (they are the
         position WITHIN the file) while meta['id'] keeps counting across passes: a looped timeline is non-monotonic
         in src_frame/src_seconds but monotonic in id, so consumers indexing a looped source must key off id, not src_frame.
+        A directory source reproduces this same restart at every file transition even without `loop`: src_frame and
+        src_seconds reset to 0 (and meta['src'] changes) each time playback moves to the next file in the directory,
+        while meta['id'] keeps counting across the whole run - so consumers of a directory source must key off id too.
 
         Stream and webcam sources have no meaningful decoder position: both keys are absent, all other meta unchanged.
 
@@ -919,6 +930,8 @@ class VideoIn(Filter):
         FILTER_MAXFPS   / VIDEO_IN_MAXFPS
         FILTER_MAXSIZE  / VIDEO_IN_MAXSIZE
         FILTER_RESIZE   / VIDEO_IN_RESIZE
+        FILTER_PATTERN  / VIDEO_IN_PATTERN
+        FILTER_RECURSIVE / VIDEO_IN_RECURSIVE
 
     S3 Configuration:
         For s3:// sources, AWS credentials are required. Set these environment variables:
--- a/openfilter/filter_runtime/filters/video_in.py
+++ b/openfilter/filter_runtime/filters/video_in.py
@@ -1,3 +1,4 @@
+import glob
 import logging
 import os
 import re
@@ -43,6 +44,25 @@ is_video_file     = lambda name: name.startswith('file://')
 is_video_webcam   = lambda name: name.startswith('webcam://')
 is_video_stream   = lambda name: bool(re_video_stream.match(name))
 is_video_s3       = lambda name: name.startswith('s3://')
+
+VIDEO_EXTENSIONS = {'mp4', 'mkv', 'avi', 'mov', 'webm', 'flv', 'wmv', 'm4v', '3gp', 'mpg', 'mpeg'}
+
+def is_video_file_ext(path: str) -> bool:
+    """Check if the given file path has a valid video file extension."""
+    ext = path.split('.')[-1].lower() if '.' in path else ''
+    return ext in VIDEO_EXTENSIONS
+
+def matches_pattern(path: str, pattern: str) -> bool:
+    """Check if path matches the given pattern (glob or regex)."""
+    if not pattern:
+        return True
+    if '*' in pattern or '?' in pattern:
+        return glob.fnmatch.fnmatch(os.path.basename(path), pattern)
+    try:
+        return bool(re.search(pattern, path))
+    except re.error:
+        return pattern in path
+
 
 
 re_size = re.compile(r'^\s* (\d+) \s* ([x+]) \s* (\d+) \s* (n(?:ear)? | l(?:in)? | c(?:ub)?)? \s*$', re.VERBOSE | re.IGNORECASE)
@@ -134,6 +154,8 @@ class VideoReader:
         resize:  str | None = None,
         region:  str | None = None,
         expiration: int | None = None,
+        pattern:  str | None = None,
+        recursive: bool | None = None,
     ):
         """Read a single video file, network stream or webcam until the end.
 
@@ -175,6 +197,10 @@ class VideoReader:
         self.ns_per_maxfps = None if maxfps is None else 1_000_000_000 // maxfps
         self.is_file       = is_file = is_video_file(source) or is_video_s3(source)
         self.as_bgr        = bool(VIDEO_IN_BGR if bgr is None else bgr)  # only validated after first frame is read (set to False if frames are grayscale)
+        self.is_dir        = False
+        self.dir_files     = []
+        self.dir_index     = 0
+        self.source_dir_path = None
 
         if self.maxsize and self.resize:
             raise ValueError(f"can not specify both 'maxsize' and 'resize' together in {self.source!r}")
@@ -192,6 +218,31 @@ class VideoReader:
                 except Exception as e:
                     raise ValueError(f'Failed to generate presigned URL for S3 source {self.source!r}: {e}')
             self.is_file = True
+
+            if os.path.isdir(source):
+                self.is_dir = True
+                self.source_dir_path = self.source
+                
+                # Scan directory with pattern and recursive options
+                if recursive:
+                    glob_pat = os.path.join(source, '**', '*')
+                    candidates = glob.glob(glob_pat, recursive=True)
+                else:
+                    glob_pat = os.path.join(source, '*')
+                    candidates = glob.glob(glob_pat)
+
+                self.dir_files = []
+                for p in sorted(candidates):
+                    if os.path.isfile(p) and is_video_file_ext(p) and matches_pattern(p, pattern):
+                        self.dir_files.append(p)
+
+                if not self.dir_files:
+                    raise RuntimeError(f'no valid video files found in directory: {self.source!r}')
+                self.dir_index = 0
+                source = self.dir_files[0]
+                self.source = f'file://{source}'
+            else:
+                self.is_dir = False
 
             if sync := VIDEO_IN_SYNC if sync is None else sync:
                 self.sync_evt = Event()
@@ -313,6 +364,26 @@ class VideoReader:
 
         return ret, image
 
+    def _open_dir_file(self, index: int):
+        self.dir_index = index
+        active_source = self.dir_files[index]
+        self.cap.release()
+        self.cap = cv2.VideoCapture(active_source)
+        if not self.cap.isOpened():
+            raise RuntimeError(f'failed to open video in directory: {active_source}')
+
+        self.source = f'file://{active_source}'
+        self.ssource = active_source
+        fps = self.cap.get(cv2.CAP_PROP_FPS) or None
+        self.native_fps = fps
+        self.fps = fps
+        if self.maxfps is not None and fps is not None and fps > self.maxfps:
+            if not self.is_file or not self.sync_evt:
+                self.fps = self.maxfps
+
+        if not self.sync_evt:
+            self.ns_per_fps = 1_000_000_000 // (self.fps or 15)
+
     def read_one(self):
         def wait() -> bool:  # returns if should return frame or keep looping
             t = time_ns()
@@ -360,6 +431,31 @@ class VideoReader:
                 # frame may be lost on a realtime video if it takes too long to query it but it will never be lost on
                 # a `sync` video.
 
+                if self.is_dir and self.dir_index + 1 < len(self.dir_files):
+                    success = False
+                    while self.dir_index + 1 < len(self.dir_files):
+                        next_index = self.dir_index + 1
+                        active_source = self.dir_files[next_index]
+                        try:
+                            logger.info(f'video transition: opening next file in directory: {active_source}')
+                            self._open_dir_file(next_index)
+                            ret, image = self._cap_read()
+                            if ret and image is not None:
+                                success = True
+                                break
+                            self.dir_index = next_index
+                        except Exception as e:
+                            logger.error(f'Error during video transition to {active_source}: {e}')
+                            self.dir_index = next_index
+
+                    if success:
+                        if wait():
+                            break
+                        continue
+
+                    # If we ran out of directory files, fall through to loop/restart logic below
+
+                # Handle loop/restart
                 if loop := self.loop:
                     self.loop = loop - 1
 
@@ -369,21 +465,39 @@ class VideoReader:
                         return None
 
                 try:
-                    logger.info(f'video loop: {self.source}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                    if self.is_dir:
+                        logger.info(f'directory loop: {self.source_dir_path}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                        success = False
+                        self.dir_index = 0
+                        while self.dir_index < len(self.dir_files):
+                            active_source = self.dir_files[self.dir_index]
+                            try:
+                                self._open_dir_file(self.dir_index)
+                                ret, image = self._cap_read()
+                                if ret and image is not None:
+                                    success = True
+                                    break
+                                self.dir_index += 1
+                            except Exception as e:
+                                logger.error(f'Error during directory loop restart for {active_source}: {e}')
+                                self.dir_index += 1
 
-                    self.cap.release()
-                    self.cap = cv2.VideoCapture(self.ssource)
-                    if not self.cap.isOpened():
-                        raise RuntimeError(f'failed to reopen video source: {self.source!r}')
+                        if not success:
+                            return None
+                    else:
+                        logger.info(f'video loop: {self.source}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                        self.cap.release()
+                        self.cap = cv2.VideoCapture(self.ssource)
+                        if not self.cap.isOpened():
+                            raise RuntimeError(f'failed to reopen video source: {self.ssource}')
+                        ret, image = self._cap_read()
+                        if not ret or image is None:
+                            return None
 
-                except Exception:
+                except Exception as e:
+                    logger.error(f'Error during loop restart: {e}')
                     wait()
 
-                    return None
-
-                ret, image = self._cap_read()
-
-                if not ret or image is None:  # no wait() here because if first frame fails then nothing means anything anymore and we might as well just end it
                     return None
 
             if wait():
@@ -450,7 +564,10 @@ class VideoReader:
                 elif not self.as_bgr:
                     image = cv2.cvtColor(image, cv2.COLOR_RGB2BGR)
 
-            self.deque.append((image, tframe, self.extras))
+            extras_copy = dict(self.extras) if self.extras else {}
+            extras_copy['source'] = self.source
+            extras_copy['fps'] = self.fps
+            self.deque.append((image, tframe, extras_copy))
 
             if cond is not None:
                 with cond:
@@ -496,7 +613,27 @@ class VideoReader:
 
     @staticmethod
     def get_info(source: str) -> tuple[int, int, str, float]:  # (height, width, format, fps)
-        cap = cv2.VideoCapture(int(source[9:]) if is_video_webcam(source) else source[7:] if is_video_file(source) else source)
+        path = source[7:] if is_video_file(source) else source
+        if os.path.isdir(path):
+            # Find the first valid video file in the directory
+            with os.scandir(path) as entries:
+                for entry in sorted(entries, key=lambda e: e.name):
+                    if entry.is_file() and is_video_file_ext(entry.path):
+                        test_cap = cv2.VideoCapture(entry.path)
+                        if test_cap.isOpened():
+                            ret, image = test_cap.read()
+                            if ret:
+                                width  = image.shape[1]
+                                height = image.shape[0]
+                                fps    = test_cap.get(cv2.CAP_PROP_FPS)
+                                is_bgr = test_cap.get(cv2.CAP_PROP_CONVERT_RGB)
+                                format = 'GRAY' if len(image.shape) == 2 else 'BGR' if is_bgr else 'RGB'
+                                test_cap.release()
+                                return height, width, format, fps
+                            test_cap.release()
+            raise RuntimeError(f'no valid video files found in directory: {source}')
+
+        cap = cv2.VideoCapture(int(source[9:]) if is_video_webcam(source) else path)
 
         try:
             ret, image = cap.read()
@@ -608,6 +745,8 @@ class VideoInConfig(FilterConfig):
             resize:     str | None
             region:     str | None
             expiration: int | None
+            pattern:    str | None
+            recursive:  bool | None
 
         source:  str
         topic:   str | None
@@ -622,6 +761,8 @@ class VideoInConfig(FilterConfig):
     maxfps:  float | None
     maxsize: str | None
     resize:  str | None
+    pattern:    str | None
+    recursive:  bool | None
 
     # Logical source URI to report as meta['src'] instead of the physical source
     # (set by an orchestrator; see resolve_override_source_uri). FILTER_OVERRIDE_SOURCE_URI
@@ -796,7 +937,7 @@ class VideoIn(Filter):
                 source.topic = 'main'
             if not isinstance(options := source.options, VideoInConfig.Source.Options):
                 source.options = options = VideoInConfig.Source.Options() if options is None else VideoInConfig.Source.Options(options)
-            if any((option := o) not in ('bgr', 'sync', 'loop', 'maxfps', 'maxsize', 'resize', 'region', 'expiration') for o in options):
+            if any((option := o) not in ('bgr', 'sync', 'loop', 'maxfps', 'maxsize', 'resize', 'region', 'expiration', 'pattern', 'recursive') for o in options):
                 raise ValueError(f'unknown option {option!r} in {source!r}')
 
         if len(set(source.topic for source in sources)) != len(sources):
@@ -820,7 +961,7 @@ class VideoIn(Filter):
             optionss.append(source.options or {})
 
         default_options      = {'bgr': config.bgr, 'sync': config.sync, 'loop': config.loop, 'maxfps': config.maxfps,
-            'maxsize': config.maxsize, 'resize': config.resize}
+            'maxsize': config.maxsize, 'resize': config.resize, 'pattern': config.get('pattern'), 'recursive': config.get('recursive')}
         self.mvreader        = MultiVideoReader(vsources, [{**default_options, **options} for options in optionss])
         self.tops_n_vids     = tuple(zip(topics, self.mvreader.videos))
         self.id              = -1  # frame id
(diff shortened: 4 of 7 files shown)


openfilter/filter_runtime/filters/video_in.py
```diff
@@ -360,31 +378,81 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
+                            raise RuntimeError(f'failed to reopen video source: {active_source}')
+
+                        if self.is_dir:
+                            self.source = f'file://{active_source}'
+                            self.ssource = active_source
+                            fps = self.cap.get(cv2.CAP_PROP_FPS) or None
+                            self.native_fps = fps
+                            self.fps = fps
+                            if self.maxfps is not None and fps is not None and fps > self.maxfps:
+                                self.fps = self.maxfps
```
lucasmundim: **Blocking.** `src_fps` regresses to `maxfps` after the first directory transition in `sync` mode.

`__init__` deliberately guards this: `if not is_file or not sync: self.fps = fps` — i.e. under `sync` the reader keeps reporting the *native* rate, which the `sync` docstring states as a contract ("the original fps is passed downstream so that any video created from this file will have the original framerate"). Both directory paths (here and line 397) drop the `not sync` guard, so `self.fps` — and therefore every frame's `meta['src_fps']` — flips to `maxfps` from the second file on.

Reproduced on this branch: `file://<dir>!sync!maxfps=1` yields `src_fps` 30.0 for the first clip and 1 for every clip after it, while the equivalent single-file `!sync!maxfps=1!loop=2` correctly stays at 30.0 across passes. A downstream `VideoOut` would then write the tail of the stream at the wrong framerate.

Note `examples/directory-pipeline-demo/main.py:122` uses exactly `!sync!maxfps=1`, so the demo shipped with this PR hits it.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -360,31 +378,81 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
+                    self.dir_index += 1
+                    active_source = self.dir_files[self.dir_index]
+                    try:
+                        logger.info(f'video transition: opening next file in directory: {active_source}')
+                        self.cap.release()
+                        self.cap = cv2.VideoCapture(active_source)
+                        if not self.cap.isOpened():
+                            raise RuntimeError(f'failed to open next video in directory: {active_source}')
+
+                        self.source = f'file://{active_source}'
```
lucasmundim: **Blocking.** `self.source` / `self.fps` are mutated by the reader thread but read by the consumer thread, so frames can be attributed to the wrong file.

`VideoIn.process` builds meta from `vid.source` / `vid.fps` *after* `mvreader.read()` returns, while the reader thread — which runs ahead under `sync` — may already have crossed into the next file and rewritten both. Nothing synchronizes the two.

Reproduced on this branch by delaying the consumer 50 ms between `read()` and reading `reader.source`:

```
frame 0: src_frame=0 meta_src=video_a.mp4
frame 1: src_frame=1 meta_src=video_b.mp4   <-- still video_a's frame
frame 2: src_frame=2 meta_src=video_b.mp4   <-- still video_a's frame
frame 3: src_frame=0 meta_src=video_b.mp4
```

With no delay the labels happen to come out right (which is why the new tests pass), but that is a race the code does not win by construction — and correct per-file `meta['src']` is the whole point of the feature.

Suggested fix: put the active source (and fps) into `self.extras` in `_cap_read`, alongside `frame_n` / `pts_s`. `extras` is already snapshotted per frame into the deque tuple, so consumers would get the source of the frame they actually hold, and `meta()` would read it from `extras` rather than from live reader state.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -193,6 +193,24 @@ def __init__(self,
                     raise ValueError(f'Failed to generate presigned URL for S3 source {self.source!r}: {e}')
             self.is_file = True
 
+            if os.path.isdir(source):
```
lucasmundim, line 251: **Blocking.** A directory source silently defeats `override_source_uri` per-file attribution.

`VideoIn.setup` gates the override on `len(self.mvreader.videos) <= 1`, which reads as "a single source == a single logical file". A directory breaks that assumption: it is one `VideoReader` covering N files, so the gate passes and every file in the directory is stamped with the same `meta['src']` — the exact mislabeling the gate exists to prevent.

`ImageIn` already hit this and guards it explicitly (see its `_apply_override` comment and `_is_single_object_source`): a directory, a glob, or multiple sources disables the override because stamping every match with the same override URI would mislabel them. `VideoIn` needs the same treatment now that it can be pointed at a directory — `_apply_override` should also require that the reader is not a directory, with the same warning, plus a test.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -193,6 +193,24 @@ def __init__(self,
...
+                self.is_dir = True
+                self.source_dir_path = self.source
+                self.dir_files = []
+                for entry in sorted(os.scandir(source), key=lambda e: e.name):
+                    if entry.is_file():
+                        test_cap = cv2.VideoCapture(entry.path)
+                        if test_cap.isOpened():
```
lucasmundim: **Blocking.** `isOpened()` is not a video test, and this scan diverges from the directory contract `ImageIn` already established in this repo.

`ImageIn` handles `file://` directories with an explicit extension allowlist (`IMAGE_EXTENSIONS`), plus `!pattern=*.jpg` (glob or regex), `!recursive` and `!poll_interval` options, all documented in its class docstring. This scan instead probes every file with `cv2.VideoCapture` and keeps whatever opens, which is depth-1 only, has no way to filter, and is not a video test: OpenCV opens .jpg/.png/.bmp/.tiff through its image backend.

Reproduced on this branch with `aaa_thumbnail.jpg` (32x32) + `zzz_video.mp4` in one directory: 4 frames instead of 3, and `VideoReader.get_info('file://<dir>')` returned `(32, 32, 'BGR', 25.0)` — the thumbnail's geometry becomes the declared geometry for the whole stream, since `get_info` takes the first file that opens.

`test_directory_skips_non_videos` only proves a `.txt` is skipped, which OpenCV rejects for unrelated reasons.

Mirroring `ImageIn` (a `VIDEO_EXTENSIONS` allowlist plus `pattern` / `recursive` options, added to the `normalize_config` option whitelist and the docstring) would fix the correctness issue and keep the two input filters consistent for users. It would also remove the O(n) `VideoCapture` opens at startup, which are slow on large or network-backed directories.

Minor, same block: `sorted(os.scandir(source), ...)` leaves the `ScandirIterator` unclosed (ResourceWarning) — `os.listdir` or a `with` block avoids that.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -360,31 +378,81 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
 
-                try:
-                    logger.info(f'video loop: {self.source}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                    ret, image = self._cap_read()
 
-                    self.cap.release()
-                    self.cap = cv2.VideoCapture(self.ssource)
-                    if not self.cap.isOpened():
-                        raise RuntimeError(f'failed to reopen video source: {self.source!r}')
+                    if not ret or image is None:
```
lucasmundim: Non-blocking. A single bad file ends the whole stream instead of being skipped.

If the next file fails to open, or opens but its first read fails, this returns `None` and terminates playback — the remaining files in the directory are never played. Since the init scan validates only `isOpened()`, a truncated or undecodable file (or one deleted between scan and playback) silently truncates the run with an ERROR log the pipeline never surfaces. Advancing to the next entry instead would match how the scan already ignores bad files.

Same block: the `raise RuntimeError('failed to open next video in directory')` on line 389 is caught by its own `except` four lines later, so the raise/catch round-trip only turns into a log line — a plain log plus advance would read better.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -360,31 +378,81 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
+                        else:
+                            logger.info(f'video loop: {self.source}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                            active_source = self.ssource
+
+                        self.cap.release()
+                        self.cap = cv2.VideoCapture(active_source)
+                        if not self.cap.isOpened():
+                            raise RuntimeError(f'failed to reopen video source: {active_source}')
+
+                        if self.is_dir:
```
lucasmundim: Non-blocking. Lines 437-445 duplicate 391-400 exactly. Since the loop restart is just "open `dir_files[0]`" and the transition is "open `dir_files[dir_index]`", both could call one `_open_dir_file(index)` helper — which would also mean the `sync` / `maxfps` fix from the first comment only has to be made in one place.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -193,6 +193,24 @@ def __init__(self,
...
+                        if test_cap.isOpened():
+                            self.dir_files.append(entry.path)
+                        test_cap.release()
+                if not self.dir_files:
+                    raise RuntimeError(f'no valid video files found in directory: {self.source!r}')
+                self.dir_index = 0
+                source = self.dir_files[0]
+                self.source = f'file://{source}'
+            else:
+                self.is_dir = False
```
lucasmundim, line 263: Non-blocking. `self.is_dir` is assigned only inside the `if is_file:` branch, so stream and webcam readers never get the attribute at all. It is safe today only because `if not self.is_file: return None` short-circuits before the `self.is_dir` check in `read_one` — a fragile invariant. Initialize it (and `dir_files` / `dir_index`) with the other attributes at the top of `__init__`, next to `self.is_file`.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -496,7 +564,26 @@ def read(self, with_tframe=False):  # -> np.ndarray | tuple[np.ndarray, int, dic
 
     @staticmethod
     def get_info(source: str) -> tuple[int, int, str, float]:  # (height, width, format, fps)
-        cap = cv2.VideoCapture(int(source[9:]) if is_video_webcam(source) else source[7:] if is_video_file(source) else source)
+        path = source[7:] if is_video_file(source) else source
+        if os.path.isdir(path):
```
lucasmundim, line 636: Non-blocking. This directory branch has no caller anywhere in the repo (`get_info` is unreferenced outside its own definition) and no test. It also re-implements the init scan with the same `isOpened()` weakness. Either drop it, or share the scan with `__init__` and cover it with a test — as it stands it is untested public API that will drift from the real scan.

tests/test_filter_video_in.py
```diff
@@ -836,6 +836,156 @@ def test_config_params(self):
             runner.stop()
             queue.close()
 
+    def test_directory_source(self):
+        import shutil
```
lucasmundim: Non-blocking. `import shutil` / `import tempfile` are repeated inside each of the three new tests; the module already imports at the top. The three setups are also near-identical — a small helper that writes N clips into a temp dir would shrink all three.

Coverage gaps worth adding: the `sync` + `maxfps` fps behavior across a transition (the regression flagged in `video_in.py`); `override_source_uri` combined with a directory source; a directory containing an image file; files with differing resolution or fps; and the `RuntimeError` path for a directory with no valid video.

examples/directory-pipeline-demo/main.py
```diff
@@ -0,0 +1,161 @@
...
+        print("-" * 60)
+        print("Pipeline execution finished successfully.")
+        print(f"Output frames saved in: {output_dir}")
+        print("List of generated files:")
+        for entry in sorted(output_dir.iterdir()):
+            print(f"  - {entry.name}")
+
+    except KeyboardInterrupt:
+        print("\nPipeline interrupted by user.")
+    except Exception as e:
```
lucasmundim, line 134: Non-blocking. `except Exception as e: print(...)` swallows the failure and the demo still exits 0, so a broken pipeline looks like a successful run. Re-raise or `sys.exit(1)`.

A few more on this file:
- No `README.md` — every other directory under `examples/` has one.
- `sys.path.insert(0, <repo root>)` on line 10 shouldn't be needed; the other examples import `openfilter` as an installed package.
- The ~60-line MP4 byte blob is copy-pasted from `tests/test_filter_video_in.py`; the demo could generate the clips with `cv2.VideoWriter` instead of embedding a literal.
- Line 122 uses `!sync!maxfps=1`, which trips the `src_fps` regression flagged in `video_in.py`.

lucasmundim (changes requested):
Nice feature and the tests cover the happy path well. Four functional issues below are worth fixing before merge — three of them I reproduced locally against this branch — plus the two red required checks.

The common root of the first two is that per-frame identity (`meta['src']`, `meta['src_fps']`) is derived from mutable `VideoReader` state that the reader thread now rewrites mid-stream. Everything else the consumer needs per frame (`frame_n`, `pts_s`) is already snapshotted into `self.extras` and appended with the image; `src` and `fps` should travel the same way.

Repro used below (this branch, `sync` mode, two copies of the test clip in a temp dir):

```
dir + sync + maxfps=1 -> (frame_n, src_fps):
  [(0, 30.0), (1, 30.0), (2, 30.0), (0, 1), (1, 1), (2, 1)]
single file + sync + maxfps=1 + loop=2 (unchanged path):
  [(0, 30.0), (1, 30.0), (2, 30.0), (0, 30.0), (1, 30.0), (2, 30.0)]
```

Worth reading `ImageIn` before revising: it already supports `file://` directory sources in this repo, with an extension allowlist plus `!pattern`, `!recursive` and `!poll_interval` options, and it already reasons explicitly about how a directory interacts with `override_source_uri`. Two of the comments below come straight from that comparison.

Also flagging: the feature is not documented anywhere. The `VideoIn` class docstring still lists only "file://, rtsp:// stream, s3:// bucket object, or a webcam:// index", `VideoReader.__init__` still says "Read a single video file", and `docs/video-in-filter.md` is untouched. Directory semantics deserve a note in the "Emitted frame meta" section too, since `src_frame` / `src_seconds` now restart per file within a single non-looping run — the same non-monotonicity the `loop` paragraph already warns about.

Separately — let's sync offline on where this fits scope-wise before you invest more in the demo; nothing blocking about the code direction.

Non-blocking nits are marked as such.

lucasmundim:
Two required checks are red and both need a push from you:

- **DCO** — the commit needs a `Signed-off-by` trailer. Amend with `git commit -s --amend` and force-push.
- **check-release-log** — `RELEASE.md must be updated on every PR to main`. Per the repo convention, add the entry under `## [Unreleased]` and leave the `VERSION` bump to a separate release PR.

shingonoide (changes requested):
This is a change to a shared open-source core filter, and I don't see an agreed design doc, ADR, or ticket anywhere (PR body, commit message, comments, reviews) sanctioning directory-based sources as a `VideoIn` capability. For new capabilities like this in a shared core filter, it's worth writing up the direction and getting it agreed before landing the code, rather than after.

Two things worth resolving:

1. Scope: this modifies the core `VideoIn` filter directly rather than adding a standalone filter. `ImageIn` already has an equivalent capability (extension allowlist, `!pattern`, `!recursive`, `!poll_interval`) as its own feature on that filter, so there's precedent for folding it into an existing filter. Worth reviewing the existing OpenFilter documentation and `ImageIn`'s approach closely before deciding which path this should take, so the implementation doesn't diverge from an established pattern or duplicate work unnecessarily.
2. lucasmundim's review already flags that the feature isn't documented anywhere (VideoIn docstring, VideoReader docstring, docs/video-in-filter.md).

Could you write up a short design note (a Confluence page works) covering the direction and the points above, and have it reviewed there before continuing? That will help make sure the approach lands right the first time.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Consolidated directory reopen and transition paths under a single `_open_dir_file(self, index)` helper. We now correctly evaluate `not self.sync_evt` before applying any `maxfps` overrides, preserving native frame rate downstream across transitions in sync mode.                                                               
 
 Verified via `test_sync_mode_fps_preservation_across_transition`.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Fixed as you suggested

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Gated `_apply_override` in `VideoIn.setup()` to explicitly require that the source is not a directory (`and not self.mvreader.videos[0].is_dir`), logging an explanatory warning.

Covered by `test_override_source_uri_rejected_for_directory`.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Replaced O(n) OpenCV startup probing with an extension-based allowlist (`VIDEO_EXTENSIONS`) and implemented glob scanning in `VideoReader.__init__` supporting both `pattern` and `recursive` options. Whitelisted these options in `VideoInConfig`. Wrapped `os.scandir` in `get_info()` inside a `with` block to prevent `ResourceWarning`.

Covered by `test_directory_pattern_and_recursive_options`.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Restructured both the video transition block and the directory loop restart block in `VideoReader.read_one()` to catch failures, log them, increment `self.dir_index` properly, and safely skip bad/deleted/truncated files.                       
       
Also guaranteed that `self.dir_index` is always incremented when a file fails to read frames, even if it does not throw an exception (e.g. truncated videos returning `False, None`).

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Created `_open_dir_file(self, index)` helper method.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Moved all core tracking variables (`self.is_dir`, `self.dir_files`, `self.dir_index`, `self.source_dir_path`) to the top of `__init__`

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Refactored `get_info` to use `is_video_file_ext`, wrapped `os.scandir` in a `with` block to prevent leaks.
Added `test_directory_get_info` test

tests/test_filter_video_in.py
ahonjaya-del: Cleaned up repeated inline test imports by moving `shutil` and `tempfile` to the top of the test suite.

examples/directory-pipeline-demo/main.py
ahonjaya-del: Rewrote the demo generator script to use `cv2.VideoWriter`, 

Added fail-exits/re-raises (`sys.exit(1)`), cleaned up imports, and added a `README.md` for running the directory demo.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -313,6 +364,26 @@ def _cap_read(self):
...
+        self.ssource = active_source
+        fps = self.cap.get(cv2.CAP_PROP_FPS) or None
+        self.native_fps = fps
+        self.fps = fps
+        if self.maxfps is not None and fps is not None and fps > self.maxfps:
+            if not self.is_file or not self.sync_evt:
+                self.fps = self.maxfps
+
+        if not self.sync_evt:
+            self.ns_per_fps = 1_000_000_000 // (self.fps or 15)
```
lucasmundim: **Blocking.** Pacing regresses to `maxfps` after the first directory transition when `sync` is off.

`__init__` sets `ns_per_fps` from the **native** rate — line 273 runs *before* the `maxfps` cap below it — and `maxfps` is enforced separately by the frame-skipping in `wait()` via `ns_per_maxfps`. Here it is recomputed from `self.fps`, which line 382 has just set to `maxfps`, so from the second file on the decoder is paced at `maxfps` instead of realtime: the video plays in slow motion and `wait()` stops skipping frames, since every decoded frame now arrives already >= `ns_per_maxfps` apart.

Reproduced on this branch (dir of two copies of the 3-frame/30 fps test clip, `!maxfps=1`, no `!sync`): `ns_per_fps` 33,333,333 after init -> 1,000,000,000 after the transition; 4 frames over 4.14s, three of them the whole of `b.mp4` at one per second. The equivalent single-file `!maxfps=1!loop=2` keeps 33,333,333 and emits 1 frame in 0.24s.

`self.native_fps or 15` instead of `self.fps or 15` restores `__init__`'s behavior. Worth a test asserting frame count/pacing across a transition in non-`sync` mode — `test_sync_mode_fps_preservation_across_transition` only covers the `sync` side.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -369,23 +465,41 @@ def wait() -> bool:  # returns if should return frame or keep looping
...
+                                self.dir_index += 1
+
+                        if not success:
+                            return None
+                    else:
+                        logger.info(f'video loop: {self.source}{f"  (last loop)" if loop == 2 else f"  ({self.loop} left)" if loop else ""}')
+                        self.cap.release()
+                        self.cap = cv2.VideoCapture(self.ssource)
+                        if not self.cap.isOpened():
+                            raise RuntimeError(f'failed to reopen video source: {self.ssource}')
```
lucasmundim: This touches a path the PR doesn't otherwise need, and it leaks a credential into the logs.

`self.ssource` for an `s3://` source is the presigned HTTPS URL, signature and all; `self.source` is the original `s3://` URI passed through `hide_uri_users_and_pwds`, which is why the original message used it. Combined with the new `except` below (main swallowed this exception silently, this logs it), a failed loop restart on an S3 video now writes a still-valid presigned URL to ERROR-level logs.

Please put `{self.source!r}` back.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -796,7 +937,7 @@ def normalize_config(cls, config):
                 source.topic = 'main'
             if not isinstance(options := source.options, VideoInConfig.Source.Options):
                 source.options = options = VideoInConfig.Source.Options() if options is None else VideoInConfig.Source.Options(options)
-            if any((option := o) not in ('bgr', 'sync', 'loop', 'maxfps', 'maxsize', 'resize', 'region', 'expiration') for o in options):
+            if any((option := o) not in ('bgr', 'sync', 'loop', 'maxfps', 'maxsize', 'resize', 'region', 'expiration', 'pattern', 'recursive') for o in options):
```
lucasmundim, line 984: **Blocking (docs).** `pattern` and `recursive` are accepted here and documented nowhere.

The `VideoIn` docstring still describes `sources` as "file://, rtsp:// stream, s3:// bucket object, or a webcam:// index" with no directory, and its `!option` list ends at `!expiration` — a user can only discover either option by reading this tuple. `VideoReader.__init__` (line 160) still opens with "Read a single video file" and documents neither kwarg. `docs/video-in-filter.md` has no mention of directory sources at all.

The "Emitted frame meta" section needs a directory note too: `src_frame` / `src_seconds` now restart at 0 at every file boundary within a single non-looping run — exactly the non-monotonicity the `loop` paragraph already warns about — and `meta['src']` now changes mid-stream. Consumers keying off those need to be told. `ImageIn`'s docstring is a ready-made template for both the option entries and the semantics wording.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -450,7 +564,10 @@ def thread_reader(self):
                 elif not self.as_bgr:
                     image = cv2.cvtColor(image, cv2.COLOR_RGB2BGR)
 
-            self.deque.append((image, tframe, self.extras))
+            extras_copy = dict(self.extras) if self.extras else {}
+            extras_copy['source'] = self.source
+            extras_copy['fps'] = self.fps
```
lucasmundim, line 587: Non-blocking. The `read()` docstring below still says "extras is `{}` for non-file sources", which these three lines make untrue — stream and webcam frames now carry `{'source': ..., 'fps': ...}`. `meta()` was correctly updated to gate on `'frame_n' in extras`, so nothing breaks, but the comment now documents the opposite of what happens.

Minor, separately: the keys are `source`/`fps` here but surface as `src`/`src_fps` in meta, while `frame_n`/`pts_s` map to `src_frame`/`src_seconds` — the two halves of the dict follow different naming conventions.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -496,7 +613,27 @@ def read(self, with_tframe=False):  # -> np.ndarray | tuple[np.ndarray, int, dic
 
     @staticmethod
     def get_info(source: str) -> tuple[int, int, str, float]:  # (height, width, format, fps)
-        cap = cv2.VideoCapture(int(source[9:]) if is_video_webcam(source) else source[7:] if is_video_file(source) else source)
+        path = source[7:] if is_video_file(source) else source
+        if os.path.isdir(path):
```
lucasmundim, line 636: Non-blocking. This branch still re-implements the scan rather than sharing it with `__init__`, and it is now visibly divergent: `__init__` honors `pattern` and `recursive`, this walks depth-1 and ignores both. For `file://dir!recursive` or `!pattern=*.cam1.mp4` it can therefore report the geometry and fps of a file the reader will never play. `get_info` still has no caller in the repo, so this is latent rather than live — extracting the file list into a small helper that both call would settle it.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -44,6 +45,25 @@
...
 is_video_s3       = lambda name: name.startswith('s3://')
 
+VIDEO_EXTENSIONS = {'mp4', 'mkv', 'avi', 'mov', 'webm', 'flv', 'wmv', 'm4v', '3gp', 'mpg', 'mpeg'}
+
+def is_video_file_ext(path: str) -> bool:
+    """Check if the given file path has a valid video file extension."""
+    ext = path.split('.')[-1].lower() if '.' in path else ''
+    return ext in VIDEO_EXTENSIONS
+
+def matches_pattern(path: str, pattern: str) -> bool:
```
lucasmundim, line 58: Non-blocking. `is_video_file_ext` and `matches_pattern` are `image_in.py`'s `is_image_file_ext` / `matches_pattern` copied verbatim (the latter identical line for line). Mirroring `ImageIn`'s *semantics* was the ask, but the second copy of the body means a future fix to the glob/regex fallback has to be made twice — `matches_pattern` is source-type agnostic and would sit fine in `filter_runtime/utils.py` with both filters importing it.

Also `glob.fnmatch.fnmatch` works only because `glob` happens to import `fnmatch`; a plain `import fnmatch` is the same length and doesn't depend on another module's internals. (Same in `image_in.py`, so leaving them consistent is defensible too.)

tests/test_filter_video_in.py
```diff
@@ -836,6 +838,348 @@ def test_config_params(self):
...
+                frames.append(frame_dict['main'])
+            self.assertTrue(len(frames) > 0)
+            # Override must be ignored because source is a directory.
+            self.assertEqual(frames[0].data['meta']['src'], f'file://{v_a}')
+        finally:
+            runner.stop()
+            queue.close()
+
+    def test_directory_transition_skip_bad_file(self):
+        import shutil
```
lucasmundim: Non-blocking. The top-level `import shutil` / `import tempfile` were added (lines 5-6), but the per-test copies are still there in eight of the nine new tests — only `test_directory_get_info` drops them. Worth deleting the rest.

tests/test_filter_video_in.py
```diff
@@ -836,6 +838,348 @@ def test_config_params(self):
...
+                queue   = (queue := FiltersToQueue.Queue()).child_queue,
+            )),
+        ], exit_time=4)
+
+        try:
+            # Delete v_b right after VideoIn initializes its directory list.
+            # When VideoIn transitions to v_b, the file won't exist anymore,
+            # which fails instantly inside OpenCV without any FFmpeg demuxer stall!
+            if os.path.exists(v_b):
+                os.unlink(v_b)
```
lucasmundim: Non-blocking. This test doesn't reliably exercise the path it describes. The `unlink` races `VideoReader.__init__`'s directory scan, which runs in the `Filter.Runner` subprocess: if the delete wins, `02_bad.mp4` is simply never in `dir_files` and the run is a plain two-file directory — six frames from `01_good` and `03_good`, every assertion green, transition-skip code never entered. Writing a file that is corrupt but present (valid extension, garbage bytes, so it passes the extension allowlist and fails at open) tests it deterministically.

Still-open coverage from the last round, while you're here: the non-`sync` `maxfps` pacing across a transition (the `ns_per_fps` bug above), and a directory whose files differ in resolution or fps.

RELEASE.md
```diff
@@ -4,10 +4,19 @@ OpenFilter Library release notes
...
+  Users can now pass a directory of video files as a source (e.g. `file:///path/to/folder`).
+  The video files inside are sorted alphabetically and played sequentially. Includes full
+  support for loop and sync, recursive subfolder scanning via `recursive`, and glob-based file
+  pattern filtering via `pattern`.
+
 ## v1.3.0 - 2026-08-17
 
 ### Added
 
+
```
lucasmundim: Non-blocking. Stray blank line added under the released v1.3.0 heading — unrelated to this change.

lucasmundim (changes requested):
Re-reviewed at `1724643f`. All ten threads from the previous round are addressed and the two red checks are green now. Verified on the branch: the `sync` `src_fps` guard, the per-frame `source`/`fps` snapshot through `extras` (that kills the attribution race by construction, not by luck), the `override_source_uri` directory gate, the extension allowlist + `pattern`/`recursive` scan replacing the `isOpened()` probing, the bad-file skip, the `_open_dir_file` dedupe, the attribute init, and the demo cleanup — all good.

Three things still open:

1. The `sync` half of the fps fix landed, the non-`sync` half didn't — `_open_dir_file` recomputes the decoder pacing from the `maxfps`-capped `self.fps`. Since `sync` defaults to False, this is the *default* path for a directory plus `maxfps`. Reproduced on this branch; details inline.
2. The feature is still undocumented. Same point as my last review and shingonoide's, and the one thing neither round touched: `!pattern` / `!recursive` are accepted by `normalize_config` but appear in no docstring, `VideoReader.__init__` still opens with "Read a single video file", and `docs/video-in-filter.md` has no mention of directory sources.
3. shingonoide's review asked for a short design note before this lands; I don't see a reply or a link to one on the PR. Theirs to clear — flagging it only so it isn't lost.

Repro for (1) — two copies of the 3-frame/30 fps test clip in a temp dir, `!maxfps=1`, no `!sync`:

```
directory : ns_per_fps 33,333,333 -> 1,000,000,000 after the transition; 4 frames over 4.14s
single file !maxfps=1!loop=2 (unchanged path): ns_per_fps stays 33,333,333; 1 frame over 0.24s
```

Nits are marked non-blocking.

shingonoide (changes requested):
The four functional bugs are genuinely fixed (verified in source), and the directory-scan logic now follows ImageIn's pattern (extension allowlist, !pattern, !recursive) — good use of that precedent. Still outstanding: no design doc/ADR/ticket has been linked anywhere, and the docs debt is unaddressed — docs/video-in-filter.md and the VideoIn/VideoReader docstrings still don't mention directory sources despite the code shipping the capability. Please close the design-doc ask and update the docs before this lands.

RELEASE.md
ahonjaya-del: removed

tests/test_filter_video_in.py
ahonjaya-del: fixed. bad.mp4 is now actually garbage bytes.

tests/test_filter_video_in.py
ahonjaya-del: removed

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: agreed. i have imported it myself.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: fixed - extracted out to its own function

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: updated

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: documentation added

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Reverted `{self.ssource}` back to `{self.source!r}` in the failed loop-restart RuntimeError message

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: updated _open_dir_file() to compute the pacing interval ns_per_fps using self.native_fps instead of the capped self.fps when sync is disabled.                                                                                                                           

Covered by the new dedicated unit test `test_nosync_mode_fps_preservation_across_transition`.

ahonjaya-del:
Small design doc at [Confluence](https://plainsight-ai.atlassian.net/wiki/spaces/ENG/pages/2967535619/Video-In+Folder+Support+Design+Doc) and JIRA ticket at [FILTER-649](https://plainsight-ai.atlassian.net/browse/FILTER-649)

[FILTER-649]: https://plainsight-ai.atlassian.net/browse/FILTER-649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

docs/video-in-filter.md
```diff
@@ -744,3 +774,5 @@ class VideoIn(Filter):
 - `FILTER_MAXFPS`: Maximum frame rate
 - `FILTER_MAXSIZE`: Maximum image size
 - `FILTER_RESIZE`: Image resize dimensions
+- `FILTER_PATTERN`: Glob/wildcard pattern for directory video files
+- `FILTER_RECURSIVE`: Scan directories recursively for matching videos
```
lucasmundim, line 778: **Blocking (docs).** Neither of these environment variables exists.

`video_in.py` reads exactly six (lines 26-31): `VIDEO_IN_BGR`/`FILTER_BGR`, `SYNC`, `LOOP`, `MAXFPS`, `MAXSIZE`, `RESIZE`. There is no `VIDEO_IN_PATTERN`/`FILTER_PATTERN` and no `VIDEO_IN_RECURSIVE`/`FILTER_RECURSIVE`. `pattern` and `recursive` reach the reader only via the config field or the `!option` suffix - both of which you documented correctly above.

`FILTER_RECURSIVE` is worse than a dead entry: it is a *real* variable, but `ImageIn`'s (`image_in.py:36`, `IMAGE_IN_RECURSIVE or FILTER_RECURSIVE`). A pipeline that exports `FILTER_RECURSIVE=true` gets `ImageIn` recursing and `VideoIn` silently not - exactly the split the `FILTER_*` standardization exists to remove.

Either drop the two lines, or wire them up the way every other option here is: `VIDEO_IN_RECURSIVE`/`VIDEO_IN_PATTERN` alongside lines 26-31, defaulted in `VideoReader.__init__` next to `maxsize`/`resize`, added to the env list in the `VideoIn` docstring (lines 915-921), and covered in `tests/test_filter_env_vars.py` (which already allowlists `FILTER_RECURSIVE`). Wiring them up is the better end state and matches `ImageIn`.

(`FILTER_PATTERN` is documented for `ImageIn` at `docs/image-in-filter.md:1084` and isn't implemented there either - that one predates this PR; just don't propagate it.)

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -720,6 +882,13 @@ class VideoIn(Filter):
             together with `maxsize`, it is one or the other. Set here to apply to all sources or can be set individually
             per source. Global env var default FILTER_RESIZE / VIDEO_IN_RESIZE.
 
+        pattern:
+            Only has meaning for folder/directory file:// sources. A glob pattern (e.g. `*.mp4`) or regular expression
+            to match and filter video files in the directory. Set here to apply to all sources or can be set individually per source.
+
+        recursive:
+            Only has meaning for folder/directory file:// sources. If True, scans the directory and its subdirectories recursively for matching files. Set here to apply to all sources or can be set individually per source.
```
lucasmundim: **Still open - 2 of the 4 docs asks.** The `!pattern` / `!recursive` entries, these two config paragraphs, the `VideoReader.__init__` docstring and the new `docs/video-in-filter.md` section all landed. Two are untouched:

1. `sources` (lines 796-797) still reads "can be file://, rtsp:// stream, s3:// bucket object, or a webcam:// index", and none of its examples show a directory. That is the first thing a user reads about what `sources` accepts, and the one remaining place that says directories aren't a thing. One clause plus a `'file:///path/to/dir!pattern=*.mp4'` entry in the existing example list would do it.

2. "Emitted frame meta" (lines 898-913) has no directory note. `meta['src']` now changes mid-stream, and `src_frame`/`src_seconds` restart at 0 at every file boundary *within a single non-looping run*. Lines 909-911 already warn about exactly that non-monotonicity for `loop` and tell consumers to key off `id` - directories reproduce it without `loop`, so that warning needs to name them.

tests/test_filter_video_in.py
```diff
@@ -836,6 +838,376 @@ def test_config_params(self):
...
+        with open(v_path, 'wb') as f:
+            f.write(RED_THEN_GREEN_THEN_BLUE_FRAME_MP4)
+
+        height, width, fmt, fps = VideoReader.get_info(f'file://{test_dir}')
+        self.assertGreater(height, 0)
+        self.assertGreater(width, 0)
+        self.assertEqual(fmt, 'BGR')
+        self.assertGreater(fps, 0.0)
+
+    def test_nosync_mode_fps_preservation_across_transition(self):
```
lucasmundim: Non-blocking. This test passes on the buggy code, so it does not guard the regression it was added for.

Checked against the branch: I reinstated `self.ns_per_fps = 1_000_000_000 // (self.fps or 15)` in `_open_dir_file` and re-ran a two-clip directory at `maxfps=10`. `extras['fps']` - which becomes `meta['src_fps']` - is `10` on every frame either way, because `self.fps` is set to `maxfps` in both versions. What moved was `ns_per_fps` (33,333,333 -> 100,000,000) and the frame count (2 -> 4: the run stretched to 1/10th realtime and `wait()` stopped skipping).

The docstring's reasoning is circular here: "`ns_per_fps` is recomputed purely from `native_fps` regardless of `maxfps`, so it would be unaffected by `maxfps` handling" describes the *fixed* code. On the pre-fix code `ns_per_fps` is precisely the discriminator, which is why I suggested it. Keeping the end-to-end `src_fps` assertion is fine - it just isn't this bug. A direct `VideoReader` assertion on `ns_per_fps` after the transition (the style `test_directory_get_info` already uses) covers it in three lines; asserting the frame count works too and is closer to what a user notices.

Also still uncovered after two rounds: a directory whose files differ in resolution or fps.

lucasmundim (changes requested):
Re-reviewed at `40bd54ac`. Both blocking bugs from the last round are genuinely fixed, and I verified each on the branch: `_open_dir_file` now paces from `native_fps` (`ns_per_fps` stays 33,333,333 across a transition at `!maxfps=10`, where it previously became 100,000,000), and the loop-restart `RuntimeError` is back to `{self.source!r}`, so a presigned S3 URL no longer reaches the ERROR log. The `scan_dir_videos` extraction settles the `get_info`/`__init__` divergence properly, `fnmatch` is imported directly, the `read()` docstring now matches what `extras` actually carries, `02_bad.mp4` is deterministically corrupt instead of racing the scan, and the stray `RELEASE.md` blank line is gone. All 19 checks green.

Three items left, all docs and tests - no functional bugs this round.

The one that needs fixing is new in this round: `docs/video-in-filter.md` now advertises two environment variables that don't exist. The other two are carry-overs from the docs thread - the `VideoIn` `sources` description and the "Emitted frame meta" section still don't know about directories - plus the non-`sync` pacing regression is still not guarded by a test (I confirmed the new one passes on the buggy code).

Separately, @shingonoide's design-doc/ADR point is still open and is not mine to close.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Verified at `40bd54ac` - `_open_dir_file` now uses `native_fps`; `ns_per_fps` stays 33,333,333 across a transition at `!maxfps=10` where it previously became 100,000,000. Fix confirmed.

One caveat on the new test, left as a separate comment on `test_nosync_mode_fps_preservation_across_transition`: it passes on the pre-fix code, so it doesn't guard this regression.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Confirmed at `video_in.py:503` - back to `{self.source!r}`, so the presigned URL no longer reaches the ERROR log. Thanks.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Partially addressed. The `!pattern` / `!recursive` entries, the two config paragraphs, the `VideoReader.__init__` docstring and the new `docs/video-in-filter.md` section all landed - thanks.

Two of the four asks are untouched, details in a new comment: the `sources` description (lines 796-797) and the "Emitted frame meta" section (lines 898-913) are still directory-unaware.

Flagging separately: `docs/video-in-filter.md` now documents `FILTER_PATTERN` and `FILTER_RECURSIVE`, neither of which `video_in.py` reads.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Confirmed - the comment now describes the real contract.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Confirmed - `scan_dir_videos` is now the single source of truth for both `__init__` and `get_info`, and `get_info` takes `pattern`/`recursive`. Exactly what I was after.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: `import fnmatch` is in, thanks.

`matches_pattern` and the extension helper are still duplicated between `video_in.py` and `image_in.py` - I flagged that as defensible either way, so I'm not pushing it further here.

tests/test_filter_video_in.py
lucasmundim: Confirmed - all per-test `shutil` / `tempfile` imports are gone.

tests/test_filter_video_in.py
lucasmundim: Confirmed - `02_bad.mp4` is now present-and-corrupt, so the transition-skip path is entered deterministically.

The two coverage gaps at the end of that comment are still open: the non-`sync` pacing one (see my comment on `test_nosync_mode_fps_preservation_across_transition`) and a directory whose files differ in resolution or fps.

RELEASE.md
lucasmundim: Confirmed - blank line gone.

shingonoide (commented):
Nice, this closes out most of what was outstanding. The design doc and ticket (FILTER-649) are now linked, and the docs got a real directory-sources section plus API reference updates.

Two things still open. First, the "Emitted frame meta" docstring only mentions src_frame/src_seconds resetting "With loop" — a directory source resets both at every within-run file transition too, independent of loop, same non-monotonicity as the loop case. Second: this round also changed non-sync pacing to derive from native_fps instead of the maxfps-capped rate. The new test confirms the reported src_fps meta value stays capped correctly, but it doesn't exercise the actual pacing/throttle interval itself, so it's not clear whether real playback speed still honors maxfps after this change or now always paces at native rate. Could you confirm whether that's intentional, and if so add a note (or a pacing-specific test) so it's not ambiguous.

docs/video-in-filter.md
ahonjaya-del: fixed - wired in both recursive and pattern arguments. Added corresponding tests

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: docs should be added now.

tests/test_filter_video_in.py
ahonjaya-del: Replaced `test_nosync_mode_fps_preservation_across_transition`, which passed even with the bug it was meant to catch, with two tests: one for meta['src_fps'] capping, one that actually regresses on ns_per_fps

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -27,6 +29,8 @@
 VIDEO_IN_MAXFPS   = None if (_ := json_getval((os.getenv('VIDEO_IN_MAXFPS') or os.getenv('FILTER_MAXFPS') or 'null').lower())) is None else float(_)
 VIDEO_IN_MAXSIZE  = os.getenv('VIDEO_IN_MAXSIZE') or os.getenv('FILTER_MAXSIZE') or None
 VIDEO_IN_RESIZE   = os.getenv('VIDEO_IN_RESIZE') or os.getenv('FILTER_RESIZE') or None
+VIDEO_IN_PATTERN  = os.getenv('VIDEO_IN_PATTERN') or os.getenv('FILTER_PATTERN') or None
+VIDEO_IN_RECURSIVE = bool(json_getval((os.getenv('VIDEO_IN_RECURSIVE') or os.getenv('FILTER_RECURSIVE') or 'false').lower()))
```
lucasmundim, line 33: Non-blocking. Wiring these up is the right call, but it puts `FILTER_PATTERN` on the other side of the split I described last round.

`ImageIn` reads `FILTER_RECURSIVE` (`image_in.py:36`), so `recursive` is now symmetric between the two filters - good. It has no env default for `pattern` at all, even though `docs/image-in-filter.md:123` tells users to `export FILTER_PATTERN="*.jpg"`. So a pipeline that exports `FILTER_PATTERN=*.mp4` now gets `VideoIn` filtering and `ImageIn` silently not.

`VideoIn`'s side of that is the correct end state, so nothing to change here - it just means `ImageIn` needs the matching `IMAGE_IN_PATTERN` / `FILTER_PATTERN` default in a follow-up, otherwise the doc keeps promising something only one of the two filters honors.

Nit while you are on this line: `VIDEO_IN_RECURSIVE` breaks the `=` alignment of the block above (lines 26-32 all line up), and the same happens at line 934 in the docstring env list, where `FILTER_RECURSIVE / VIDEO_IN_RECURSIVE` is one space short of the column the others use.

openfilter/filter_runtime/filters/video_in.py
```diff
@@ -495,8 +628,31 @@ def read(self, with_tframe=False):  # -> np.ndarray | tuple[np.ndarray, int, dic
         return image_n_tframe if with_tframe else image_n_tframe[0]
 
     @staticmethod
-    def get_info(source: str) -> tuple[int, int, str, float]:  # (height, width, format, fps)
-        cap = cv2.VideoCapture(int(source[9:]) if is_video_webcam(source) else source[7:] if is_video_file(source) else source)
+    def get_info(source: str, pattern: str | None = None, recursive: bool | None = None) -> tuple[int, int, str, float]:  # (height, width, format, fps)
+        path = source[7:] if is_video_file(source) else source
+        if os.path.isdir(path):
+            # First file per scan_dir_videos, so this reports the file __init__ will actually play
+            dir_files = scan_dir_videos(path, pattern, recursive)
```
lucasmundim, line 638: Non-blocking. `get_info` does not pick up the env defaults that `__init__` now applies, so the two can disagree on which files a directory contains.

Lines 218-219 resolve `pattern` and `recursive` from `VIDEO_IN_PATTERN` / `VIDEO_IN_RECURSIVE` when the caller passes `None`. `get_info` defaults both to `None` too, but hands them straight to `scan_dir_videos` without that resolution step. With `FILTER_RECURSIVE=true` and no explicit argument, `VideoReader('file:///dir')` scans recursively while `VideoReader.get_info('file:///dir')` scans depth-1 - and the first file can genuinely differ, since `sorted()` puts `/dir/sub/00.mp4` before `/dir/01.mp4`. `get_info` would then report the geometry and fps of a file the reader never opens first.

This is the same `__init__` / `get_info` divergence `scan_dir_videos` was extracted to remove, reopened one level up at the argument-resolution step. Still latent - `get_info` has no caller in the repo - so non-blocking, and it is two lines to settle: apply the same `VIDEO_IN_PATTERN if pattern is None else pattern` defaulting here, or move that resolution into `scan_dir_videos` so both callers get it for free.

lucasmundim (approved):
Re-reviewed at `5e9a4952`. Everything I raised last round is genuinely addressed, and I verified each on the branch.

The `pattern` / `recursive` env vars are wired the way the rest of the options are - `VIDEO_IN_PATTERN` / `VIDEO_IN_RECURSIVE` alongside lines 26-31, resolved in `VideoReader.__init__` at lines 218-219, added to the docstring env list, and covered in `tests/test_filter_env_vars.py` across the FILTER_*, legacy, precedence and default cases. The two remaining docs asks landed: `sources` now names directories and carries a directory example, and the "Emitted frame meta" section now tells consumers that `src_frame` / `src_seconds` restart and `meta['src']` changes at every file transition, even without `loop`.

The test split is the right shape, and the new regression test does its job. I checked it the same way as last time - reinstated `self.ns_per_fps = 1_000_000_000 // (self.fps or 15)` at `video_in.py:400` in a scratch copy and re-ran it:

```
test_nosync_mode_ns_per_fps_native_after_transition
AssertionError: 1000000000 != 33333333.0
```

It passes on the real code and fails on the bug, which is exactly what was missing before. `test_directory_files_differing_fps_and_resolution` closes the last coverage gap. `tests/test_filter_video_in.py` + `tests/test_filter_env_vars.py` at this head: 144 passed locally.

No functional findings this round. Approving on the code - but this is not ready to merge yet, for two reasons that are not about the diff:

1. **The branch conflicts with `main` and CI never ran on this head.** `RELEASE.md` changed on both sides, so the merge state is `DIRTY`, GitHub cannot build the merge ref, and the `pull_request` workflows did not fire: `5e9a4952` has 3 checks (`DCO`, `dismiss`, `[code]smith`) where `40bd54ac` had 19. A rebase on `origin/main` will restore them - please get them green before merging.
2. @shingonoide's design-doc ask is theirs to clear. You linked a design doc and a ticket in a PR comment after their last review, so the ball is with them, not you - flagging it only so it is not lost.

Two non-blocking notes inline, neither of which needs to hold the merge.

Also, when you get a chance, please resolve the threads from the previous rounds - nine of them are still open with only my "confirmed" as the last comment.

docs/video-in-filter.md
lucasmundim: Confirmed - and you took the better of the two options. `VIDEO_IN_PATTERN` / `VIDEO_IN_RECURSIVE` are read alongside lines 26-31, defaulted in `VideoReader.__init__` at lines 218-219, documented in the `VideoIn` env list, and covered in `tests/test_filter_env_vars.py` across all four cases the other options get (FILTER_*, legacy, precedence, default). The two lines here are no longer dead, and `FILTER_RECURSIVE` no longer means one thing to `ImageIn` and nothing to `VideoIn`.

One leftover, which I put in a new comment on line 33 rather than reopening this: `FILTER_PATTERN` is now the asymmetric one - `ImageIn` still has no env default for `pattern` while `docs/image-in-filter.md:123` tells users to export it. That is `ImageIn`'s gap to close in a follow-up, not yours here.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Confirmed - both landed.

`sources` (lines 797-816) now names the directory case and carries a `file:///path/to/dir!pattern=*.mp4` entry through all three example forms, so a user reading the first paragraph about what `sources` accepts learns directories are a thing.

"Emitted frame meta" (lines 917-922) now says what I was after: `src_frame` / `src_seconds` reset to 0 and `meta['src']` changes at every file transition, `meta['id']` keeps counting, and consumers must key off `id`. It sits right after the `loop` paragraph it parallels. Thanks - the docs debt from the first round is closed as far as I am concerned.

tests/test_filter_video_in.py
lucasmundim: Confirmed, and the new test genuinely guards the bug. I checked it the same way I checked the old one - reinstated `self.ns_per_fps = 1_000_000_000 // (self.fps or 15)` at `video_in.py:400` in a scratch copy of this head and re-ran:

```
FAILED tests/test_filter_video_in.py::TestVideoIn::test_nosync_mode_ns_per_fps_native_after_transition
AssertionError: 1000000000 != 33333333.0
```

Passes on the real code, fails on the bug. The split is the right shape too, and both docstrings now say honestly what each test does and does not cover.

`test_directory_files_differing_fps_and_resolution` closes the other gap - asserting per-file `src_fps` and image shape across the transition covers exactly the "not the previous file's, not the first file's" case. Nothing further from me on tests.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: Agreed will update ImageIn in a future PR.

openfilter/filter_runtime/filters/video_in.py
ahonjaya-del: fixed. Now the helper resolves the arguments.

lucasmundim (approved):
Re-reviewed `5b06c279..518ecdc`.

`518ecdc` closes the last thread I had open: `pattern` / `recursive` env resolution now happens once inside `scan_dir_videos` (lines 78-79) instead of in `VideoReader.__init__`, so `__init__` and `get_info` can no longer disagree on which files a directory contains. An explicit `!no-recursive` still wins over `VIDEO_IN_RECURSIVE=true`, since `False is not None`.

I also went back over `f8ba157`, which landed just after the tree I approved against. `test_nosync_mode_maxfps_paces_delivery_across_transition` is the right complement to the `ns_per_fps` test: at `native=100fps` / `maxfps=20` it asserts delivered-frame spacing - including the gap spanning the transition - never collapses toward the native rate. That also answers @shingonoide's second open question directly: pacing from `native_fps` is intentional, `maxfps` is still enforced by the separate `ns_per_maxfps` skip logic in `wait()`, and the test now proves it end to end rather than only checking the reported `src_fps` label. Their first point is closed too - the "Emitted frame meta" docstring now states that a directory source resets `src_frame`/`src_seconds` at every file transition independent of `loop`.

All 19 checks green on `518ecdc` and the branch is level with `main`. No findings from me - approving.

Two housekeeping items, neither blocking on the code: the PR still carries a `CHANGES_REQUESTED` from @shingonoide's earlier review at `1724643f`, which their later `COMMENTED` review does not clear - it needs an explicit re-review from them before this can merge. And most of the threads from the last two rounds are answered but still open; worth resolving them so what's left is visible at a glance.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Confirmed at `518ecdc`. `scan_dir_videos` resolves `VIDEO_IN_PATTERN` / `VIDEO_IN_RECURSIVE` itself now (lines 78-79) and `__init__` no longer pre-resolves them, so `get_info` and `__init__` scan identically by construction rather than by both remembering to apply the same defaults. Better than what I suggested - thanks.

openfilter/filter_runtime/filters/video_in.py
lucasmundim: Understood, a follow-up PR is the right place for it - `VideoIn` shouldn't carry `ImageIn`'s gap. Nothing further from me here.

shingonoide (approved):
Design doc (Confluence) and FILTER-649 now linked, and the docs/docstring gap is closed - video-in-filter.md gets a solid quick-start section for directory sources, and the VideoIn/VideoReader docstrings cover the pattern/recursive env-var defaults, the !pattern=/!recursive URL-option syntax, and the src_frame/src_seconds reset behavior at directory transitions. The scan_dir_videos refactor centralizes env-var default resolution so __init__ and get_info can't diverge. The new ns_per_fps-after-transition regression test is a solid mechanism-level check, not just a smoke test. One small non-blocking note: matches_pattern always dispatches to fnmatch when the pattern contains */?, even if a regex containing those chars was intended - untested edge case, not worth blocking on. Nothing else from our side.

PR #162, opened 2026-09-03
fix(analytics): report usage off-thread and honour the opt-out in the log

lucasmundim:
The Scarf event reported once per `Filter` init goes out through a **synchronous POST with a 3s timeout**, run inline in `Filter.__init__`.

Anywhere the endpoint is reachable-but-silent — a firewall that DROPs rather than REJECTs, a captive network, an air-gapped host — every filter pays up to 3s of startup, multiplied by the number of filters in the pipeline. A ten-filter pipeline can lose half a minute before doing any work. This is not hypothetical: it is what a filter does on a cluster whose egress allowlist does not include `python.openfilter.io`.

## Changes

**Report on a daemon thread.** A slow or blackholed endpoint can no longer delay construction, and being a daemon it never holds up process exit. What is sent is unchanged, and the SDK still has the final say on the opt-out.

**Fix the startup log.** `Filter` printed `Usage analytics enabled via Scarf` unconditionally, including for users who had set `DO_NOT_TRACK`. Telling someone their opt-out did nothing is worse than saying nothing. `_usage_analytics_disabled()` mirrors the two variables the SDK honours, so the log matches reality and no thread is started when opted out.

This does **not** change what is collected, and does not touch the opt-out contract — analytics stay on by default.

## Tests

`tests/test_usage_analytics.py`, 8 cases:

- opt-out matrix: both variables, truthy (`1`/`true`/`TRUE`/`True`) and falsy (`0`/`""`/`false`/`no`) values, and either variable being sufficient. The falsy cases matter — an explicit `DO_NOT_TRACK=0` must not be read as a generic "variable is present" signal.
- dispatch: forwards the filter name; swallows SDK errors (an exception escaping a thread would print a bare traceback with nothing to catch it); and returns immediately when the reporter stalls, which is the regression itself.

## Verification

Measured against a blackholed address, dispatch returns in **0.0002s** where the inline call took the full ~3s.

Full unit suite green: **835 passed, 1 skipped, 2 deselected, 0 failed** (`pytest tests/ --ignore=tests/integration --benchmark-disable -m "not slow"`). The 8 new tests pass both with and without `DO_NOT_TRACK` set in the environment.

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -4,6 +4,24 @@ OpenFilter Library release notes
 
 ## [Unreleased]
 
+### Changed
+
+- **Usage analytics no longer block filter startup.** The Scarf event reported once
+  per `Filter` init went out through a synchronous POST with a 3s timeout, run
+  inline in `Filter.__init__`. Anywhere the endpoint is unreachable but not
+  actively refused — a firewall that DROPs, a captive network, an air-gapped
+  host — every filter paid up to 3s of startup, multiplied by the number of
+  filters in the pipeline. The report now runs on a daemon thread, so a slow or
+  blackholed endpoint can neither delay startup nor hold up process exit. What is
+  sent, and the `DO_NOT_TRACK` / `SCARF_NO_ANALYTICS` opt-out, are unchanged.
+
+### Fixed
+
+- **The startup log no longer claims analytics are enabled when they are opted
+  out.** `Filter` printed `Usage analytics enabled via Scarf` unconditionally,
+  including for users who had set `DO_NOT_TRACK`. It now reports the actual state
+  and skips the reporting thread entirely when opted out.
+
 ## v1.3.0 - 2026-08-17
 
 ### Added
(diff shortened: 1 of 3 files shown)


shingonoide (approved):
This looks solid. Moving the Scarf POST off the critical path is the right fix — I checked the scarf package source and confirmed log_event really is a synchronous request with a 3s default timeout, so a blackholed endpoint would add that delay per filter in a pipeline. Since the SDK's own opt-out check already gates whether the network call happens at all, the daemon thread doesn't change behavior for anyone using DO_NOT_TRACK or SCARF_NO_ANALYTICS today. The new disabled-check just keeps the startup log honest and avoids spinning up a thread that would do nothing.

Tests cover the opt-out matrix well, including the falsy-value case (DO_NOT_TRACK=0), which is the one people usually get wrong, plus the non-blocking dispatch itself. CI is green across the board.

Nice fix.

